PR #7, opened 2019-01-02
button.components-button.kt-size-btn position

bucketpress:
In the Tabs block settings or blocks with kt-size controls, under controls such as Title Spacing, after selecting 'Linked' button, the second button 'Individual' becomes inaccessible due to positioning/z-index issue. Adding position:relative to solve the problem.


kadencethemes:
Thanks for the pull! chrome didn't have this issue so I missed it.

PR #8, opened 2019-01-03
Added styles for 60-40 columns

bucketpress:
Was missing the styles for 60-40 columns.


kadencethemes:
thanks!

bucketpress:
Hi, just checking when will these latest merges be built to dist?

PR #14, opened 2019-01-16
Render CSS outside of main content 

bucketpress:
Style tags within main content is treated as part of visible content by some plugins and causes havoc to content rendering.


kadencethemes:
It's a good option. I suppose it does mean there could be a flash of unstyled content. I preferred CSS in the head but that has other issues and having talked to some Gutenberg leads they recommended the switch to some style tags.

For now, I am going to add this and maybe look at adding an option for the user to use the parsing method in the head. I could even look at running a function when the block runs to see if the css has been rendered or not then add inline if it hasn't.

PR #20, opened 2019-02-10
add support for accordion on tabs desktop layout

toffebjorkskog:
I added support for accordion tabs also on desktop layout.


kadencethemes:
Hey, Thanks for this! But because this block is more designed to be tabs and can collapse as accordion I think it's best if we make a different accordion block. This would allow for a live preview of the accordion in the editor, it would also make loading a little better since the titles wouldn't need to be recreated via javascript and input into the content to create the accordion fall back.  I will look at your pull and maybe there is a use case for it but I've already started on an accordion block that I think would do this a little better.

kadencethemes:
I updated to include the accordion block, please let me know what you think of this option.

toffebjorkskog:
This is fantastic!

toffebjorkskog:
I see also that you can have default settings sometime in the future which looks very promising.

PR #37, opened 2019-03-20
Declare and export block's attributes for later use of default values

RodrigoDAgostino:
The Advanced Button's attributes have been declared and exported as a variable in «block.js» for later use of the default values in «edit.js»


kadencethemes:
Hey,
 Thanks for your efforts! I won't push in though. The reason is that defaults can be set by the user which filters and applies to the block on load by assigning attributes on load. The other reason is to me it makes more sense to copy the first buttons styles when creating more, not start from whatever the default is. Thanks!

RodrigoDAgostino:
Not a problem! It was mainly a suggestion. Thanks for your time! :)

PR #66, opened 2019-09-27
Add support for EditorsKit navigator toolbar

phpbits:
Hi,

I've just recently added Block Navigation toolbar on my [EditorsKit](https://wordpress.org/plugins/block-options/) plugin. I think this will be of great help on user navigation when both plugins are activated. The toolbar will only be shown on blocks with multiple inner blocks capability.

- [x] Row
- [x] Column

 
Here's how it works:
Screen Capture on 2019-09-27 at 12-23-16

Let me know your thoughts.

Thanks,
Jeffrey

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+**/node_modules/**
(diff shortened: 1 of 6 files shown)


kadencewp:
Sweet, I'll merge this in and get an update pushed out tomorrow. Thanks for posting!

phpbits:
Thanks a lot @kadencewp! I really appreciate it, I'll look for other user-beneficial integrations when both plugins were activated 💯

PR #122, opened 2020-12-18
Add hooks for automatic Tooslet Dynamic Sources everywhere.

kmgalanakis:
This pull request adds a few filters, the minimum amount of them, in order for automatic Toolset Dynamic Sources integration to be allowed for the targeted blocks.


kmgalanakis:
I just noticed that you might most probably be releasing not from `master` but from `movetowpscripts` so I changed the base of my branch.

kadencewp:
Thanks! I need to get some cleanup done on Github here soon.

PR #136, opened 2021-05-25
Increase the number of groups displayed from MailerLite to 500

oakesjosh:
This also switches from the `SelectControl` component to the  react-select `Select`. This allows users to search the list of groups


src/blocks/form/mailerlite.js
```diff
@@ -207,13 +206,17 @@ class MailerLiteControls extends Component {
 						{ ! isFetching && hasGroup && (
 							<Fragment>
 								<h2 className="kt-heading-size-title">{ __( 'Select Group', 'kadence-blocks' ) }</h2>
-								<SelectControl
+								<div className="mailerlite-select-form-row">
+								<Select
```
kadencewp, line 210: Because react select saves as an array holding both the value and the label we will want to change this for backward compatibility. Also if they change their group name in mailerlite it could break this as well, so instead, we need to specify to only save the value which is the id of the group.

src/blocks/form/mailerlite.js
```diff
@@ -207,13 +206,17 @@ class MailerLiteControls extends Component {
...
 									value={ ( undefined !== this.props.settings[ 0 ].group && this.props.settings[ 0 ].group[0] ? this.props.settings[ 0 ].group[0] : '' ) }
```
kadencewp, line 211:
```suggestion
									value={ ( group ? group.filter( ( { value } ) => value.toString() === ( undefined !== this.props.settings[ 0 ].group && this.props.settings[ 0 ].group[0] ? this.props.settings[ 0 ].group[0].toString() : '' ) ) : '' ) }
```

kadencewp (changes requested):

src/blocks/form/mailerlite.js
oakesjosh: That makes sense. I committed those changes, updated the build files, tested with an existing list selection (saved from the SelectControl), and tested with a new list selection

PR #174, opened 2022-02-12
Add option to send remote IP using a hidden field on (email) forms.

VenomXNL:
A small and simple implementation which makes it possible to send the remote IP of the form submitter along with the form/email message to the website owner etc.
It was one of the few features I missed in the  Kadence Form Block, which does however make it (alot) easier to block IP's etc from the site when abuse is detected.

I have seen this implemented in several other 'form plugins' for Wordpress, however with this small 'feature' implemented in the Kadence Form Block it would not be needed at all for me to use any other plugin :)


VenomXNL:
Extra note:
If desired I could also simply add some other aliases/methods for other 'remote address' variables which can for example be used when people are using cloudflare etc (like HTTP_X_REAL_IP and HTTP_CF_CONNECTING_IP)

PR #185, opened 2022-03-14
Create Show More block

oakesjosh:
I feel like this block is missing something...? Any Ideas or feedback?


oakesjosh:
Updated with the changes we discussed last week. This feels like a **much** better block

PR #189, opened 2022-03-23
Fix KadencePanelBody error on advanced button

oakesjosh:
This panel shouldn't have had an index appended to its key


oakesjosh:
I'm going to go ahead and merge this since it's small and is needed to fix an active bug on master branch

Issue #165, opened 2022-02-01
Build block styles wihtout codekit


superewald:
### summary
Hi, I am trying to add a new feature (indented icon-lists). It works so far in the editor but the styles are not applied at the frontend.

### details
I modified the `src/blocks/icon-list` in order to allow indention of entries. Styles and scripts are both modified.

Running `npm run build` does compile scripts and the main stylesheet (`dist/build/blocks.css`) but it doesn't compile frontend styles to `dist/blocks/icon-list.style.build.css` and thus, changes to `src/blocks/icon-list/style.css` are not reflected at frontend! .

**How do I compile the frontend stylesheets (`dist/blocks/[name].style.build.css`) for the blocks ?**
The only guess I have is that codekit is used to compile such styles (which would be very bad for me as I don't own a license for that).

oakesjosh:
Hello! We are using codekit for compiling many of the styles. The codekit config is in the repo and they do have a free trial, but moving away from codekit is on our to-do list.

superewald:
> Hello! We are using codekit for compiling many of the styles. The codekit config is in the repo and they do have a free trial, but moving away from codekit is on our to-do list.

Thanks for the quick reply!  Two things:

1. I opened a PR for a simple feature which needs to recompile those block styles. Would it be sufficient to include every change except the recompiled block styles and outsourcing the compilation to someone who owns codekit ?

2. I may be able to help with refactoring to webpack. The only change required is to compile all block styles with webpack and save them to `dist/blocks/[name].style.build.css`, right ? I've already done that while fiddling with this plugin and could create a PR for it.

oakesjosh:
1. Yeah that works! Thank you for contributing

2. That would be great! That's the main thing codekit does, but there's a few others. If you just want to make a PR for what you have now that'd be great starting point too.

Here's the list I've come up with
`/src/blocks/*/style.scss` to `/dist/blocks/*.style.build.css`
`/dist/assets/css/src/*.scss` to `/dist/assets/css/*.css`
`/dist/assets/js/src/*.js` to `/dist/assets/js/*-min.js` 
`/dist/assets/js/kb-form.js` to `/dist/assets/js/kb-form-min.js`
`/dist/settings/dashboard.scss` to `/dist/settings/styles.css`
`/dist/vendor/kt-blocks-slick.scss` to `/dist/vendor/kt-blocks-slick.css`
`/dist/kt-accordion.js` to `/dist/kt-accordion-min.js`
[...]

PR #190, opened 2022-03-25
Refactor build system to gulp

superewald:
## summary

This PR introduces gulp scritps/configuration which should supersede codekit as discussed in #165.

## changes

Since this PR is a little bigger, below is a summary of the changes made. I tried to change only necessary things but it's still a bit bigger.

1. replace vendor js/css from `dist/` with npm packages
2. move js/scss source files from `dist/` to `src/`
3. move php & static files to `php/`
4. add gulp scripts for
    - compile & minify standalone js & scss using babel/sass
    - copy vendor assets for standalone scripts (js/css) from `node_modules` to `dist/`
    - build webpack bundles (`blocks.js` & `plugin.js`)
5. adjust paths of js/css files inside `*.php` files
6. minify all source files by default

> I tried to place the source files at meaningful locations but struggled with the folder structure (look at [restructure src directory](#proposal-restructure-the-src-directory) for more info).

As you can see, the `dist/` folder is now gitignored since it should be only exist localy after building. The output should not be committed to git.

## gulp

Gulp will handle compilation of standalone source files (= not contained in webpack bundles), compilation of webpack bundles and provide vendor assets. 

The configuration and scripts (gulp tasks) are located in `gulp/`.

In order to create a full build you should run: `gulp build` which is equivalent to

- `gulp clean`
- `gulp js`
- `gulp styles`
- `gulp php`
- `gulp webpack`

And creates the `dist/` folder which contains all files that needs to be distributed.

You can create a zip archive from `dist/` by running `gulp zip` which creates `zip/kadence-blocks.zip`.

## proposal: restructure the `src/` directory

Before this PR, my understanding is that files located at `src/` were meant to be used inside webpack bundles. Everything outside `src/` was subject of individual compilation.

In my opinion, this directory currently is poorly structured and looks overwhelming at first glance. We could solve this, and also make project structure more semantic by refactoring it like so:

<details><summary><b>proposed directory structure</b></summary>

```
.
└── kadence-blocks/
    ├── gulp/
    │   ├── tasks/
    │   │   └── ...
    │   └── config.js
    ├── src/
    │   ├── frontend/                
    │   │   ├── plugin/                
    │   │   │   └── index.js
    │   │   ├── blocks/
    │   │   │   └── index.js
    │   │   ├── settings/
    │   │   │   ├── dashboard.scss
    │   │   │   └── scripts.js
    │   │   └── misc/
    │   │       ├── kb-gallery-simple-init.js
    │   │       └── ...
    │   └── backend/
    │       ├── templates/
    │       │   └── ...
    │       ├── init.php
    │       └── ...
    ├── static/
    │   └── images/
    │       ├── masks/
    │       │   └── ...
    │       ├── ajax-loader.gif
    │       └── graphy.png
    ├── kadence-blocks.php
    ├── package.json
    ├── gulpfile.js
    └── ...
```

</details>

Since this PR already introduces complex changes to the folder structure, this would be a perfect opportunity to refactor it for clarity. 

## issues

This PR is not finished yet and has some bugs/issues that needs to be resolved:

1. - [x] blocks are not registered within gutenberg (seems like webpack is missconfigured atm)
1. - [x] clean up gulp scripts (they are poorly written atm)
1. - [x] implement propper gulp config
1. - [ ]  handle sub-directories within `src/blocks/` while compiling with gulp (currently creates sub-directories in `dist/blocks/` which probably is not what should happen)


superewald:
The output of the webpack entries `blocks.js` & `plugin.js` looks wrong to me currently, I guess there is some wrong/missing configuration.

@kadencewp @oakesjosh Would you mind sharing the webpack config from codekit if there is one ? In the same matter, I could also need babel configuration (but thats lower priority as the standalone files seem to work).

superewald:
All issues with the build system have been resolved, I didn't test all features rather than confirming (random) blocks are loaded and styled in gutenberg aswell as in the frontend. 
There might be some edge cases I overlooked so its up to you to validate functionality.

oakesjosh:
This is fantastic! Things are working well in my testing

Only issue I've found is we include `row.style.build.css` on the front end. The block folder is named `row-layout` so the generated file name is `row-layout.style.build.css`. Mostly making a note if this for my own reference. We'll need to look at other multi-word block folders where the name of the front end CSS doesn't match the block folder name.

superewald:
> This is fantastic! Things are working well in my testing

Glad to hear :)

> Only issue I've found is we include `row.style.build.css` on the front end. The block folder is named `row-layout` so the generated file name is `row-layout.style.build.css`. Mostly making a note if this for my own reference. We'll need to look at other multi-word block folders where the name of the front end CSS doesn't match the block folder name.

Yep, same counts for every file which filename differs in src and dist variants. For eg, the `restaurant-menu` is called `restaurant` after compiling and contains some files from within subdirectories. That's one of the "edge cases" I was talking about as honestly I'm too lazy to fix wrong naming patterns :)

kadencewp:
Hey, Thanks for all your work on this!

I like the idea of refactoring the src files for better clarification. Blocks are all backend for gutenberg so I think I'll want to move some things around from how you currently have it.

I don't love moving php/static files around. Compiling the js to builds is simple enough to follow but instead of moving everything in the static folder I wonder what your thoughts are on just using an includes folder and making that part of the plugin? Any reason we shouldn't do that?

Generating language files has previously been a huge pain as WordPress.org wasn't able to create the files if we used SplitChunkName for both entries. I'll need to test that a bit for this build but right now it seems to be running.

superewald:
> Hey, Thanks for all your work on this!

I'm happy to contribute to this plugin! It is by far the best collection of gutenberg blocks I have found and maybe the only one which I consider real user-friendly (for wordpress authors).

> I like the idea of refactoring the src files for better clarification. Blocks are all backend for gutenberg so I think I'll want to move some things around from how you currently have it.

The current structure wasn't well thought, I just moved the necessary files from `dist/` to `src/` but honestly I didn't find a good place for them which is clear. I already thought about some refactorings but my knowledge about specific sources is just too limited for now. 

> I don't love moving php/static files around. Compiling the js to builds is simple enough to follow but instead of moving everything in the static folder I wonder what your thoughts are on just using an includes folder and making that part of the plugin? Any reason we shouldn't do that?

That would be perfectly fine, the only reason I moved those files was to get them off the `dist/` folder - which is in my perspective not a place for source files. But you are indeed right, moving the whole PHP stuff for every build is just overload not needed here. I tried to change least of the source files which is why I was copying php stuff to `dist/` instead of replacing paths.

> Generating language files has previously been a huge pain as WordPress.org wasn't able to create the files if we used SplitChunkName for both entries. I'll need to test that a bit for this build but right now it seems to be running.

Oh right, I didn't think about that at all. If you need help with webpack configs for that I'm here :)


I also want to mention there's still a known issue left which I don't know how to solve: `src/blocks/restaurant-menu` contains subfolders which in the current gulp configuration will output several individual files (one for every *.scss inside every subfolder). I can't see `menu-category` or `category-item` in the current master `dist/blocks/` folder so I don't know wether the subfolders need to be included and if they should be concatenated to the respective root folder.

Issue #193, opened 2022-03-27
[Feature] configurable value for options


superewald:
# summary

Currently we can only specify the label (which will be used as value also) in form field options (in select, checkbox, radiobox components).

# use case

 In order to use the forms for api's and such the value of the options should be configurable in the gutenberg editor.

# proposal

Add field "Value" to the options of select, checkbox and radiobox inside gutenberg editor.

oakesjosh:
I think this would be a good addition to #144. Users could set a dynamic or static field value

PR #194, opened 2022-03-27
[Feature] value textbox form field options

superewald:
# summary

Add value textbox field to form field options as described in #193.

# changes

- added value textbox in gutenberg editor for form field options

> This PR relies on #190 which is why it contains alot of changes. You can ignore all but the last commit of this PR.


superewald:
I'm currently trying to align the value and option textbox in one row but fail to do so, any hints how I can create a row of 2 textbox aside ? This is what it currently looks like:

Screenshot_20220327_125305

PR #220, opened 2022-10-31
dotLottie support

oakesjosh:
- Use dotLottie player exclusively on back end
- Use dotLottie player going forwards on front end, but keep old player for backwards compatibility
- Make block dynamic


kadencewp:
Task linked: [CU-4dwcwkg Lottie block issue with new .lottie files](https://app.clickup.com/t/4dwcwkg)

PR #228, opened 2022-12-21
CountUp aria label to read the end number to screen readers.

mark-c-woodard:
Hiding the current counting number so they don't get confused.

Also made a deprecation to address the markup change.

Addresses https://app.clickup.com/t/3waf50n


kadencewp (commented):
Thanks for looking into this!
1. We can't set static Ids in HTML because if there are two count up blocks on the same page that would mean two Ids and those need to be unique. We would need to add the unique ID as the final part of the label ID. 
2. I don't believe it's correct to add an aria-labelledby on a div without giving the div a role, but we should look into this and see because aria-hidden might change that, also something that is aria-hidden might not need a label in which case we just need to add the screen reader text without a label tag.
3. Might be worth doing this in the javascript when the count-up is initiated, that would give it an instant fix instead of forcing users to have to open the page/post with the block and updating it so that new markup is there.

mark-c-woodard:
Ya, the js method works well too and then we don't need a deprecation either. Simplifies it quite a bit.

Updated with that method and cleaned up all the id's labeled by stuff, etc. I think you're right, not need if they'd be labeling something that's hidden to the screen reader anyway.

PR #229, opened 2022-12-21
fix frontend output of advanced image block border radius

mark-c-woodard:
adding in css generation for the border radius breakpoints.

Addresses https://app.clickup.com/t/3wemwte


kadencewp (commented):
We have a new function for rendering things like this in the CSS class. render_measure_output. I'll push up a change so you can see if you want to implement in other places

kadencewp:
resolved with: https://github.com/kadencewp/kadence-blocks/commit/023b18db9c1a41090bb4f57aca0f2689552f6a5a

PR #237, opened 2023-01-09
[Row Layout] Fix "Add another section" button not increasing column count

oakesjosh:
PRing this change in case there's some functionality I'm missing. This button inserts a section but does not increase the column count.


kadencewp:
Not increasing the column count is intentional with this new button. It's a way to add a section without changing the column count in cases where you want the columns to be two but wants 4 sections, etc. If you want to increase the column count, you can use the columns option, by adding a column a new section is already generated to fill it.

PR #240, opened 2023-01-11
Feature/slider updates

mark-c-woodard:
addresses: https://app.clickup.com/t/863ftx70e

removing slick and tiny slider files.
testimonials is all working with splide
the row layout background slider is not working yet, still needs a fix for the "fade" mode in splide which has an issue where its slides after the first are translated out of position.


mark-c-woodard:
Row layout / other sliders using fade on the frontend should be working now. a translatex 0 for the frontend solved that.

There's still quite some funkiness around sliders though. Their styles are quite brittle. Since their primary stylesheet "kadence-splide" is shared between theme, blocks pro, and blocks styles can randomly change depending on what you have active.
We should make all the primary splide style sheets consistent across the products.

There's also an issue with dot styles, that I think I solved if you are only using blocks. You can see those styles in the "kadence-splide.scss" file in the pr.
But actually fixing it requires updates to theme / pro or doing more blocks free overrides, which only exacerbates the whole brittle issue.

advanced gallery dots are not showing unless pro is enabled.

So all in all, most slider features are working, but there's still quite a bit of wonkiness here

PR #241, opened 2023-01-11
adding aos data attributes to button frontend output

mark-c-woodard:
Addresses https://app.clickup.com/t/863ftyk9h

Accomplished via a helper function that other blocks can use if they're in a similar situation.


kadencewp:
#863ftyk9h[complete]

PR #244, opened 2023-01-13
Fixes from 3.0 google feedback form 61-66

mark-c-woodard:
A slew of minor fixes that address issues 61 - 66 in the 3.0 google feedback form: https://docs.google.com/forms/d/1WpajNh40NcCMawonBeEOVm_zN5F2WIBk5n0PTed0HUs/edit?ts=63bc35cf#responses

### DIFF
--- a/includes/blocks/class-kadence-blocks-row-layout-block.php
+++ b/includes/blocks/class-kadence-blocks-row-layout-block.php
@@ -1164,18 +1164,19 @@ class Kadence_Blocks_Rowlayout_Block extends Kadence_Blocks_Abstract_Block {
 	 * @param array $attributes for the block.
 	 */
 	public function get_video_render( $attributes ) {
-		if ( empty( $attributes['bgVideo'][0]['local'] ) ) {
+		if ( empty( $attributes['backgroundVideo'][0]['local'] ) ) {
 			return '';
 		}
 		$output = '';
+		$video_attributes = $attributes['backgroundVideo'][0];
 		$prevent_preload = $this->prevent_preload_when_hidden( $attributes );
 		$video_args = array(
 			'class' => 'kb-blocks-bg-video',
 			'id' => 'bg-row-video-' . $attributes['uniqueID'],
 			'playsinline' => '',
-			'muted' => ( isset( $attributes['bgVideo'][0]['local'] ) && false === $attributes['bgVideo'][0]['local'] ? 'false' : '' ),
-			'loop' => ( isset( $attributes['bgVideo'][0]['loop'] ) && false === $attributes['bgVideo'][0]['loop'] ? 'false' : '' ),
-			'src' => $attributes['bgVideo'][0]['local'],
+			'muted' => ( isset( $video_attributes['local'] ) && false === $video_attributes['local'] ? 'false' : '' ),
+			'loop' => ( isset( $video_attributes['loop'] ) && false === $video_attributes['loop'] ? 'false' : '' ),
+			'src' => $video_attributes['local'],
 		);
 		if ( ! empty( $attributes['bgImg'] ) ) {
 			$video_args['poster'] = $attributes['bgImg'];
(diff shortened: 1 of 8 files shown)


includes/blocks/class-kadence-blocks-row-layout-block.php
```diff
@@ -1266,6 +1266,9 @@ public function build_html( $attributes, $unique_id, $content, $block_instance )
 			if ( ! empty( $attributes['anchor'] ) ) {
 				$wrapper_args['id'] = $attributes['anchor'];
 			}
+			if ( ! empty( $attributes['kadenceDynamic'] ) || ! empty( $attributes['kadenceDynamic'] ) || ! empty( $attributes['kadenceDynamic']['bgImg'] ) ) {
```
kadencewp, line 1269: The logic here doesn't make sense to me. What are we checking for here?

includes/blocks/class-kadence-blocks-singlebtn-block.php
```diff
@@ -189,7 +189,7 @@ public function build_css( $attributes, $css, $unique_id ) {
 	 * @return mixed
 	 */
 	public function build_html( $attributes, $unique_id, $content, $block_instance ) {
-		$classes = array( 'kb-button', 'button', 'kb-btn' . $unique_id );
+		$classes = array( 'kb-button', 'kt-button', 'button', 'kb-btn' . $unique_id );
```
kadencewp, line 192: Is this because someone was adding custom CSS?

src/blocks/rowlayout/layout-controls.js
```diff
@@ -502,11 +503,11 @@ import { __ } from '@wordpress/i18n';
 							onChangeMobile={ ( value ) => {
 								setAttributes( { topSepWidthMobile: value } );
 							} }
-							min={ 100 }
-							max={ 400 }
+							min={ 0 }
```
kadencewp, line 506: The min should be 100 here, the width option is if someone wants to expand the divider, but the idea was that it always fills the edge, leave show unit so the unit is seen

includes/blocks/class-kadence-blocks-row-layout-block.php
```diff
@@ -215,15 +215,15 @@ public function get_template_columns( $css, $columns, $layout, $inner_selector,
...
+	 * @param array  $attributes the blocks attributes.
 	 * @param string $css the css class for blocks.
 	 * @param string $unique_id the blocks attr ID.
 	 */
 	public function build_css( $attributes, $css, $unique_id ) {
 		$css->set_style_id( 'kb-' . $this->block_name . $unique_id );
 		if ( ! empty( $attributes['kbVersion'] ) && $attributes['kbVersion'] > 1 ) {
 			$base_selector = '.kb-row-layout-id' . $unique_id;
-			$inner_selector = '.kb-row-layout-id' . $unique_id . ' > .kt-row-column-wrap';
+			$inner_selector = '.kb-row-layout-id' . $unique_id . ' > .kt-row-column-wrap.kt-row-column-wrap'; // Doubling .kt-row-column-wrap here to increase specificty to avoid conflict with alignment.
```
kadencewp, line 226: Can you explain how to recreate the issue, I want to look for a different way to fix the problem, this will effect all styles and I would like to avoid that

kadencewp (changes requested):

includes/blocks/class-kadence-blocks-row-layout-block.php
mark-c-woodard: checking to see if the bg img is dynamic. If it is it needs to be applied directly to the element style attribute so that it overrides the bg img set via css. For the case where the dynamic image is in a post loop.

We could have all bg imgs set this way and cut out the css entirely, but I think the image being dynamic is a smaller case and less efficient than writting it into the style sheet so added this condition in here.

includes/blocks/class-kadence-blocks-row-layout-block.php
mark-c-woodard: ya this one felt sketchy to me too..

you set a custom left or right padding on a row layout, then set the alignment to full width. On the fronted the custom padding won't apply.

includes/blocks/class-kadence-blocks-row-layout-block.php
mark-c-woodard: removed, will resolve this issue in pro

src/blocks/rowlayout/layout-controls.js
mark-c-woodard: changed back

src/blocks/rowlayout/layout-controls.js
```diff
@@ -585,10 +587,11 @@ import { __ } from '@wordpress/i18n';
 							onChangeMobile={ ( value ) => {
 								setAttributes( { bottomSepWidthMobile: value } );
 							} }
-							min={ 100 }
-							max={ 400 }
+							min={ 0 }
```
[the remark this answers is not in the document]
mark-c-woodard: changed back

includes/blocks/class-kadence-blocks-row-layout-block.php
mark-c-woodard: fixed this the other way by lowering the specificity of the other selector on alignment

PR #246, opened 2023-01-19
Updating row layout column resizing to better support rtl.

mark-c-woodard:
addresses https://app.clickup.com/t/4gxg719

Formerly tried to trick the editor into rtl with css. 
This caused the underlying javascript resizer to move the wrong way. 
Now instead of css fliping, we flip theresizer attributes/values as they are being applied to the underlying columns. This makes them apply correctly to their rtl flipped columns underneath. 
Now that the resizers are not flipped via css, the slide appropriately.


kadencewp:
So I was testing, and this only works if the user is always RTL or always LTR. We have users on multi-lingual sites that need the "first" column to be a specific size width. That first column might be on the left or right based on the language. The direction attribute in CSS makes this work, but when we flip the CSS that part breaks. 

I'm going to see if I can figure this out and push an update

kadencewp:
Task linked: [CU-4gxg719 [RTL] Adjusting Row Layout column widths is in the reverse direction](https://app.clickup.com/t/4gxg719)

PR #249, opened 2023-01-25
Support negative z-index on sections

oakesjosh:
PR-ing this since it's a row layout CSS change. This is adding back two CSS properties that were [set previously](https://github.com/kadencewp/kadence-blocks/blob/new-image-block/src/blocks/row-layout/style.scss#LL24C17-L24C17). 

Issue: Section block within a row layout has a -200 z-index. Section was not visible on front end. 

This is more of a backwards compatibility fix. We could adjust the starter template to work properly. Sections with a negative z-index that are outside of row layout will still not be visible on the front end.

ClickUp ID: #867780m2f

### DIFF
--- a/src/blocks/rowlayout/style.scss
+++ b/src/blocks/rowlayout/style.scss
@@ -9,6 +9,7 @@
 .kt-row-layout-inner, .kb-row-layout-wrap {
 	position: relative;
 	border: 0 solid transparent;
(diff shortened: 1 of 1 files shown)


kadencewp:
Task linked: [CU-867780m2f missing entire section](https://app.clickup.com/t/867780m2f)

PR #250, opened 2023-01-26
Updating splide / gallery scripts to support ajax loading

mark-c-woodard:
addresses: https://app.clickup.com/t/39j0c98

I've gone with the in-template script approach.

There's a small script in the gallery template that fires a slider init event that's defined in the main init script.
This makes it so the galleries themselves attempt to initialize themselves whenever they're rendered on the page. Such as after an ajax load.

I've also updated the init script to be re-run safe. So it won't try to re-init any sliders.


kadencewp:
I went with putting the reload trigger on whatever is running ajax that way we are not dropping an inline script into the content. For infinite scroll, that event will be called from Kadence Pro, which loads infinite scroll. It’s a more generic event because I could see us adding support for tabs, modals, or other things in the future.

mark-c-woodard:
Got ya.

As an aside; the reason I went the inline script route was to have broader support beyond just our plugins and themes. WooCommerce plugins that provide ajax results on the search page were my working example.

PR #260, opened 2023-02-15
bug fix to Accordion block

saylorrain:
h tags now match from editor to frontend in the accordion block


saylorrain:
https://app.clickup.com/t/863g27jaj

PR #278, opened 2023-03-10
Adding filter to make sure wp offload media runs on all content that would render css

mark-c-woodard:
addresses: https://app.clickup.com/t/8677kzz4k

Post content that gets passed into css and html build processes that happen after the 'render_callback' entry point are appropriately filtered by wp offload media

but for the row/column layout (and potentially others), they have image attributes that get output through a different process, namely the 'frontend_build_css' function

This function gets block content with the parse_blocks function and its post content straight from the global $post object. Straight from the tap so speak. So wp offload media doesn't have the chance to filter it.

Adding in their custom filter higher up in the chain when we initially grab post content resolves issues where the correct url is not used when outputting these attributes.


kadencewp:
Assuming this works we should remove our check in the row layout block as well:

https://github.com/stellarwp/kadence-blocks/blob/master/includes/blocks/class-kadence-blocks-row-layout-block.php

PR #285, opened 2023-04-04
Fix autofill issue when using basic honeypot

oakesjosh:
According to [this article](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields), most browsers no longer respect the autocomplete="off" attribute. I've implemented the [proposed unofficial approach](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#preventing_autofilling_with_autocompletenew-password) of using `autocomplete="new-password"`, which browsers do not autofill or save. 

Although I was unable to reproduce the autocomplete issue using Chrome or Firefox's native autofill, it did occur when using 1Password and LastPass. I've added the `data-1p-ignore` and `data-lpignore` attributes to address that problem.

I can send a zip of these changes to Karla to determine if this solve the native autofill issue with Chrome.


kadencewp:
I think because we are not doing a deprecation lets ship this today and we can think of what the future of this feature should be.

PR #311, opened 2023-06-22
Repeater Block

mark-c-woodard:
Accompanying blocks free PR for the pro Repeater Block feature.

This currently goes into `threepointone` because that's already been merged into this branch and is more indicative of what this merge will look like going into master after the 3.1 release. 
**I would NOT recommend this be included in the 3.1 release**


src/blocks/column/edit.js
```diff
@@ -137,11 +133,8 @@ function SectionEdit( props ) {
 		} else {
 			addUniqueID( uniqueID, clientId );
 		}
-		const isInQueryBlock = getInQueryBlock( context, inQueryBlock );
-		if ( attributes.inQueryBlock !== isInQueryBlock ) {
-			attributes.inQueryBlock = isInQueryBlock;
-			setAttributes( { inQueryBlock: isInQueryBlock } );
-		}
+
+		setAttributes( { inQueryBlock: getInQueryBlock( context, inQueryBlock ) } );
```
kadencewp, line 137: I have a memory of there being some reason I had to set attributes.inQueryBlock even though I was using setAttributes. Can we leave this like it was just to prevent a previous issue?

kadencewp (changes requested):

PR #324, opened 2023-08-04
Fix for head rendered css with in loop dynamic content

mark-c-woodard:
Addresses: https://app.clickup.com/t/8684xd7qk

An option that will fix dynamic css rendering in the head when the block itself might be in a loop, wnting to use the loop post context.


kadencewp:
We don't need this with our two other fixes to this between PHP filter and the fix for elements. Closing.

PR #326, opened 2023-08-08
AI in the intake form

aykanburcak:
- Implementing "Improve Mission Statement" service to the mission statement text area
- Implementing "Suggest Keywords" service to the keyword selection screen and the Chips Input component
- Design updates on the mission statement progress bar, chips input
- Adding max limit to chips input
- Tooltip feature on the mission statement text area


chumpgrub (commented):
Great work! 

### Improve my Mission Statement
The content getting displayed after triggering this action appears to be scrambled. Here's a screenshot as an example. I copied the example mission statement for Spencer Sharp, clicked the "Improve" button and this was the result: http://p.tri.be/i/M4XNGo

### Suggested Keywords
This feature is really nice, but we should check on the desired user experience for adding/deleting AI-generated keywords. Last I recall, when ‘deleting’ an AI-generated keyword from the active keyword selections, the keyword went back into the pool of AI-generated keywords.

aykanburcak:
@chumpgrub thanks for the great feedback!
After debugging the mission statement issue I saw that the event stream sometimes sends the data as a block like 
```
event: content
data: {"content":""}

event: content
data: {"content":"He"}

event: content
data: {"content":"aling"}
```
instead of one single content as
```
event: content
data: {"content":" Touch"}
```
I have revised the convertStreamDataToJson utility function according to that and it should be working fine now.
Also made another update for adding deleted keyword suggestions back to the suggestions pool.
Screen Capture on 2023-08-09 at 15-12-47

chumpgrub (approved):
Awesome stuff! Thanks for the update!

PR #348, opened 2023-10-12
Inline options for radio buttons and checkboxes in Form(Adv)

laloptk:


src/blocks/advanced-form/fields/radio/edit.js
```diff
@@ -223,6 +223,11 @@ function FieldRadio( { attributes, setAttributes, isSelected, clientId, context,
 								>
 									{__( 'Add Option', 'kadence-blocks' )}
 								</Button>
+								<ToggleControl
+									label={__( 'Show inline?', 'kadence-blocks' )}
```
mark-c-woodard, line 227: I like "display inline?" better for this label.
I think "show", for me, implies revealing something that wasn't there before.

includes/blocks/form/class-kadence-blocks-radio-block.php
```diff
@@ -83,10 +83,14 @@ public function build_html( $attributes, $unique_id, $content, $block_instance )
...
 		$radio_label['inputName'] = 'rb' . $class_id;
-
+		$inline_options = '';
+		if( isset($attributes['inline']) && true === $attributes['inline']) {
+			$inline_options = ' kb-radio-check-items-inline';
+		}
 		$inner_content .= '<fieldset class="kb-radio-check-item-wrap" id="' . $this->field_name( $radio_label ) . '" data-type="radio" data-required="' . $is_required . '" ' . $this->additional_fieldset_attributes( $attributes ) . '>';
 		$inner_content      .= $this->field_legend( $radio_label );
 		$inner_content      .= $this->field_aria_label( $attributes );
+		$inner_content 		.= '<div class="kb-radio-check-items-wrap' . $inline_options . '">';
```
mark-c-woodard, line 93: Is this wrap necessary? 
If it is, then I think we should only wrap if we need to for inline. Only add as much div's as we need to get the job done.

src/blocks/advanced-form/fields/radio/edit.js
```diff
@@ -392,7 +397,7 @@ function FieldRadio( { attributes, setAttributes, isSelected, clientId, context,
 					/>
 
 					{isSelected ?
-						<div className={'kb-form-multi'}>
+						<div className='kb-form-multi' >
```
mark-c-woodard, line 400: Is this change meaningful or just syntax preference?
Not important, just curious.

composer.lock
mark-c-woodard, line 1: I don't know how the rest of the team feels about it, but I like to keep composer / package lock changes out of my commits.
I prefer to let whoever's in charge of the build process set the lock file so I don't accidentally mess with something in the build upstream.

src/blocks/advanced-form/fields/radio/edit.js
laloptk: Non of the above. I was just returning things to its initial state after I realized that was the wrong place to place the code.

includes/blocks/form/class-kadence-blocks-radio-block.php
laloptk: It is only necessary if we want to use flexbox, but yes, I will made the wrap conditionally rendered. Or you think it is better to use `display: inline-block`?

composer.lock
laloptk: Cool, I will place the initial package.lock file in there again. The composer file is the same as it was when I cloned, I was just trying to fix the strauss issue and I missed a change I made when testing.

includes/blocks/form/class-kadence-blocks-radio-block.php
mark-c-woodard: I mean, if inline looks the same / like people would expect, then I'd say go with that because it uses one less container. 
But if flex works better than stick with that.

src/blocks/advanced-form/fields/radio/edit.js
mark-c-woodard: cool

includes/blocks/form/class-kadence-blocks-radio-block.php
laloptk: Got rid of the wrap in the frontend. I decided not to do the same in the backend because it implies to modify the parent div and that is related to all form fields.

composer.lock
laloptk: Done

includes/blocks/form/class-kadence-blocks-radio-block.php
laloptk: Done

src/blocks/advanced-form/fields/radio/edit.js
laloptk: Done

mark-c-woodard:
Okay, Ben pointed out that using margin for the gapping will mean a bit of inaccurate wrapping.
So we definetly want to use flex. 

I tried using flex on the `<fieldset>` element, without an extra container, and it looks like that might work out.
image

PR #355, opened 2023-10-19
Allow decimal numbers in the number field for Advanced Form Block

laloptk:


kadencewp:
Let's make this an option. I can see people wanting a whole number, so lets add a toggle for "allow decimals" in the edit of the field, and then, if enabled, we allow decimals

PR #357, opened 2023-10-24
Add aria label to link when lightbox enabled in Advanced Gallery Block

laloptk:
KAD-425


laloptk:
> It may be preferable to add this in php when we render the block. That would prevent an additional Javascript dependency on the front end.
> 
> I believe this line is where that class in generated in PHP https://github.com/stellarwp/kadence-blocks/blob/master/includes/blocks/class-kadence-blocks-advancedgallery-block.php#L790

laloptk:
@oakesjosh I agree with what you say Josh, but we might want to check that with @kadencewp, he left a comment in the Jira issue saying he wanted me to do it with JS, I just assumed he saw something that might affect other things other than only the link.

PR #370, opened 2023-11-08
Feature/analytics wizard start

dotjon:


includes/analytics/class-analytics.php
```diff
@@ -86,6 +86,7 @@ public static function get_prophecy_token_header( $args = [] ) {
 		$site_name    = get_bloginfo( 'name' );
 
 		$defaults = [
+			'date_time'       => current_time( 'mysql' ),
```
tarecord, line 89: We don't need to explicitly pass the date_time because Mixpanel (and any other analytic software should) automatically handles this.

includes/class-kadence-blocks-prebuilt-library-rest-api.php
```diff
@@ -600,6 +607,46 @@ public function get_remaining_credits( $request ) {
...
+					'industry'          => $event_data['industry'] ?? '',
+					'mission_statement' => $event_data['missionStatement'] ?? '',
+					'keywords'          => $event_data['keywords'] ?? '',
+					'tone'              => $event_data['tone'] ?? '',
+					'collections'       => $event_data['customCollections'] ?? '',
+				];
+				break;
+		}
+
+		if ( $event ) {
```
tarecord, line 641: Let's do a little more nuanced checking here.

ChrisK [shared this article with me](https://www.beberlei.de/post/when_to_use_empty_in_php_i_say_never) a little while ago and while you're not using `empty()` here, I think the suggestions are still helpful.

```php
// Replace
if ( $event ) { }

// With
if ( strlen($string) === 0 ) { }
```

It allows for strict type checking and better communicates the type to whoever is reviewing this in the future.

tarecord (changes requested):
Just some minor things, but really nice work overall 🤘🏼

includes/class-kadence-blocks-prebuilt-library-rest-api.php
```diff
@@ -600,6 +607,46 @@ public function get_remaining_credits( $request ) {
...
+		$event_label = $request->get_param( self::PROP_EVENT_LABEL );
+		$event_data  = $request->get_param( self::PROP_EVENT_DATA );
+		$event_data  = json_decode( $event_data, true );
+
+		$event       = '';
+		$context     = array();
+
+		switch ( $event_label ) {
+			case 'ai_wizard_started':
+				$event = 'ai_wizard_started';
```
tarecord, line 623: Thinking through this a little more, everything I've seen in [Mixpanel's documentation](https://docs.mixpanel.com/docs/data-structure/events-and-properties#name-events-and-properties-consistently) suggests event names like "AI Wizard Started" so let's be consistent with the event names. You're `event_label` here is fine to collect the necessary event data.

includes/class-kadence-blocks-prebuilt-library-rest-api.php
```diff
@@ -600,6 +607,46 @@ public function get_remaining_credits( $request ) {
...
+		$event       = '';
+		$context     = array();
+
+		switch ( $event_label ) {
+			case 'ai_wizard_started':
+				$event = 'AI Wizard Started';
+				break;
+
+			case 'ai_wizard_save':
+				$event = 'AI Wizard Save';
```
tarecord, line 627: Is this supposed to be "AI Wizard Updated"? And are we sending _all_ of the event_data each time, or only the changes?

I'm not totally sure what's passed in through `data` [here](https://github.com/stellarwp/kadence-blocks/blob/35abeb96a6d6358e1d2470b158b7cdf9587e5bf3/src/plugins/prebuilt-library/ai-wizard/hooks/use-database.js#L18).

PR #379, opened 2023-11-28
Style accordion title container to disply images inline

laloptk:
[KAD-1698](https://stellarwp.atlassian.net/browse/KAD-1698)

[KAD-1698]: https://stellarwp.atlassian.net/browse/KAD-1698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/blocks/accordion/editor.scss
+++ b/src/blocks/accordion/editor.scss
@@ -213,6 +213,9 @@ button.components-button.kt-accordion-add.is-primary {
     margin-left: 0;
     margin-right: auto;
 }
+.kt-blocks-accordion-title {
+	display: flex;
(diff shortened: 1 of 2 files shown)


mark-c-woodard:
According to the zendesk task this bug is originally based on (https://ithemeshelp.zendesk.com/agent/tickets/534654) the user fixed and/or moved away from trying to do this inline image thing.

And our title icons feature still looks like it's working fine and inline to me. I don't think we need to make this change.

PR #380, opened 2023-11-28
Ensure all but polyfills are prefixed, move prefixed folder

defunctl:
## Changes
- Ensures all dependencies are prefixed besides symfony polyfills.

## Instructions

1. Delete your root `vendor` and `vendor-prefixed` folders .
2. Run `composer install`.

@kadencewp run though your normal tests for AI functionality before merging just to make sure it's all good.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ zip/
 bin/strauss.phar
 .idea
 /vendor/
-/vendor-prefixed/
 bin/pup.phar
 bin/wp-cli.phar
 .pup-build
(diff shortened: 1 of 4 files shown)


kadencewp:
This seems fine, my only question is why move vendor-prefix into vendor folder? Is that just to clean things up?

defunctl:
> This seems fine, my only question is why move vendor-prefix into vendor folder? Is that just to clean things up?

That's the original way Matt sets up projects but I had changed it before when I was debugging something and forgot to change it back.

But yeah, just keeps the root folder a bit cleaner.

PR #381, opened 2023-11-28
Remove skip-lazy class when lazy load enabled

laloptk:
[KAD-1700](https://stellarwp.atlassian.net/browse/KAD-1700)

[KAD-1700]: https://stellarwp.atlassian.net/browse/KAD-1700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kadencewp:
The ticket is not correct. Skip lazy class is added intentionally to prevent third party plugins from breaking the carousel

PR #385, opened 2023-12-04
Add Events for AI Inline Complete

tarecord:
This adds functionality to capture events when users request AI content within the **Text (Adv)** block.

## Within Kadence Blocks
Markup 2023-12-04 at 16 41 52

## Within Mixpanel
Markup on 2023-12-04 at 16:42:56

### DIFF
--- a/includes/class-kadence-blocks-ai-events.php
+++ b/includes/class-kadence-blocks-ai-events.php
@@ -224,9 +224,8 @@ class Kadence_Blocks_AI_Events {
 			case 'ai_inline_completed':
 				$event   = 'AI Inline Completed';
 				$context = [
-					'page_id'        => $event_data['page_id'],
-					'page_title'     => get_the_title( $event_data['page_id'] ),
 					'tool_name'      => $event_data['tool_name'],
+					'type'           => $event_data['type'],
 					'initial_text'   => $event_data['initial_text'],
 					'result'         => $event_data['result'],
 					'credits_before' => $event_data['credits_before'],
--- a/includes/class-kadence-blocks-ai-events.php
+++ b/includes/class-kadence-blocks-ai-events.php
@@ -221,6 +221,18 @@ class Kadence_Blocks_AI_Events {
 					'pattern_categories' => $event_data['categories'] ?? [],
 				];
 				break;
+			case 'ai_inline_completed':
+				$event   = 'AI Inline Completed';
+				$context = [
+					'page_id'        => $event_data['page_id'],
+					'page_title'     => get_the_title( $event_data['page_id'] ),
+					'tool_name'      => $event_data['tool_name'],
+					'initial_text'   => $event_data['initial_text'],
+					'result'         => $event_data['result'],
+					'credits_before' => $event_data['credits_before'],
+					'credits_after'  => $event_data['credits_after'],
+					'credits_used'   => $event_data['credits_used'],
+				];
 		}
 
 		if ( strlen( $event ) !== 0 ) {
--- a/src/blocks/advancedheading/ai-text/ai-text.js
+++ b/src/blocks/advancedheading/ai-text/ai-text.js
@@ -36,6 +36,7 @@ import {
 	toggleFormat,
 	applyFormat,
 } from '@wordpress/rich-text';
+import { sendEvent } from '../../../extension/analytics/send-event';
 const name = 'kadence/ai-text';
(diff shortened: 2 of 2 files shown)


includes/class-kadence-blocks-ai-events.php
```diff
@@ -221,6 +221,18 @@ public function handle_event_endpoint( WP_REST_Request $request ): WP_REST_Respo
 					'pattern_categories' => $event_data['categories'] ?? [],
 				];
 				break;
+			case 'ai_inline_completed':
+				$event   = 'AI Inline Completed';
+				$context = [
+					'page_id'        => $event_data['page_id'],
+					'page_title'     => get_the_title( $event_data['page_id'] ),
```
kadencewp, line 228: We don't need this

includes/class-kadence-blocks-ai-events.php
```diff
@@ -221,6 +221,18 @@ public function handle_event_endpoint( WP_REST_Request $request ): WP_REST_Respo
 					'pattern_categories' => $event_data['categories'] ?? [],
 				];
 				break;
+			case 'ai_inline_completed':
+				$event   = 'AI Inline Completed';
+				$context = [
+					'page_id'        => $event_data['page_id'],
```
kadencewp, line 227: We don't need this

src/blocks/advancedheading/ai-text/ai-text.js
```diff
@@ -81,6 +82,7 @@ export const AIText = {
 		const activeStorage = SafeParseJSON( localStorage.getItem( 'kadenceBlocksPrebuilt' ), true );
 		const savedCredits = ( undefined !== activeStorage?.credits && '' !== activeStorage?.credits && null !== activeStorage?.credits ? activeStorage.credits : 'fetch' );
 		const currentCredits = ( '' !== credits ? credits : savedCredits );
+		const currentPostId = parseInt((new URLSearchParams(window.location.search)).get('post')) ?? 0;
```
kadencewp, line 85: This won't work in widget areas, customizer, FSE, lets just not worry about capturing that in the events as we don't need it

src/blocks/advancedheading/ai-text/ai-text.js
```diff
@@ -173,16 +189,28 @@ export const AIText = {
 							setTempCredits(parseInt( currentCredits ) - 1);
 							setCredits( 'fetch' );
 							setIsOpen( true );
+							sendEvent('ai_inline_completed', {
+								page_id: currentPostId,
+								page_title: '',
+								tool_name: AIText.title,
+								initial_text: '',
```
kadencewp, line 196: Lets make sure we capture the initial text

src/blocks/advancedheading/ai-text/ai-text.js
```diff
@@ -211,11 +240,22 @@ export const AIText = {
 							setIsOpen( true );
 							setTempCredits(parseInt( currentCredits ) - 1);
 							setCredits( 'fetch' );
+							sendEvent('ai_inline_completed', {
+								page_id: currentPostId,
+								page_title: '',
+								tool_name: AIText.title,
+								initial_text: '',
```
kadencewp, line 247: Lets make sure we capture the initial text

src/blocks/advancedheading/ai-text/ai-text.js
```diff
@@ -211,11 +240,22 @@ export const AIText = {
 							setIsOpen( true );
 							setTempCredits(parseInt( currentCredits ) - 1);
 							setCredits( 'fetch' );
+							sendEvent('ai_inline_completed', {
+								page_id: currentPostId,
+								page_title: '',
+								tool_name: AIText.title,
```
kadencewp, line 246: We should add capturing what is happening, in this case, the type

kadencewp (commented):
Thanks, overall seems like a good add. I would like to request that we change some things in the event. 

I think it's best to understand what the request is, so can capture the request type. For example generating content, editing content, making shorter, making longer, improving, etc. 

I would also think we should capture the initial text when transforming. I see you have a place for it but are not adding the data.

The code you have to capture the ID won't work in widget areas, or FSE areas so I think you can leave that off. It's not really relevant data anyway. Similarly the page title (which won't work in widget areas, etc). None of that is needed to I suggest we just remove it from what gets sent to the events.

PR #397, opened 2023-12-13
Design library style filters

aykanburcak:
- Adds style filters into the pattern list in the design library
http://p.tri.be/i/2GDScH


kadencewp:
Thanks!

PR #399, opened 2023-12-15
Add default colors for icon list items

laloptk:
[KAD-1774](https://stellarwp.atlassian.net/browse/KAD-1774)


kadencewp:
@laloptk  and @oakesjosh This is too opinionated, the list should inherit colors set based on its container and this breaks the link styling in the block itself.  I'm reversing this. Let's solve for KAD-1774 a different way.

PR #400, opened 2023-12-15
Prevent unit options from being changed in mobile and desktop (Adv Form line height and letter spacing help text)

laloptk:
[KAD-1594](https://stellarwp.atlassian.net/browse/KAD-1594)

[KAD-1594]: https://stellarwp.atlassian.net/browse/KAD-1594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mark-c-woodard:
https://www.loom.com/share/084c31d064eb451291e6cb724f540739?sid=f8c63e30-ab48-4b8f-b909-69fad094c4c1

laloptk:
The `render_typography` method was only rendering letter spacing units with the `letterSpacingType` attribute and the form is passing the 'letterType' attribute, so units are not rendering correctly, as the form attributes come from post meta, the data is saved in the DB and if I change the name of the attribute in the form block to `letterSpacingType` I will orphan the custom field and that will break the letter spacing (and other styles) of the existing forms, I decided to make the method to accept both `letterSpacingType` and `letterType`, sacrificing consistency a bit, but making the fix compatible with what the users using the plugin already have.

PR #401, opened 2023-12-15
Draft for using bun instead of npm

borkweb:
This is just a draft PR so folks can test to see if [bun](https://bun.sh) works. The build is fast! I haven't tested functionality.

For Kadence Blocks, on my local machine, `npm install` and `npm run build` takes 58s
`bun install` and `bun run build` takes 32s

The webpack part of run build takes 28s for each approach


.puprc
```diff
@@ -1,14 +1,16 @@
 {
 	"build": [
+		"rm -rf vendor",
```
borkweb, line 3: This is added to clean up the vendor dir so Strauss works cleanly each time.

PR #408, opened 2024-01-02
Add Collection Updated event

tarecord:
This adds a new event: "Collection Updated"

Any time a user visits the "Update Design Library Images" and clicks the "Update My Design Library" button, this event is sent to Mixpanel.

Screen Capture on 2024-01-02 at 11-00-45


src/plugins/prebuilt-library/ai-wizard/kadence-ai-wizard.js
```diff
@@ -134,9 +134,12 @@ export function KadenceAiWizard( props ) {
...
 		// Submit wizard data on finish buttton click.
 		if (event.type === 'click' && event.target.classList.contains('components-wizard__primary-button')) {
-			handleSave();
 
-			if ( ! photographyOnly ) {
+			if (photographyOnly) {
+				handleEvent('collection_updated');
+				onWizardClose();
+			} else {
+				handleSave();
```
kadencewp, line 142: By moving the handle save here doesn't that make it so any photo change won't be saved?

tarecord: Ah, you're right. I was trying to isolate the event. I've moved it outside that if block so it's handled in both cases (Photo collection updates _and_ Wizard data updates).

PR #420, opened 2024-01-22
[KAD-1959]: AI Content request/response refactor + "Context Generation" event updates

defunctl:
### Main Changes

> ⚠️ I'm not 100% sure how to test all the functionality that this affects as the `get_remote_contents()` method is called in many places in this file, but everything should work exactly as it did before these code changes.

- Refactors `get_remote_contents()` to return the raw response array/WP_Error, so we can decide contextually what to do with the specific responses.
- `Context Generation Failed` event now properly attaches response errors.
- **Removed** a second `Context Generation Failed` which will now be tracked in `prophecy-wordpress` with much more detail.
- **Renames** the event `mission_statement` property to `description` as per Donaldo's request.
- Some code cleanup/simplification, docBlock + type updates etc...


kadencewp:
This looks great!

PR #423, opened 2024-01-25
Use the toc attribute if it exists instead of the heading content in …

at-benni:
…the Table of Contents

Hello, with this users can set the custom attribute toc to a heading and the table of contents will use it instead of the heading content.
Like this: `<h* toc="Short title for TOC">Long title that is displayed in the page but looks stupidly long in the toc</h*>`
https://wordpress.org/support/topic/alternate-title-for-toc/


benni1516:
This feature does not alter any functionality for existing users but it does add the possibility for users who want to use a feature like this to do it.

kadencewp:
Thanks for this; I think we'll need to make the attribute data-alt-title (data so it's valid HTML) and then perhaps add an option in the advanced text block to put content into that attribute through a text box in the advanced settings.

benni1516:
Hello! Thanks for your message, I have changed the attribute name from toc to data-alt-title, adding an option to the block editor would be very cool and would probably add so much value to the seo/ux aspect of the toc block, I would love tho if this small change to merge in advance.

kadencewp:
Thanks!

at-benni:
Thanks for adding this!

gilbert-hernandez:
https://stellarwp.atlassian.net/browse/KADTRIA-569

PR #427, opened 2024-01-31
[KAD-1961]: Properly track collection_updated events

defunctl:
- If a collection wasn't custom (aka premade), it would be an empty string in the event payload.
- This event would previously send the user's old collection, because it was firing before saving was complete.
- This properly tracks changed collections each time a user changes it either through the Wizard, or through: _Design Library > Settings Icon > Update Design Library Images_.
- The event will now tell us if it's a custom collection.
- I wasn't sure why [handleSave() ](https://github.com/stellarwp/kadence-blocks/compare/bugfix/KAD-1961/collection-updated-event?expand=1#diff-3f260d3b955791cd26c02f4447e0032530d5546fa94b033c702d86704702b031L146)is called even after [handleComplete() ](https://github.com/stellarwp/kadence-blocks/compare/bugfix/KAD-1961/collection-updated-event?expand=1#diff-3f260d3b955791cd26c02f4447e0032530d5546fa94b033c702d86704702b031L142) is, both functions which save the AI data. Let me know if this change was wrong? I was seeing duplicate `Saved` notices come up.

### DIFF
--- a/includes/class-kadence-blocks-ai-events.php
+++ b/includes/class-kadence-blocks-ai-events.php
@@ -228,10 +228,8 @@ class Kadence_Blocks_AI_Events {
 				];
 				break;
 			case 'collection_updated':
-				$event = 'Collection Updated';
-				$context = [
-					'collection_name' => $this->get_custom_collection_name_by_id( $event_data['customCollections'], $event_data['photoLibrary'] ),
-				];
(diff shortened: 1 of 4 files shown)


kadencewp (commented):
All makes sense to me

PR #435, opened 2024-02-13
When the attribute data-add-anyway is set ignore the allowed levels a…

at-benni:
…nd add anyway

Hello :)
It would be very practical to allow for including individual headings which level is not in the allowed levels with an html attribute.

If I set a heading to example <h3 data-add-anyway"true"> the heading is included even tho the allowed levels do not allow h3's


at-benni:
Thanks @oakesjosh very cool idea, I refactored the code a bit so I don't have to make the same change 6 times and I added the possibility to add/remove individual headings

at-benni:
@oakesjosh I added your idea, could you consider adding this feature? :)

at-benni:
thanks for this cool feature @oakesjosh :)

PR #436, opened 2024-02-16
Prevent evergreen hours and minutes from being undefined

laloptk:
[KAD-1385](https://stellarwp.atlassian.net/browse/KAD-1385)


laloptk:
I broke the issue in two, since I have to work in point number two in the ticket. Now is ready for review. @oakesjosh

PR #448, opened 2024-03-15
Bugfix / Row layout default column gap

laloptk:
[KAD-2061](https://stellarwp.atlassian.net/browse/KAD-2061)

[KAD-2061]: https://stellarwp.atlassian.net/browse/KAD-2061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/blocks/rowlayout/edit.js
+++ b/src/blocks/rowlayout/edit.js
@@ -1572,7 +1572,7 @@ function RowLayoutEditContainer(props) {
 						)}
 					{columns &&
 						columns === 3 &&
-						'grid-layout' !== previewLayout && 
+						'grid-layout' !== previewLayout &&
 						'default' !== columnGutter &&
 						('Desktop' === previewDevice ||
 							('Tablet' === previewDevice && tabLayoutClass === 'inherit')) && (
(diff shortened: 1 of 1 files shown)


mark-c-woodard:
looks good, just fix the lint

mark-c-woodard:
Actually we need to take another look at this.
According to the zendesk ticket, it's the frontend styles that aren't correct when the column gutter is default. You make the backend match the frontend in this commit.

The problem was that the frontend wasn't doing that minmax / calc thing because it sees the attribute as empty.

See:
https://ithemeshelp.zendesk.com/agent/tickets/556855

kadencewp:
This issue was resolved in 3.2.22 https://github.com/stellarwp/kadence-blocks/commit/42562e72f62dad508f1460e7077f9b8d661ae032

It is also important to note that if you pushed this in, custom column widths would not work in the editor if the gap was set to default. Please review your code changes in more than one use case so that you understand what they affect.

PR #464, opened 2024-04-12
Break words in row layout columns

laloptk:
[KAD-2322](https://stellarwp.atlassian.net/browse/KAD-2322)


kadencewp:
I don't think we want to add this to Kadence Blocks, this is specific to the theme and not blocks so I think we should continue to control it on the theme side.

PR #486, opened 2024-04-30
Enhancement/progressbar type defaults

laloptk:
[KAD-2560](https://stellarwp.atlassian.net/browse/KAD-2560)
...

### Issue Info

### Checklist

- [x] No.


laloptk:
Hey Josh, I realized using the `barType` variable just complicates the code and gives the same functionality, since with barType we have to take into account every time the progressbar type changes (line->line-mask, line->circle, circle->semicircle, etc), while when using the `key` variable the only thing that we have to know is if it is a 'lined-mask' or not.

PR #490, opened 2024-05-06
Feature: add health check

defunctl:
### Main Changes

1. We no longer use the Monolog error logger if the `error_log` function is disabled, preventing a fatal error for an undefined function.
2. Adds a Health Check system, with a PHP function checker, this way, logs aren't just silently writing no where when debugging needs to happen and will warn the user.
3. If any PHP functions we define are disabled, notices will appear on the Kadence Settings Page, which support can now ask users to visit to verify there are no notices displayed there.

> Note: If you have better wording for the notices, please let me know.

image

🎫  #[Jira Ticket]

https://ithemeshelp.zendesk.com/agent/tickets/589622

### Issue Info

### Checklist

- [x] No.

### DIFF
--- a/includes/resources/App.php
+++ b/includes/resources/App.php
@@ -5,6 +5,7 @@ namespace KadenceWP\KadenceBlocks;
 use InvalidArgumentException;
 use KadenceWP\KadenceBlocks\Adbar\Dot;
 use KadenceWP\KadenceBlocks\Cache\Cache_Provider;
+use KadenceWP\KadenceBlocks\Health\Health_Provider;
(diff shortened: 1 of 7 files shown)


defunctl:
@oakesjosh @kadencewp a few more questions:

1. Should I provide a filter so these can be disabled, it would make determining support issues a lot more complicated, but some people might get annoyed by this?
2. I now think they should be dismissible at least, I'll make that change quickly.

kadencewp:
I don't think a filter is needed, looks great

PR #501, opened 2024-06-13
bugfix/KAD-2883

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KAD-2883](https://stellarwp.atlassian.net/browse/KAD-2883)
...

### Issue Info

### Checklist

- [x] No.

### DIFF
--- a/src/blocks/tabs/save.js
+++ b/src/blocks/tabs/save.js
@@ -34,6 +34,9 @@ function KadenceTabsSave({ attributes }) {
 		widthType,
 		tabWidth,
 		linkPaneCollapse,
+		size,
+		tabSize,
+		mobileSize
 	} = attributes;
 	const layoutClass = !layout ? 'tabs' : layout;
 	const tabLayoutClass = !tabletLayout ? 'inherit' : tabletLayout;


mark-c-woodard (changes requested):
This did not fix the issue. Check the saved block attributes for size, tabSize, mobileSize and you'll see that they aren't saving when you use number values.
You can check what the block has for it's saved attributes in a few ways. console log the attribute in the edit function, use a plugin like block x-ray attributes, or check the post in the database.

It's weird, but the save function isn't in charge of saving attributes, just block html content. So your changes in that file will not affect the attributes.

gilbert-hernandez:
The font size now saves as a string.

gilbert-hernandez:
From Josh: 

Since lineHeight and Size are using the same save function, I don’t think you can convert them to a string because lineheight is a number
		"mobileSize": {
			"type": "string"
		},
		"mobileLineHeight": {
			"type": "number"
		},
May have to duplicate the function for size and only convert the size values to a string

PR #503, opened 2024-06-18
KAD-2680 splide carousel gap unit fixed

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KAD-2680](https://stellarwp.atlassian.net/browse/KAD-2680)
...

### Issue Info

### Checklist

- [x] No.


mark-c-woodard (commented):
This works great and the following is all very nitpicky.
I think the conditional would be better as `key.includes('sliderGap')`.
I think it would be better combined with the numeric check as well. Because it's trying to avoid getting treated like a number when it needs to be processed like a string.
So maybe something like `if (!Number.isNaN(parsedInt) && ! key.includes('sliderGap'))`

gilbert-hernandez:
removed my extra conditional and integrated it with the one that was existing.

PR #506, opened 2024-06-27
Bugfix/kad 2894

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KAD-2894](https://stellarwp.atlassian.net/browse/KAD-2894)
...

### Issue Info

### Checklist

- [x] No.


mark-c-woodard (changes requested):
It's very uncommon to need changes to vendor files and usually not recommended. It's kind of similar to editing plugin code on a site. If this library ever gets updated, this change would be lost. We also have very little insight into how this library works so unexpected consequences are more likely.

Anyway, I think you can make this fix without it. Looks like the library supports the options you're trying to add https://github.com/nk-o/jarallax?tab=readme-ov-file#options
So it's just a matter of getting the attributes your already adding formatted the "Data attribute way". Looks like `data-img-size` etc should work.

Also add a `!empty($attributes['bgImgAttachment'])` to the top conditional.

PR #509, opened 2024-06-28
KAD-2926 height/width removed from kadence-info-box-image-intrisic

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KAD-2926](https://stellarwp.atlassian.net/browse/KAD-2926)
...

### Issue Info

### Checklist

- [x] No.


mark-c-woodard (changes requested):
Those height and width properties have been there for a long time, 2+ years according to the blame. I'm worried removing them will have some unintended side effect.

I think just ensuring the block does not break out of it's container with max-width:100% will be a safer choice.

image

PR #515, opened 2024-07-16
Fix String Translation for admin-kadence-home in class-kadence-blocks-settings.php

DAnn2012:
Fix String Translation for `admin-kadence-home` in `class-kadence-blocks-settings.php`


DAnn2012:
Hello @oakesjosh 

Please could you take a look at this PR of mine?

Thanks.

PR #519, opened 2024-07-22
Captcha not validating when inserted into a form and relying on global kadence_blocks_settings keys

oakesjosh:
When `useKbSettings` was added the default value was set to true, but the attribute value isn't passed in the data if it's still the default value. On this line we only check if it's set and true, when we should also assume it's true when not set.

If this is set to false, we don't pull in the global keys and thus don't validate the form when these are otherwise set.

Mostly PRing for a second check to make sure that `using_kadence_blocks_settings` defaulting to true won't mess anything up.

https://github.com/stellarwp/kadence-blocks/commit/ca679a58dfce3dcf3c66a416bbeba0f5a8df5aa9#diff-2d4af50a9a461b86a01d175bdb2480e9c66bed9a9aaae021d98c559d27cdd153R121-R128

### DIFF
--- a/includes/advanced-form/advanced-form-captcha-settings.php
+++ b/includes/advanced-form/advanced-form-captcha-settings.php
@@ -119,7 +119,7 @@ class Kadence_Blocks_Form_Captcha_Settings {
 	}
 
 	private function is_using_kadence_blocks_settings( $attributes ) {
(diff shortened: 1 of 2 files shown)


mark-c-woodard:
might affect people who have been relying on the setting not working correctly, but they'll just need to set the setting to false how it should be. Looks good.

mark-c-woodard (approved):

PR #522, opened 2024-08-06
Bugfix/kad 3058

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KAD-3058](https://stellarwp.atlassian.net/browse/KAD-3058)
...

### Issue Info

### Checklist

- [x] No.

### DIFF
--- a/src/blocks/rowlayout/edit.js
+++ b/src/blocks/rowlayout/edit.js
@@ -724,7 +724,7 @@ function RowLayoutEditContainer(props) {
 	const hasBG = bgColor || bgImg || gradient || overlay || overlayGradient || overlayBgImg ? 'kt-row-has-bg' : '';
 	const isKadenceT = typeof kadence_blocks_params !== 'undefined' && kadence_blocks_params.isKadenceT ? true : false;
 	const paddingSidesTheme = isKadenceT && true === inheritMaxWidth ? 'var(--global-content-edge-padding)' : '0px';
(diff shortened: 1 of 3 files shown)


gilbert-hernandez:
The padding controls now show "sm" if there is a BG and there aren't L/R padding set. The front end used the customizer content edge while the backend used the "sm" space. I updated the front end to use the "sm" spacing as well. Otherwise, I would need to add to the src/blocks/rowlayout/constants.js file to show something custom. 

Instead, now a background adds a "sm" L/R padding by default (for front and back), and the padding controls show it in the block settings.

mark-c-woodard:
Good idea following some of the logic already on that component for placeholders top/bottom values.

Since the bg only applies left/right padding, we need to respect the values already in place for top/bottom though. So I updated the padding placeholders to use the set top/bottom values.

I also reverted the frontend change. I don't think we want to change people's frontends like that. 
Also, looking at the code surrounding this bg color padding in the background, it reads like a compromise. 
Like we actually want to set `--global-content-edge-padding` when a bg color is applied, but it's really inconvenient to apply that value in the backend. The spacing visualizers and other components can't really handle a value like `var(--global-content-edge-padding)` so I think `sm` was the closest thing that made everything play nice. Not exactly true to frontend, but a compromise.

So long story short I think we want to match the frontend. It's just too inconvenient, so `sm` is close enough.

PR #571, opened 2024-09-11
KADTRIA-1286 updated wpml config for auto translate adv text

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KADTRIA-1286](https://stellarwp.atlassian.net/browse/KADTRIA-1286)
...

### Issue Info

### Checklist

- [x] No.

### DIFF
--- a/wpml-config.xml
+++ b/wpml-config.xml
@@ -28,7 +28,7 @@
 			<key name="title" />
 		</gutenberg-block>
 		<gutenberg-block type="kadence/advancedheading" translate="1">
-			<xpath>//div//*[self::h1 or self::h2 or self::h3 or self::h4 or self::h5 or self::h6 or self::p or self::div or self::span.wp-block-kadence-advancedheading]</xpath>
+			<xpath>//*[self::h1 or self::h2 or self::h3 or self::h4 or self::h5 or self::h6 or self::p or self::div or self::span.wp-block-kadence-advancedheading]</xpath>
 		</gutenberg-block>
 		<gutenberg-block type="kadence/tabs" translate="1">
 			<xpath>//span[@class="kt-title-text"]</xpath>
@@ -112,6 +112,6 @@
 					<key name="caption" />
 				</key>
 			</key>
-		</gutenberg-block>		
+		</gutenberg-block>
 	</gutenberg-blocks>
 </wpml-config>


Pawel-H-OTGS:
Hey, I'm a developer at https://onthegosystems.com/ and I'm working on https://wpml.org/

I don't have access to the jira ticket and I don't know what the content is, but during our internal tests it turned out that your change causes an incorrect "capture" of the test for translation (instead of just the text, the entire HTML block is captured)

image

The additional selected HTML block goes into translation, and only the text at the bottom should remain (we don't want to edit the HTML during translation) - so the entire HTML block is used instead of just the isolated text

gilbert-hernandez:
Hello @Pawel-H-OTGS ,

The change was made to help facilitate Deepl auto translation. Before the change, Text (Adv) blocks were not auto-translated with Deepl. Your screenshot shows that WPML provides the entire HTML block for translation. Is this correct? Does this happen for all Text (Adv) blocks or only certain ones? Is there a way you can give a video showing more information? As it is a third-party premium plugin, access to testing is sometimes difficult.

Pawel-H-OTGS:
Hello @gilbert-hernandez 

Sorry for writing back so late, but I had to find out a few things on our end.

First of all, thank you for your quick response.

> Your screenshot shows that WPML provides the entire HTML block for translation. Is this correct? Does this happen for all Text (Adv) blocks or only certain ones?

the xpath rule changed in this commit only applies to this one specific block visible in the screenshot.

For modifying these rules we have more detailed documentation available here:
- https://wpml.org/documentation/support/language-configuration-files/make-custom-gutenberg-blocks-translatable/

In this case, the most important thing is that the rule should be prepared in such a way that it "catches" only "innerHTML" with only string/text - it usually should not contain any HTML elements (such as `<div>`)

> In this case, the most important thing is that the rule should be prepared in such a way that it "catches" only "innerHTML" with only text - it usually should not contain any HTML elements (such as `<div>`)

I don't know if a video would be useful, because the problem is in adjusting the xpath - the video wouldn't differ much from the screenshot, it would be best if you could recreate it yourself in your environment

I don't know if you have access to WPML, if so I can prepare a step-by-step description of how to recreate the result on the screenshot - so you can check any changes/adjustments

---

It seems to us that the previous xpath formula was correct and did not require modification (it correctly captured the text itself).

Could I ask you to tell us know what problem you tried to solve with this fix, with a possible description of how to reproduce the problem/effect (+ what plugins are active and in what version).
Maybe there is something that causes compatibility issues and we could check it for ourselves - what made you have to change this xpath rule

Pawel-H-OTGS:
Hi @gilbert-hernandez 
are there any steps planned in this regard? I would like to be able to answer customer questions as accurately as possible

Best regards

PR #580, opened 2024-09-27
KAD-3459 Prevent flashing of design library in Lifter LMS

oakesjosh:
This prevents the flashing of the icon when editing an Lifter LMS course. Both us and lifter are trying to inject content, using the same method, and are overwriting each others buttons. I tried modifying the selector to make sure we're in the same div, different divs, etc, but switching to a mutation observer seems to be the most reliable way to prevent this from happening. Open to other ideas though

https://github.com/user-attachments/assets/8a12f889-1ed2-4d6b-97d6-aa3d69b7cff6

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -58,6 +58,7 @@ const eslintConfig = {
 		Masonry: 'readable',
 		IntersectionObserver: 'readable',
 		getComputedStyle: 'readable',
+		MutationObserver: 'readonly',
 	},
 	rules: {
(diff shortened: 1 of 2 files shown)


mark-c-woodard:
It worked for me too, but I am curious why. We’re still putting the button in the same place we were before. so why isn’t lifter just deleting our button like it did before?

mark-c-woodard:
@oakesjosh you still want to push this forward or is this irrelevant now?

PR #609, opened 2024-11-14
KAD-3806 Show the max progress as the initial state

oakesjosh:
Thoughts on changing this to the default behavior? 

On page load progressMax will show as the value until JS is executed. This prevents issues where the JS never inits and it looks like you have "0% uptime" or "0 sales"


mark-c-woodard:
The downside of showing max first is potential stuttering on load right? Because it's going to jump from max% down to zero when the js loads before it starts progressing up.

I think this is a good default with the option to turn it off for people who run into the stuttering. I imagine it also shows the more relevant max percent to search engines as well when they crawl.

It's looks like you've set the default to true in blocks.json, but the frontend implementation only works if it's actually been assigned. So it won't take effect until you toggle it off/on.

mark-c-woodard (approved):

PR #617, opened 2024-11-20
bugfix/KAD-3784

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KAD-3784](https://stellarwp.atlassian.net/browse/KAD-3784)
...

### Checklist


gilbert-hernandez:
Hmm... Forgot frontend

gilbert-hernandez:
This fix adds a conditional to the border radius for the image. If there is a border radius without padding, the radius is only added to the media container. This fixes the problem described in JIRA. However, if there is padding on any side, the border radius is also added to the image. This is so  the image gains border radius and the corners aren't visible when there is padding.

PR #618, opened 2024-11-25
Update dependences, register block metadata,  & use components, icons, and helpers from repos

oakesjosh:
This shouldn't be merged into master yet -- Opening a PR for dev testing, then we can send to support for testing. Not trying to tie this to a specific release or date.

This can be a pain to switch back and forth locally. A `rm -rf node_modules` and fresh install is needed. We no longer need the `--legacy-peer-deps` when doing an npm install though. Then install the submodules `git submodule update --init --recursive`. To switch back to a previous branch you'll have to delete the 3 submodule directories.  `rm -rf src/packages/{components,helpers,icons}`

### What this does 

- Replace `dr-kobros/react-webfont-loader` in the KadecnceWebFontLoader and FontLoader components.
- Replace `react-container-dimensions` with similar function named ContainerDimensions in [column-drag-resizer/index,js](https://github.com/stellarwp/kadence-components/blob/free-to-repo/src/column-drag-resizer/index.js) component.
- Update `react-select`
- Update `@wordpress/i18n`
- Update `@wordpress/scripts`
- Generally updated dependences

- Update the gulp files to not call deprecated functions
- Update scss files processed by gulp to use `@use` instead of deprecated `@import`.

### Support new 6.7 block manifest metadata registration
With the updated @wordpress/scripts we can use the new  wp-scripts build-blocks-manifest. This is registered in kadence-blocks.php. This improves performance since PHP only has to read and parse a single file from the server to get all block metadata. When attempting the register the blocks later, i'll check if it's already pulled in the metadata before opening the blocks specific block.json.

### Moving packages to repos
- Delete existing /src/packages/{components, icons, helpers}
- Add submodule for components, icons, helpers

### DIFF
--- a/kadence-blocks.php
+++ b/kadence-blocks.php
@@ -141,7 +141,7 @@ function kadence_blocks_init(): void {
 	) {
 		wp_register_block_metadata_collection(
 			KADENCE_BLOCKS_PATH . 'dist/blocks/',
-			KADENCE_BLOCKS_PATH . 'dist/blocks/manifest.php'
+			KADENCE_BLOCKS_PATH . 'dist/blocks/blocks-manifest.php'
 		);
 	}
 


mark-c-woodard:
Did not see any issues walking around the block editor, a few different blocks, a slew of settings, just using free, free and pro, the header and a few frontend pages.

I think this should move forward.

mark-c-woodard (approved):

mark-c-woodard:
Actually did find one thing
image
image

Looks like the flex settings on the section block are wonky. 

I also get an error when trying to use the reset on the vertical gap setting there.

mark-c-woodard (commented):
see previous comment

PR #626, opened 2024-12-05
bugfix/KAD-3677

gilbert-hernandez:
🎫  #[Jira Ticket]

...

### Checklist


gilbert-hernandez:
The initial issue was that the slider/carousel galleries weren't showing the shadow the same as the other gallery types. However, there is a note that the sliders and carousels will cutoff shadows due to the way they are hidden. But, I noticed that the thumbnails in the thumbnail sliders weren't showing the shadow on the frontend. This fix is to show the shadows on thumbnails in the slider.

PR #637, opened 2024-12-31
bugfix/KAD-3933

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KAD-3933](https://stellarwp.atlassian.net/browse/KAD-3933)
...

### Checklist


gilbert-hernandez:
Is the placement of the new code ok? The problem is that the render_row_gap_property() function doesn't return the correct value when the tabletGutter and mobileGutter attributes are empty but columnGutter isn't. Ideally, the tablet and mobile gutters should inherit, but the render_row_gap_property() function returns a different result. This fix adds tabletGutter mobileGutter if columnGutter is available.

gilbert-hernandez:
@oakesjosh

Moved the fix to the CSS class function as we discussed. Also, made some updates as the previous fix didn't consider custom gap values.

PR #643, opened 2025-01-06
bugfix/KAD-3949

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KAD-3949](https://stellarwp.atlassian.net/browse/KAD-3949)
...

### Checklist


gilbert-hernandez:
I opted to add the $attributes['textShadow'] array if it was empty instead of messing with the other conditionals. This fix ensures that the textShadow attribute exists and has values if the default values aren't changed and the option for the text shadow is enabled.

PR #646, opened 2025-01-09
audit/KAD-4008

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KAD-4008](https://stellarwp.atlassian.net/browse/KAD-4008)
...

### Checklist


gilbert-hernandez:
This adds the em/rem sizes to the icon border and border radius. It also updates the .5 step to .1 for the stroke width.

gilbert-hernandez:
Image Border now uses em/rem

Image Border Radius now uses em/rem

Number Size now uses em/rem

Number Border uses em/rem

Number Border Radius uses em/rem

Fixed an issue where the numbers didn’t show in frontend.

gilbert-hernandez:
@oakesjosh 

I removed the changes to the save.js file and made a new ticket for that particular problem. 

Also, I updated the checks for the mediaIcon[0].borderRadiusUnit for the max attribute in some of the settings as we discussed. The ?. operator returns undefined if the value isn't there. In this case, if the value is undefined, the max needs to be for a pixel value. So, if it is undefined (because it didn't exist before the update) the max should be the pixel value instead of the EM/REM value. 

In other places, I used the ?? operator, so I think these changes should avoid runtime errors and still provide a truthy response for backward compatibility. If a user has already set a size before the update, they should still be able to update the slide without running into a problem where the max value is 12px.

PR #649, opened 2025-01-13
bugfix/KAD-3960

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KAD-3960](https://stellarwp.atlassian.net/browse/KAD-3960)
...

### Checklist

### DIFF
--- a/src/blocks/tableofcontents/style.scss
+++ b/src/blocks/tableofcontents/style.scss
@@ -296,3 +296,7 @@
 	letter-spacing: inherit;
 	font-family: inherit;
 }
+// sub-menu list style to match editor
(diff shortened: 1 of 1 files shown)


gilbert-hernandez:
@oakesjosh 
I updated the backend to match the frontend instead. I then realized that there are already options to "Select a List Style" on the block toolbar in Gutenberg. There are settings for "Numbered", "Bullets", and "None". If we wanted to add additional styles, I would need to import one of the Kadence icons like the other options, and add CSS classes for the submenus. 

However, the updated changes move the CSS to the editor so that it will display as disc by default instead of reverting.

PR #651, opened 2025-01-14
bugfix/KAD-3599

gilbert-hernandez:
🎫  #[Jira Ticket]
[https://stellarwp.atlassian.net/browse/KAD-3599](https://stellarwp.atlassian.net/browse/KAD-3599)
...

### Checklist

### DIFF
--- a/src/packages/components/src/range/radio-range-control/index.js
+++ b/src/packages/components/src/range/radio-range-control/index.js
@@ -123,6 +123,7 @@ export default function RadioRangeControl( {
 								step={ step }
 								help={ help }
 								allowReset={ true }
+								initialPosition={ sizeValue === 0 ? 0 : "" }
 							/>
 						</div>
 						{ ( onUnit || showUnit ) && (


gilbert-hernandez:
[Video of the affects of the change](https://share.zight.com/2NuQj4yq)

The fix doesn't address the Jira issue exactly. However, Karla saw another issue in her video (in her comment in the JIRA ticket). This fix is for the radio range controls. 

When the component refreshes, if the value is 0, the input field is empty. You can set a default value of 0, but this is a problem in the responsive radio range control when the values are inheriting. This fix ensures the initialPosition of the radio range control is 0 when the value is 0 and is blank otherwise. This way, when the value is 0 and the component refreshes, the _input placeholder_ is 0 instead of empty. However, if the value is inheriting any other size, the initialPosition is empty.

gilbert-hernandez:
after some testing, this may not work as expected. There is a bug that is different from the JIRA ticket. The problem is that the customGutter attribute is shared by desktop, tablet, mobile, which can reset the gap values.

gilbert-hernandez:
@oakesjosh 

This fix is ready for review. It solves two problems. First, the 0 value was disappearing in the input fields when the component was reloaded. Now, the 0 should continue to show in the input field when you set it to 0 and reload the component. 

Second, a 0 value would save as an empty string when you changed the value for the gaps. For example, set desktop column gap to 0, then change tablet/mobile column gap to 100. When you returned to the desktop column, the 0 wasn’t showing in the input field and it was inheriting the “MD” setting instead of a 0 gap. The fix uses the ?? to treat 0 as truthy.

PR #674, opened 2025-02-27
KAD-4108 Fix setting numeric padding/margin on submenu links

oakesjosh:
[KAD-4108](https://stellarwp.atlassian.net/browse/KAD-4108)

This issue is due to CPT needing meta to be stored as a string, we convert the numeric padding values to a string. The padding component sees a string and thinks it's a variable value (SM, MD, LG). I created the `arrayStringToInt` helper a while back to support converting strings of number back into integers so the component handles them correctly.  We use this function in this file and the advanced header.

[KAD-4108]: https://stellarwp.atlassian.net/browse/KAD-4108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -174,6 +174,10 @@ Please report security bugs found in the Kadence Blocks plugin's source code thr
 
 == Changelog ==
 
+= 3.4.12 =
+Release Date: TBD March 2025
+* Fix: Setting numeric padding & margin on Adv Navigation sub menu links.
+
 = 3.4.11 =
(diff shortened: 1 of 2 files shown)


gilbert-hernandez (approved):
I tested this and the fix works for me. The custom link padding and margin now lets me update each side without disappearing.

PR #675, opened 2025-02-28
KAD-4209 Init progress bars on query loop filter

oakesjosh:
When filtering in QL, the progress bars within the query loop aren't initialized.

On the `kb-query-loaded` event we only initialize progress bars inside the query loop card class, otherwise progress bars elsewhere on the page will re-init and animate again.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -177,6 +177,7 @@ Please report security bugs found in the Kadence Blocks plugin's source code thr
 = 3.4.12 =
 Release Date: TBD March 2025
 * Fix: Progress bars not initializing insdie query loops.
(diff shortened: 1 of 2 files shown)


mark-c-woodard (commented):
This is working, but the `onlyInQueryLoop` approach isn't very extendable.
In the future, we may also be asked to make the progress bar initialize in the quickview modal or other areas that get loaded async.

What I've done in other situations like this is adding a marker class to the block when an `init` functions runs over it. Then make the `init` functions check for that class before they run. So you can always loop over all progress bars on the page.
Then you can just add another event listener when it needs to re-init on something new.

gilbert-hernandez (approved):
This fix works for me. The progress bars load again after the QL is filtered.

PR #680, opened 2025-03-10
bugfix/KAD-4148 Text Shadow opacity set to 100% after any change

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4148](https://stellarwp.atlassian.net/browse/KAD-4148)

The text shadow opacity wasn't saving correctly. I found some issues that I had solved in the box shadow for the Info Box. The opacity is now saved and the color is handled in Hex. The array for the color picker is also handled now so that it updates correctly.


gilbert-hernandez:
@oakesjosh 
I couldn't quite reproduce the warning you received, but I believe this fix should address it. The way the color was generated previously depended on an rgb or rgba color without an opacity value. Now there is an opacity value so that the color and color picker are more stable. However, I've added checks to determine if rgba is used. If they are used, then the opacity value will be ignored. This should provide for backwards compatibility. I address the backend and frontend. Also, in the frontend, I cleaned up the frontend values by using a get_cascading_value() function for responsive inheritance.

mark-c-woodard:
suggested using a more re-usabel approach. Bringing the logic for outputting responsive shadow styles out into the class-kadence-blocks-css class.

mark-c-woodard:
@oakesjosh @gilbert-hernandez What's the deal with this one? did we say this was going to go out with the shadow presets?

gilbert-hernandez:
@mark-c-woodard I made some changes to make a render_text_shadow function on the CSS class. Josh reviewed it and I've made changes as requested.

PR #681, opened 2025-03-10
bugfix/KAD-4101 Header Block Row Padding and Margin always set to px

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4101](https://stellarwp.atlassian.net/browse/KAD-4101)

The header rows were using marginUnit and paddingUnit as attributes when the kadence-blocks-css class function render_measure_output() requires the variables to use "Type" instead of "Unit". This fix updates the component variable names to marginType and paddingType. The old variable names are kept in the block.json to update to the new names.


gilbert-hernandez:
@oakesjosh 
updated with a much simpler fix!

PR #684, opened 2025-03-12
Vector Block

oakesjosh:
Create vector block

### DIFF
--- a/src/blocks/vector/edit.js
+++ b/src/blocks/vector/edit.js
@@ -166,25 +166,27 @@ export function Edit(props) {
 			path: '/kb-vector/v1/vectors',
 			data: { vectorSVG: tmpSvgContent, title },
 			method: 'POST',
-		}).then((response) => {
-			setIsSaving(false);
-			if (has(response, 'value') && has(response, 'label')) {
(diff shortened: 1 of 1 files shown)


gilbert-hernandez:
I tested this and I notice a couple of things. First, the icon for the vector graphic block is black, not the Kadence blue. Second, after I add a vector and save it, the code from the previous vector is still in the "Code" box when I try to add another vector. I would expect the code box to be empty after it is saved. Also, while the custom classes added in the Advanced section appear on the frontend, an HTML anchor does not. The other settings seem to work and are responsive.

gilbert-hernandez:
@oakesjosh 
I noticed another problem. When you press the "Save" button, there is a moment of loading before the UI goes away. The "Save" button doesn't become busy and I can press save a bunch of times. Someone is impatient may do the same. It then creates the custom post several times and increases the load time.
[https://share.zight.com/v1uGZ7ze](https://share.zight.com/v1uGZ7ze)

gilbert-hernandez:
Looks good to me.

mark-c-woodard (approved):

PR #687, opened 2025-03-20
quickwin/KAD-4245_custom_css_indicator

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4245](https://stellarwp.atlassian.net/browse/KAD-4245)

Adds a CSS indicator similar to the Conditional Display indicator. This indicator can be enabled/disabled in the Kadence Blocks Settings (page refresh required).

[https://share.zight.com/v1uz5Ep1](https://share.zight.com/v1uz5Ep1)


mark-c-woodard (changes requested):
Like the data store work, but lets rename everything in the data store from customCssIndicator to something like blockRenderKey. Something more generic.
That way if we want to expand this out or something else wants to re-render our blocks later it will be more clear what the purpose of this is.

mark-c-woodard (approved):

PR #689, opened 2025-03-24
bugfix/KAD-4223-3

mark-c-woodard:
https://stellarwp.atlassian.net/browse/KAD-4223

Should do this with a deprecation, but could not get one to work.


gilbert-hernandez (approved):

mark-c-woodard:
> Should this apply to select and textarea elements too?

it should be, added

PR #692, opened 2025-03-26
accessibility/KAD-4242 Accessibility labels for carousels (Gallery Adv and Post Grid/Carousel blocks)

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4242](https://stellarwp.atlassian.net/browse/KAD-4242)
added an aria-label to the gallery(adv) carousel in the PHP. Adding it in the JS would require altering a vendor file (splide.min).


mark-c-woodard (commented):
The ticket mentions aria-labels for the lightbox captions as well, didn't see any changes there.

Was that intentional because those changes are the vendor file?

gilbert-hernandez:
@mark-c-woodard @oakesjosh 
I didn't include the aria-label for the lightbox because 

> The ticket mentions aria-labels for the lightbox captions as well, didn't see any changes there.
> 
> Was that intentional because those changes are the vendor file?

Actually, the aria-label for the lightbox was the label that read "Display this image in a lightbox". They thought it might be too generic. However, after discussing with Josh, it seems appropriate so no changes were made to it. 

[https://share.zight.com/xQuoADbb](https://share.zight.com/xQuoADbb)

PR #693, opened 2025-03-27
KAD-4276 Strip style tags and their content from FAQ schema

oakesjosh:
We strip tags from the block content, but strip_tags retains the inner content. This prevents the contents of our inline styles from appearing as content in the FAQ schema.


gilbert-hernandez (commented):
I tested this and the styles were removed. However, the classes were still include. For example, below is what I saw in the schema. Should the classes be removed too?

<p class="kt-adv-heading2621_4ba159-c0 wp-block-kadence-advancedheading" data-kb-block="kb-adv-heading2621_4ba159-c0">This is the second answer</p>

PR #694, opened 2025-03-28
stripping links from the main gallery captions when the link is set u…

mark-c-woodard:
…p to trigger the lightbox gallery.

https://stellarwp.atlassian.net/browse/KAD-4231


gilbert-hernandez (approved):
Looks good and works for me. I don't see repeating images in the lightbox when the caption has a link.

PR #695, opened 2025-03-31
enhancement/KAD-4232

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4232](https://stellarwp.atlassian.net/browse/KAD-4232)

adds the custom category colors to the free Posts block.


gilbert-hernandez:
Found an issue where this doesn't work when there is a custom slug for the category. Need to rework the solution.

mark-c-woodard (approved):

PR #698, opened 2025-04-03
enhancement/KAD-4155_text_block_copy_paste

gilbert-hernandez:
This change lets you copy and paste into the Text(Adv) more like the standard Gutenberg Paragraph block.
Here is a video of how copy and paste works in the text block with the update: [https://share.zight.com/4guvbPNz](https://share.zight.com/4guvbPNz)


gilbert-hernandez:
I tested this and it still doesn't work with blocks like the gallery and query blocks. Will reopen when that is addressed.

PR #703, opened 2025-04-11
bugfix/KAD-4117_fix_collapse_sub_menus

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4117](https://stellarwp.atlassian.net/browse/KAD-4117)

In the editor, when Orientation value is set to "inherit" for tablet or mobile, they will see the "Vertical" options as expected if inheriting "Vertical". 

Additionally, tablet and mobile vertical navigation will respect the 'Collapse Vertical Sub Menus' option assigned to it.


mark-c-woodard (changes requested):
Looks good, just one change. Seems like that lined out the inheritance to me.

Please also run the same kind of testing on the other responsive toggle options in that "Other Options" section. They may need this same change.

mark-c-woodard (approved):

PR #710, opened 2025-04-16
quickwin/KAD-4316

mark-c-woodard:
https://stellarwp.atlassian.net/browse/KAD-4316


mark-c-woodard:
> Working well for me. Should we show apply the video background in the editor if one exists?

We could/should, but in the name of making this really quick I did not.

PR #723, opened 2025-05-05
quickwin/KAD-4258_gallery_overflow

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4258](https://stellarwp.atlassian.net/browse/KAD-4258rl)

Adds overflow gallery option to carousel. There is also a selector change for the padding/margin on the frontend to help carousel look like the editor. This may need to be reviewed to make sure it doesn't affect other gallery types.


mark-c-woodard (changes requested):
Looks good. 
The overflow did not work when the slider was within a section in a row

gilbert-hernandez:
> Looks good. The overflow did not work when the slider was within a section in a row

This is expected. The gallery needs left/right padding/margin for the overflow to be seen in most cases. If the gallery is by itself on a page layout that has left/right padding, you would see the overflow without adding additional spacing to the gallery. But, inside of a row or section, the gallery would need left/right spacing for the overflow to be seen.

mark-c-woodard (approved):

PR #731, opened 2025-05-15
KAD-4442 Fix pasting to default adv text overwriting content

oakesjosh:
🎫  [KAD-4442
](https://stellarwp.atlassian.net/browse/KAD-4442)

When using the adv text as the default block, and text is pasted, we call `replaceBlocks` with the new content. This removed any existing content in the block.

I've updated this to append the text to the existing string. I've the user has pasted multiple blocks/paragraphs of text this will append the text from the first paragraph and insert new blocks below the existing blocks. If the user is pasting block HTML this will just insert below the existing content.


gilbert-hernandez (approved):
This works for me. I can now text additional text into the text blocks instead of completely replacing the text block.

mark-c-woodard (approved):

PR #735, opened 2025-05-20
bugfix/KAD-4448_fluid_carousel_gutter

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4448](https://stellarwp.atlassian.net/browse/KAD-4448)

 If the gallery has a gutter, the overflow and the gutter can cause a strange layout. Additionally, when you set the gallery to a regular ‘Carousel’ and enable the carousel overflow, the editor shows a scroll bar for x-axis. This fix address both issues.


mark-c-woodard (approved):
works for me

PR #740, opened 2025-05-27
KAD-4405 suppress filters added for tribe_events when order_by is pos…

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4405](https://stellarwp.atlassian.net/browse/KAD-4405)

This is a pro feature, but the fix is in free. This uses ssuppress_filters in the pre_get_posts hook when the query displays tribe_events and the user selects individual posts.

### DIFF
--- a/includes/class-kadence-blocks-posts-rest-api.php
+++ b/includes/class-kadence-blocks-posts-rest-api.php
@@ -243,6 +243,13 @@ class Kadence_Blocks_Post_Rest_Controller extends WP_REST_Controller {
 				}
 			}
 		}
+		// Add filter to suppress query modifications for tribe_events when using post__in ordering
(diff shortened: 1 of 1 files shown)


mark-c-woodard (commented):
get_query_items is only used for getting the posts for the editor right?
I don't think this will address the frontend issue shown in the ticket. 

Also, did you test if using suppress filters affect loops that are using the inherit option? Like if there using the query loop to replace the events archive page?

gilbert-hernandez:
> get_query_items is only used for getting the posts for the editor right? I don't think this will address the frontend issue shown in the ticket.
> 
> Also, did you test if using suppress filters affect loops that are using the inherit option? Like if there using the query loop to replace the events archive page?

The fix for the frontend is actually in pro after all. [https://github.com/stellarwp/kadence-blocks-pro/pull/172](https://github.com/stellarwp/kadence-blocks-pro/pull/172) 

As for the issue with inheriting queries, I've added a check to address a fatal error that was occurring when using the inherit query setting. I also added it to the pro PR.

PR #743, opened 2025-05-28
KAD-4492 disabled links now use span

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4492](https://stellarwp.atlassian.net/browse/KAD-4492)

Disabled links in the navigation block now use <span> instead of <a>. Previously, the disabled links used an <a> without an href which caused SEO audits to raise an error about links not crawlable.


mark-c-woodard (changes requested):
This has gotten brought up on a few occasions for links in the theme nav and adv navigation.
https://stellarwp.atlassian.net/browse/KAD-839

The problem for us styling. Changing the markup of a really visible element like this is bound to create styling issues for us and users' custom styles.

There are some alternatives for empty hrefs mentioned here that I think we should try before changing the markup to span. 
https://stackoverflow.com/questions/5637969/is-an-empty-href-valid
Let's see if we can get lighthouse to play along without changing the markup.

mark-c-woodard (approved):

skyshab:
Hi folks! I am a member of the Stellar Web team, and was investigating solutions for making the empty links compliant with our accessibility recommendations. 

I had come up with a fairly simple solution before seeing this PR. 

Not sure if this is useful, but you can see how I tried to handle it here: https://github.com/stellarwp/kadence-blocks/pull/749

PR #751, opened 2025-06-09
KAD-4506 Add color styling for select options in advanced form

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4506](https://stellarwp.atlassian.net/browse/KAD-4506)

Adds color style for form select field options. The style only is only added when the form is dark.


src/blocks/advanced-form/style.scss
```diff
@@ -499,3 +499,7 @@
 	// content: "a";
 	// content: attr(placeholder);
 }
+
+.kb-form-is-dark .kb-adv-form-field.kb-adv-form-infield-type-input.wp-block-kadence-advanced-form-select select option {
```
mark-c-woodard, line 503: For organization, I would move this under the other .kb-form-is-dark styles. 
Also this seems like it has a really high specificity. Is that necessary?

mark-c-woodard (changes requested):

src/blocks/advanced-form/style.scss
gilbert-hernandez: @mark-c-woodard 
I moved the selector and made it less specific. It didn't need to be so specific.

PR #753, opened 2025-06-10
KAD-4284 Add screen reader text format to advanced heading block

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4284](https://stellarwp.atlassian.net/browse/KAD-4284)

Here is a video of how this currently works: [https://share.zight.com/8LuEEdyz](https://share.zight.com/8LuEEdyz).

The one problem I had with this is the frontend style. For some reason, the \kadence-blocks\src\blocks\advancedheading\style.scss doesn't load for the advanced heading block, so I added the CSS style to the PHP file instead.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -180,6 +180,7 @@ Release Date: 19th June 2025
 * Fix: Form select option color for Windows.
 * Fix: Section responsive vertical alignment.
 * Fix: Responsive controls not working in widget editor.
+* Add: Screan reader text styles for Advanced Text block.
 
 = 3.5.10 =
 Release Date: 5th June 2025


mark-c-woodard:
Found an issue when starting with screen reader text enabled and beginning to type
https://www.loom.com/share/2aadfe6311074769975f33345ed395c9?sid=611eec90-7fab-4fe2-80d8-91c1aabab6ff

includes/blocks/class-kadence-blocks-advanced-heading-block.php
```diff
@@ -422,6 +422,15 @@ public function build_css( $attributes, $css, $unique_id, $unique_style_id ) {
...
 
+		$css->set_selector( '.kb-screen-reader-text' );
+		$css->add_property( 'position', 'absolute' );
+		$css->add_property( 'width', '1px' );
+		$css->add_property( 'height', '1px' );
+		$css->add_property( 'padding', '0' );
+		$css->add_property( 'margin', '-1px' );
+		$css->add_property( 'overflow', 'hidden' );
+		$css->add_property( 'clip', 'rect(0, 0, 0, 0)' );
+
```
mark-c-woodard, line 433: I'm not sure why the stylsheet for this block isn't loading, but looks like it's been a thing for a while.
If you look further down there's a spot where we do a wp_add_inline_style to attach some styles (which appear to be the same ones in the sheet) to the heading stylesheet.

This whole thing seems really wrong and I feel like we should just be loading the stylesheet. But there's probably a reason, so you should do some digging.
Find out why we're not loading the styles from the sheet, and see if we can start loading the sheet.

If we can't, then add these styles along with the other ones to the wp_add_inline_style call.

mark-c-woodard (changes requested):

gilbert-hernandez:
@mark-c-woodard 

I addressed several issues you pointed out. First, the style.scss now properly loads in the frontend. Second, I solved the issue in the editor when you don't have text selected and you enable the screen reader text. The Advanced Highlight feature does something similar, but it isn't quite noticeable like the screen reader text. 

The problem was that the span was added to every character instead of wrapping a single span around the entire string. I found that the core bold and italic settings work correctly and I found that they used a data attribute to solve the problem. Now, there is a single span that surrounds the screen reader text while the setting is enabled. Additionally, the text block now gains focus when the screen reader text is enabled, which is the same behavior as bold/italic. 

Lastly, I updated the editor style to prevent the "Screen Reader Text" content to wrap to new lines, which was causing it to hide the user input until it was long enough for the "Screen Reader Text" to display on a single line.

gilbert-hernandez:
I also added a fix to the advanced heading feature. When the advanced heading feature is enabled, the text block now gains focus and the user can input characters until they disable the advanced heading feature, and there will only be a single mark element around the entire string instead of around each individual characters.

gilbert-hernandez:
Added frontend styles for screen reader text to the php file and stopped loading the style.scss file.

mark-c-woodard (approved):

PR #755, opened 2025-06-18
KAD-3459 Prevent design library re-render with lifter LMS

oakesjosh:
When lifterLMS is active the design library button is re-rednering and flashes while editing a course.


gilbert-hernandez:
@oakesjosh 
When I edit a course page, the design library button is visible, but then flickers and disappears as the "Launch Course Builder" button becomes visible. Then, I only see the Design Library button once I click in the editor to edit a block. However, at that point it does stop flickering on and off. Is that to be expected with this fix or was the goal to prevent the initial flickering as well?

[https://share.zight.com/wbudEOO4](https://share.zight.com/wbudEOO4)

PR #771, opened 2025-07-07
Video posters for the video popup block

mark-c-woodard:
https://stellarwp.atlassian.net/browse/KAD-4017


gilbert-hernandez:
@mark-c-woodard 

When the Pro plugin is not activated ,you can't tell there is a premium feature for the video poster. The dropdown is just missing altogether. In previous features where I had this issue, I showed the dropdown but deactivated the pro selection and added (Pro) in the text. This way, free users see the dropdown with the option that is only available for pro.

mark-c-woodard:
good point! I'll update that.

PR #812, opened 2025-08-15
bugfix/KAD-4781_tooltip_zindez

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4781](https://stellarwp.atlassian.net/browse/KAD-4781)
Appends the tooltip to the body instead of to the parent. Per [Tippy documentation](https://atomiks.github.io/tippyjs/v6/all-props/#appendto): "Sometimes the tippy needs to be appended to a different DOM context due to accessibility, clipping, or z-index issues."


gilbert-hernandez:
@mark-c-woodard 
Yeah, I'm not finding a way to access the tooltip content with a screen reader when I append the content to the document body instead of the parent. I could potentially only append to the document body in some cases, but those cases would still be inaccessible, which doesn't seem reasonable.

Should I decline this ticket again as in times past? Generally, I've recommended using custom CSS to solve the problem when it arises.

gilbert-hernandez:
Declining as this fix would cause the tooltips to be inaccessible.

PR #828, opened 2025-09-18
KAD-4953 Remove TOC links that are not conditionally displayed.

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4953](https://stellarwp.atlassian.net/browse/KAD-4953)

When adding the anchors to the TOC, this fix removes links with missing headings (hidden with conditional display) and cleans up missing links. 

This is a JS fix as it seemed simpler than trying to come up with a PHP fix. That might require making updates in free and pro blocks, but I didn't explore a PHP solution.


gilbert-hernandez:
Closing for a PHP solution.

PR #830, opened 2025-09-24
KAD-4873 Add CSS to php file.

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4873](https://stellarwp.atlassian.net/browse/KAD-4873)

The style.scss file isn't loading for some reason. Added a grid style to the PHP file instead.


gilbert-hernandez:
> @gilbert-hernandez in Jira you mentioned that style file wasn't loading, but now it is. Also there aren't any changes here relating to file loading. So was that a fluke or what was the deal with that?

Yeah, it was a face palm moment. Not sure what I was seeing the first time around, but the css was loading just fine the second look.

PR #836, opened 2025-09-29
Adding some admin bar nodes for kadence headers.

mark-c-woodard:
https://stellarwp.atlassian.net/browse/KAD-5005


mark-c-woodard:
ya, I could've made a tracking thing like Elements has to show which headers are rendered on this page. 
But didn't feel worth it considering most of the time you'll probably only have one header, and it'll probably be the one you've chosen for your theme header.

PR #841, opened 2025-09-30
Moving attaching hash and query loading event listeners for accordion…

mark-c-woodard:
…s outside of init.

Turns out if you add an event listener that triggers init inside of init you get hyperbolic stacking of calls on recurring events. https://stellarwp.atlassian.net/browse/KAD-4950


gilbert-hernandez (approved):
Good find. I was able to recreate the problem and this branch fixes the issue.

PR #842, opened 2025-09-30
Pull icons from REST endpoint instead of passing in editor assets

oakesjosh:
This loads icons via a REST request after page load, and into the data store. The components in pro will need to be updated too, but this is backwards compatible due to `should_return_icons()`

While icons are loading, a loading spinner is show in its place. Once loaded, the icons are stored in the browsers [CacheStorage](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage).  On subsequent requests, the data store will load icons from the browser cache storage, while also sending off a request to the rest endpoint. The rest endpoint returns a hash of the icons which is compared against the CacheStorage and will update if needed.

CacheStorage requires https so if it isn't working locally, verify that the local site is using https

### DIFF
--- a/includes/class-kadence-blocks-icons-rest.php
+++ b/includes/class-kadence-blocks-icons-rest.php
@@ -127,7 +127,7 @@ class Kadence_Blocks_Icons_REST_Controller extends WP_REST_Controller {
         // Only return user custom icons (Pro). Bundled social icons are included in solid/line files.
         $custom_icons = [];
 
-        if ( ! $this->has_pro_support() ) {
+        if ( ! $this->has_pro_support() || ! current_user_can( 'edit_posts' ) ) {
             return $custom_icons;
         }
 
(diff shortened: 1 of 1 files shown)


includes/class-kadence-blocks-icons-rest.php
```diff
@@ -0,0 +1,189 @@
...
+	 */
+	public function register_routes() {
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base,
+			array(
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_icons' ),
+					'permission_callback' => '__return_true',
```
gilbert-hernandez, line 38: Should there be any permissions? Should we check if the user can edit posts?

mark-c-woodard (approved):

gilbert-hernandez (approved):

PR #847, opened 2025-10-08
Tabs accessibility fixes

sitdev:
...

### Checklist


cyruscollier:
Hi @mark-c-woodard, thanks for the added context. It's not my place to question an accessibility audit, but in my opinion this block's implementation of the ARIA tabs pattern is highly unconventional and likely would not survive another audit or direct screen reader user testing. I based these recommended changes directly from the experience of two audits by two separate firms of our clients' sites. 

Since the site we are currently building using Kadence Blocks will be going through its own audit in the coming months, if that audit finds serious issues with this block, would you re-consider these changes?

PR #852, opened 2025-10-18
Bugfix: Rowlayout margin issues

cyruscollier:
There are 2 margin-related issues with the Row Layout block that this PR addresses:

1. Rowlayout uses its own version of `getSpacingOptionOutput()` that does not correctly return an empty string for all false-y values, including `''` and `null`. This results in the margin reset state to not work. Now the block uses the shared version of this function from `@kadence/helpers` that correctly handles these false-y values.

2. There are editor-only styles that zero out the top and bottom margins of the Rowlayout block. This can conflict with theme styles, especially due to their high specificity, resulting in inconsistent visual spacing between the editor preview and the front end display. Since there isn't an apparent need for these styles, they have been removed so that default theme margins can render correctly in both the preview and front end.

### Checklist
- [X] I have performed a self-review.
- [X] No unrelated files are modified.
- [X] No debugging statements exist (Ex: console.log, error_log).
- [X] There are no warnings or notices in the wordpress error log.
- [X] Passes all tests (linting, acceptance, & unit)

- [X] Tested with an existing instance of this block .
- [X] Tested creating a new instance of this block.
- [X] Tested with Dynamic content & Elements.

### DIFF
--- a/src/blocks/rowlayout/editor.scss
+++ b/src/blocks/rowlayout/editor.scss
@@ -737,14 +737,6 @@ div.kb-blocks-bg-video {
 	height: auto;
 	width: 100%;
 }
-.block-editor-block-list__block[data-type="kadence/rowlayout"] {
-	margin-top: 0;
-	margin-bottom: 0;
-}
-.block-editor-block-list__block[data-type="kadence/rowlayout"] > .wp-block {
-	margin-top: 0;
-	margin-bottom: 0;
-}
 .wp-block-kadence-rowlayout.has-border-radius {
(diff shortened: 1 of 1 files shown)


mark-c-woodard:
@cyruscollier 
Point one looks good, thanks for the find!

Point two, we reset the margins on most of our blocks because gutenberg adds margins on blocks, so this is just to ensure that doesn't mess with us.
Was there a specific inconsistency your were dealing with? I think I'd prefer to handle those indivdually.

PR #853, opened 2025-10-20
KAD-4956 Updated nth-of-type to nth-child for collapse order and brea…

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-4956](https://stellarwp.atlassian.net/browse/KAD-4956)

Changed nth-of-type to nth-child in the selector for collapse order and break left/right settings

### DIFF
--- a/includes/class-kadence-blocks-prebuilt-library-rest-api.php
+++ b/includes/class-kadence-blocks-prebuilt-library-rest-api.php
@@ -1342,7 +1342,7 @@ class Kadence_Blocks_Prebuilt_Library_REST_Controller extends WP_REST_Controller
 			if ( ! empty( $pattern_style ) ) {
 				$args['style'] = $pattern_style;
 			}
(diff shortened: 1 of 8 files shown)


oakesjosh:
It looks like the last change to these lines was changing it from `:nth-child` to `:nth-of-type` to fix KAD-3421

We may need to find a different approach so we don't have a regression on that

gilbert-hernandez:
@oakesjosh 

This is strange. The problem with KAD-3421 doesn't seem to exist anymore. When I recreate 3421 with this fix, everything works as expected. The problem was that there was a <style> tag inserted before each column. However, I can't seem to get those <style> tags to appear anymore, so nth-child() works just fine. 

I'm not sure why those <style> tags appeared back then, but I can't trigger them now.

mark-c-woodard:
what a selector :o

PR #856, opened 2025-10-23
KAD-4847 Change testimonial carousel UL back to a DIV as is not the s…

gilbert-hernandez:
…plide__list.

[https://stellarwp.atlassian.net/browse/KAD-4847](https://stellarwp.atlassian.net/browse/KAD-4847)

Reverted the carousel testimonial UL back to a DIV as it is not the splide__list and is not the parent to the list items in the carousel.


gilbert-hernandez:
> Does the proposed solution of removing the left padding work or is there something else going on too?

@oakesjosh 
Yeah, I think this is related to other issues with the UL and LI changes. Let me look into #860 more to see if that helps with this issue.

PR #864, opened 2025-11-10
Fix mask based progress bars not showing on the frontend in Safari

oakesjosh:
There's still an issue with the display in the editor. This fixes the bigger issue, which is frontend display.


oakesjosh:
@gilbert-hernandez Sorry, forgot to tag it with the ticket. https://stellarwp.atlassian.net/browse/KAD-5239

Steps to reproduce the behavior:

Add a Progress Bar block inside of a Section block

Set the flex direction of Section block to horizontal

Select Star setting for the Progress Bar display

See that it does not render in editor or front-end when using Safari

gilbert-hernandez (approved):
@oakesjosh 
Thanks! I was able to reproduce the problem and this fix works for me.

PR #876, opened 2025-12-03
Linter fixes

oakesjosh:


oakesjosh:
Merged different branch with these fixes into master. This branch had a reversion of some commits that already existed on master, so merging would revert those changes.

PR #882, opened 2025-12-12
bugfix/KAD-5340_carousel_versioning

gilbert-hernandez:
[https://stellarwp.atlassian.net/browse/KAD-5340](https://stellarwp.atlassian.net/browse/KAD-5340)

This fixes two issues. First, it removes some margin and padding from advanced galleries related to column gutter. Additionally, it fixes the gallery HTML when versioning is under 2 or missing.


gilbert-hernandez:
@mark-c-woodard 
I have made changes and I believe I have addressed the issue you mentioned. Thanks for pointing out that line 375 in the edit file and I just sort of converted that to PHP next to the column fix that was there.

PR #968, opened 2026-04-03
Fix: Escape key dismisses hover-opened submenus (WCAG 2.1 SC 1.4.13)

Contributolo:
## Problem

The Navigation (Adv) Block already handles `Escape` for submenus opened via toggle button click (`keyCode === 27` in `kb-navigation-block.js`). However, submenus opened via CSS `:hover` cannot be dismissed with `Escape`.

The CSS rule responsible:

```css
.wp-block-kadence-navigation.navigation-desktop-orientation-horizontal
    .menu-item:not(.kb-nav-link-sub-click):hover > .sub-menu {
    opacity: 1;
    visibility: visible;
}
```

Since `:hover` is a CSS pseudo-class that cannot be removed via JavaScript, pressing `Escape` while hovering a submenu has no effect.

This violates:
- **WCAG 2.1 SC 1.4.13 (Content on Hover or Focus):** _"A mechanism is available to dismiss the additional content without moving pointer hover or keyboard focus."_
- **WCAG 2.1 SC 2.1.1 (Keyboard):** All functionality must be operable through a keyboard interface.

Since June 2025, the **European Accessibility Act (EAA)** makes WCAG 2.1 AA compliance a legal requirement for commercial websites in the EU.

## Solution

- Track hovered `.menu-item-has-children` elements via `mouseenter`/`mouseleave`.
- Add a document-level `keydown` listener for `Escape` that adds a `kb-nav-esc-close` class to the deepest hovered menu item.
- A CSS rule `.menu-item.kb-nav-esc-close > ul.sub-menu { display: none !important }` overrides the `:hover` visibility. The `!important` is necessary because CSS `:hover` cannot be programmatically disabled.
- The class is automatically removed on `mouseenter` (re-hover) or `mouseleave` (pointer exits), restoring normal behavior.
- The existing `Escape` handler for toggle-opened submenus (keyboard focus inside submenu) is unchanged; the new global handler skips if focus is already inside a `sub-menu`.

## Changes

- **`src/assets/js/kb-navigation-block.js`**: Added `hoveredItems` tracking, `initHoverEscDismiss()`, and `handleHoverEscDismiss()`.
- **`src/blocks/navigation/style.scss`**: Added `.kb-nav-esc-close` override rule.

## Testing

1. Open a page with Navigation (Adv) Block containing submenus.
2. **Hover test:** Hover a menu item with children -- submenu appears. Press `Escape` -- submenu closes. Move pointer away and back -- submenu reopens normally.
3. **Keyboard test:** Tab into a submenu, press `Escape` -- submenu closes, focus returns to toggle button (existing behavior, unchanged).
4. **Nested hover:** Hover a top-level item, then hover a nested child with its own submenu. Press `Escape` -- only the deepest submenu closes.
5. **Vertical navigation:** Verify vertical orientation is unaffected (hover submenus are not used in vertical mode).


src/assets/js/kb-navigation-block.js
```diff
@@ -6,6 +6,7 @@
 (function () {
 	const focusableElementsString =
 		'a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), iframe, object, embed, [tabindex="0"], [contenteditable]';
+		const hoveredItems = new Set();
```
gilbert-hernandez: Could you please fix the indentation. there is an extra tab.

src/blocks/navigation/style.scss
```diff
@@ -1621,3 +1621,8 @@ $link-color-property: var(
 		color: $link-color-property;
 	}
 }
+
+// Escape key dismissal for hover-opened submenus (WCAG 2.1 SC 1.4.13).
+.menu-item.kb-nav-esc-close > ul.sub-menu {
+	display: none !important;
+}
```
gilbert-hernandez: The new CSS is global but should be limited to the horizontal orientation as hover based submenus only exist on horizontal navigations.  It should look more like:

```
.navigation-desktop-orientation-horizontal .menu-item.kb-nav-esc-close > ul.sub-menu {
      display: none !important;                                                         
 }
```

src/assets/js/kb-navigation-block.js
```diff
@@ -402,6 +403,72 @@
 		});
 		runSubMenuContentSize();
 	};
+	/**
+	 * Allow Escape key to dismiss hover-opened submenus (WCAG 2.1 SC 1.4.13).
+	 *
+	 * CSS :hover cannot be removed programmatically, so a class-based override
+	 * is applied and cleaned up when the user re-engages with the menu item.
+	 */
+	const initHoverEscDismiss = function (nav) {
```
gilbert-hernandez, line 412: Could you also add a guard here to make sure you don't add the listeners to the vertical orientation.

```
// Hover submenus only exist on horizontal orientation.                           
if (nav.classList.contains('is-vertical')) {
      return;                                                                       
} 
```

gilbert-hernandez (changes requested):

src/assets/js/kb-navigation-block.js
Contributolo: Thanks for the review! All three points addressed:

1. Fixed the extra tab on `const hoveredItems`.
2. Added the vertical orientation guard at the top of `initHoverEscDismiss`.
3. Scoped the `kb-nav-esc-close` rule to `.navigation-desktop-orientation-horizontal`, matching the existing hover rule scope.

src/blocks/navigation/style.scss
Contributolo: Thanks for the review! All three points addressed:

1. Fixed the extra tab on `const hoveredItems`.
2. Added the vertical orientation guard at the top of `initHoverEscDismiss`.
3. Scoped the `kb-nav-esc-close` rule to `.navigation-desktop-orientation-horizontal`, matching the existing hover rule scope.

src/assets/js/kb-navigation-block.js
Contributolo: Thanks for the review! All three points addressed:

1. Fixed the extra tab on `const hoveredItems`.
2. Added the vertical orientation guard at the top of `initHoverEscDismiss`.
3. Scoped the `kb-nav-esc-close` rule to `.navigation-desktop-orientation-horizontal`, matching the existing hover rule scope.

Contributolo:
@gilbert-hernandez ?

Contributolo:
All three points from the review on 28 April are addressed in the current branch:

1. Removed the extra tab on `const hoveredItems`.
2. Added the vertical orientation guard at the top of `initHoverEscDismiss`.
3. Scoped the `kb-nav-esc-close` rule to `.navigation-desktop-orientation-horizontal`, matching the scope of the existing hover rule.

The branch still merges cleanly against master, and I re-checked that the anchor points the patch relies on are unchanged. Glad to rebase onto current master if that makes review easier.

Re-requesting review so this shows up in the queue again — it has been sitting in "changes requested" since the points were addressed.

src/blocks/navigation/style.scss
```diff
@@ -1621,3 +1621,8 @@ $link-color-property: var(
 		color: $link-color-property;
 	}
 }
+
+// Escape key dismissal for hover-opened submenus (WCAG 2.1 SC 1.4.13).
+.navigation-desktop-orientation-horizontal .menu-item.kb-nav-esc-close > ul.sub-menu {
+	display: none !important;
+}
```
rahulacaleffi:
```suggestion
```
Thanks @Contributolo for the contribution! The feature works. One small thing to fix before:
`display: none !important` also outranks `.menu-item.menu-item--toggled-on > .sub-menu (style.scss:227)`, which is how submenus open via the ▾ toggle button:

1. Hover a parent item so the submenu opens.
2. Press `Escape` - hides correctly.
3. Without moving the pointer, activate that item's ▾ toggle with the keyboard - `Enter`.
It works, but the submenu stays with `display:none`, so a screen reader is told it's open while nothing is visible.

Suggested fix: Aside dropping the rule here `style.scss:1625-1628`, narrow the existing one at `style.scss:226` instead, adding the `:not(.kb-nav-esc-close)` attribute:
```
.menu-item:not(.kb-nav-link-sub-click):not(.kb-nav-esc-close):hover > .sub-menu,
.menu-item.menu-item--toggled-on > .sub-menu {
	@include show-sub-menu();
}
```

No !important needed, and only the hover path is affected. It also resolves the earlier scoping request properly: `navigation-desktop-orientation-horizontal` is on the wrapper at every viewport width, so the current rule doesn't actually scope to horizontal. `horizontal-navigation-base-styles` is already wrapped in the correct media queries.

Otherwise, it looks good to me. 

cc @gilbert-hernandez

rahulacaleffi (changes requested):

Contributolo:
Thanks for the thorough review, and for catching the toggle conflict — that scenario is a real problem, an element reported as expanded while nothing is visible is worse than the original issue.

Implemented exactly as suggested in `cb783f8`:

- Dropped the separate rule at the end of `style.scss` along with its `!important`.
- Narrowed the existing rule instead:

```scss
.menu-item:not(.kb-nav-link-sub-click):not(.kb-nav-esc-close):hover > .sub-menu,
.menu-item.menu-item--toggled-on > .sub-menu {
	@include show-sub-menu();
}
```

Only the hover path is affected now, so activating the ▾ toggle after Escape opens the submenu normally and the visual state matches what is exposed to assistive tech.

Good point on the scoping too — `navigation-desktop-orientation-horizontal` sits on the wrapper at every viewport width, so the old rule never actually scoped to horizontal. Since `horizontal-navigation-base-styles` already carries the correct media queries, the narrowed rule inherits that scoping properly.

rahulacaleffi:
@Contributolo, thanks for quickly providing the updates.
We have this PR mapped internally, and this should be out in one of our next releases. Thank you, again!

rahulacaleffi (approved):

rahulacaleffi:
@Contributolo good work!

Contributolo:
Thanks for the review and the merge, @rahulacaleffi, much appreciated. The two remaining one-liners (#1259 for the sticky cleanup branch, #1224 for scroll spy) are ready whenever they fit into your queue.

PR #975, opened 2026-04-27
Fix custom color palette conflict with block themes (e.g. Blocksy)

ravinderk:
Resolve: [KAD-5436]

## Description

Fixes a conflict where custom block themes color palettes override by our plugin such as Blocksy. Ensures the correct color palettes are applied to color picker  in the editor and on the frontend.

## Changes
- Sync updated palette colors to both the legacy `colors` setting and `__experimentalFeatures.color.palette.theme` on color state change in `color-palette-defaults.js`
- Remove duplicate `updateSettings({ colors })` call from `saveConfig` (now handled reactively)
- Display theme color palette in Kadence color palette.

[KAD-5436]: https://stellarwp.atlassian.net/browse/KAD-5436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/plugins/block-defaults/color-palette-defaults.js
```diff
@@ -52,6 +53,22 @@ export default function KadenceColorDefault() {
...
+				existingFeatures.color.palette.theme = colors;
+				updateSettings({
+					colors, // Legacy support for color palette, will be removed in future versions.
+					__experimentalFeatures: existingFeatures, // Update the experimental features with the new color palette.
+				});
+			} catch (error) {
+				console.error('Failed to update colors', error);
+			}
+		}
+	}, [colors, existingFeatures, updateSettings]);
```
gilbert-hernandez: I'm not sure existingFeatures and updateSettings should be included in the dependency array as it could lead to an infinite loop. I think they should probably be removed.

src/plugins/block-defaults/color-palette-defaults.js
```diff
@@ -52,6 +53,22 @@ export default function KadenceColorDefault() {
...
 
+	// This effect will run whenever the colors state changes and will update the block editor settings with the new color palette.
+	// It included support for legacy color palette and new color palette, but the legacy support will be removed in future versions.
+	useEffect(() => {
+		if (colors) {
+			try {
+				existingFeatures.color.palette.theme = colors;
+				updateSettings({
+					colors, // Legacy support for color palette, will be removed in future versions.
+					__experimentalFeatures: existingFeatures, // Update the experimental features with the new color palette.
```
gilbert-hernandez: existingFeatures is a live reference to the Redux store's state object, obtained via useSelect. Mutating it directly is a React/Redux anti-pattern and undefined behavior. The store may or may not reflect the change depending on whether it does shallow or deep comparisons, and it bypasses immutability guarantees.               
                            
Maybe this instead:                                                                           
`updateSettings({`
`      colors,`                                                                                                      
`     __experimentalFeatures: {`
`         ...existingFeatures,` 
`          color: {`            
`              ...existingFeatures?.color,`                                                                          
`              palette: {`                 
`                  ...existingFeatures?.color?.palette,`                                                             
`                  theme: colors,`                      
`              },`                
`          },`    
`      },`    
`  });`

gilbert-hernandez (changes requested):

src/plugins/block-defaults/color-palette-defaults.js
ravinderk: I agree. I kept only `colors`  as dependency

src/plugins/block-defaults/color-palette-defaults.js
ravinderk: Good suggestion. Fixed.

gilbert-hernandez (approved):
LGTM

PR #987, opened 2026-05-19
Add scope attribute to the table block's header

pramodjodhani:
🎫  https://linear.app/nexcess/issue/SMTNC-219/add-scope-option-for-tables-when-headers-are-set

...

I realised that we already have a merged PR which does a similar job but I think that is too manual and time-consuming to assign the scope of each individual cell. Maybe we should undo this PR: https://github.com/stellarwp/kadence-blocks/pull/957

Loom: https://www.loom.com/share/7b9de9f9a62c4ded87ce5639849957fb

### Checklist


includes/blocks/class-kadence-blocks-table-data-block.php
```diff
@@ -93,35 +93,44 @@ public function build_css( $attributes, $css, $unique_id, $unique_style_id ) {
...
-		if ( 'th' === $tag && ! empty( $attributes['scope'] ) ) {
-			$allowed_scopes = [ 'col', 'row' ];
-			if ( in_array( $attributes['scope'], $allowed_scopes, true ) ) {
-				$scope_attr = ' scope="' . esc_attr( $attributes['scope'] ) . '"';
-			}
-		}
+		$scope = $this->get_block_header_scope( $attributes, $block_instance->context );
+		$tag   = $scope ? 'th' : 'td';
+
+		$scope_attr = $scope ? sprintf( ' scope="%s"', esc_attr( $scope ) ) : '';
```
WayneRocha, line 99:
```suggestion
		$scope_attr = apply_filters(
            'kadence_blocks_table_data_scope_attr',
            $scope ? sprintf( ' scope="%s"', esc_attr( $scope ) ) : '',
            $scope,
            $attributes,
            $unique_id
        );
```

`build_html()` is a public method whose rendered HTML output now differs from shipped behavior (adds `scope="col"` / `scope="row"` to `<th>` elements).
Downstream consumers — child themes with attribute-targeted CSS, JS reading the DOM, or snapshot/acceptance tests — can break silently.
A filter gives them an escape hatch without blocking the accessibility gain.

includes/blocks/class-kadence-blocks-table-data-block.php
```diff
@@ -93,35 +93,44 @@ public function build_css( $attributes, $css, $unique_id, $unique_style_id ) {
...
 	}
 
-	private function is_this_block_header( $attributes, $context ) {
-		$isFirstRowHeader = ! empty( $context['kadence/table/isFirstRowHeader'] );
-		$isFirstColumnHeader = ! empty( $context['kadence/table/isFirstColumnHeader'] );
+	/**
+	 * Get the header scope for a table data cell.
+	 *
+	 * @param array $attributes Block attributes.
+	 * @param array $context      Block context.
```
WayneRocha:
```suggestion
	 * @param array $context    Block context.
```

WayneRocha (changes requested):
Just a backward compatibility note

includes/blocks/class-kadence-blocks-table-data-block.php
```diff
@@ -93,35 +93,44 @@ public function build_css( $attributes, $css, $unique_id, $unique_style_id ) {
...
+	 * @return string|false 'col', 'row', or false when the cell is not a header.
+	 */
+	private function get_block_header_scope( $attributes, $context ) {
```
gilbert-hernandez, line 120: We don't have any workflow for this like I've seen in Give and LD. However, we should add the `@since TBD` for this block as Eric had mentioned in slack.

gilbert-hernandez (approved):

pattonwebz:
I am all for simplicity, and this does simplify the implementation I provided a while back, but it also lacks the ability to change a single row or column scope, which was the solution I was originally needing.

I saw the path that would auto-set the value when row or column headers were selected, but I specifically chose not to do that for backwards compatibility.

includes/blocks/class-kadence-blocks-table-data-block.php
pramodjodhani: Thanks, good catch. We should always add this.

pramodjodhani: Fixed: https://github.com/stellarwp/kadence-blocks/pull/987/commits/a16fe16b53bb1a99b4bb9ca06e32d6b7b71baf18

pramodjodhani:
@pattonwebz Thanks for your reply. I'm going to replace your PR with this one so its a simpler implementation for most Table block users. 

However, you can use the [kadence_blocks_table_data_scope_attributes](https://github.com/stellarwp/kadence-blocks/pull/987/changes) filter and some JS filters to add custom custom to table kadence/table-data ([reference](https://awhitepixel.com/add-custom-settings-to-existing-wordpress-gutenberg-blocks/)).

includes/blocks/class-kadence-blocks-table-data-block.php
pramodjodhani: @WayneRocha Thanks. I added the filter to the `get_block_header_scope` function as that is the single source of truth.

pramodjodhani: Commit: https://github.com/stellarwp/kadence-blocks/pull/987/changes/a16fe16b53bb1a99b4bb9ca06e32d6b7b71baf18

pattonwebz:
> @pattonwebz Thanks for your reply. I'm going to replace your PR with this one so its a simpler implementation for most Table block users. 
> 
> However, you can use the [kadence_blocks_table_data_scope_attributes](https://github.com/stellarwp/kadence-blocks/pull/987/changes) filter and some JS filters to add custom custom to table kadence/table-data ([reference](https://awhitepixel.com/add-custom-settings-to-existing-wordpress-gutenberg-blocks/)). 

Sounds good to me, thanks for the reference links. Those look like they solve my original needs :)

WayneRocha (approved):

PR #991, opened 2026-05-27
feat: pass unified LWSW license key for pattern imports (CONS-371)

jonwaldstein:
## Summary

**This adds support for LWSW licenses for interactions with `{*}.startertemplatecloud.com/wp-json`. The rest of the PR description is AI generated:**

Two commits, both scoped to `includes/helper-functions.php`:

**1. Pass the unified LWSW key when no legacy key exists** — `kadence_blocks_get_current_license_key()` now falls back to `lw_harbor_get_unified_license_key()` so unified-license customers' keys reach the pattern-server validators.

**2. Filter to an *authorized* key before shipping to remotes** — new `kadence_blocks_get_authorized_license_key()` runs the same local `is_authorized()` / `lw_harbor_is_product_license_active( 'kadence' )` checks already used elsewhere (no extra network calls) so the editor and REST controller only send keys that are actually entitled. This fixes two mixed-license edge cases:

- Legacy key still stored but no longer authorized (e.g. customer migrated to LWSW) → stale key would otherwise be sent to the legacy server and rejected with `"Invalid or expired license key"` even though the LWSW key is valid.
- LWSW key without the `kadence` product entitlement → same generic invalid-license error downstream. Now `_data()` returns empty key, surfacing a clearer "no license" state.

`kadence_blocks_get_current_license_key()` is left as a "what's stored" primitive, so the settings UI and `kadence_blocks_is_legacy_license_authorized()` keep their existing semantics.

Pairs with the LWSW dispatch added in three server-side repos (each its own PR on `feature/CONS-371`):

- `stellarwp/kadence-cloud-kadencewp-license` — gates `/kadence-cloud/v1/single` (the Pricing Table 11 path)
- `stellarwp/kadence-blocks-plugin-endpoints` — gates the SVG endpoint and other relay consumers
- `stellarwp/kadence-starter-api` — gates full-site starter template imports

Until at least one of the server-side PRs lands, this plugin change has no user-visible effect (unified keys will reach the server but still be rejected). Legacy customers with valid authorization see no behavior change at any rollout stage.

## Test plan

- [ ] Premium Kadence patterns, starter templates, and icons can be installed with an active LW license

## DEMO

https://www.loom.com/share/f3a5fe6ce2c34aa588143dbc63daa13d


gilbert-hernandez (approved):
LGTM. No blockers from me, but I'm not really sure how to go about testing it. I will also say that I recently added that "Guard against TBD" workflow recently and that I haven't had a chance to test the "Replace  TBD" workflow yet.

PR #992, opened 2026-05-29
Add package-lock.json and update @kadence/components version to 1.1.9

gilbert-hernandez:
🎫  #[Jira Ticket]
Summary
  
  Restores the flex alignment control types that the Kadence Column / Section block relies on but
  that were dropped when its controls moved to this shared component.

  Problem

  In the Section block's Flex Settings, the Direction, Alignment, and Vertical Alignment controls
  were all rendering as a 3-option text-alignment toolbar (left / center / right) instead of their
  proper, distinct controls.

image

  Root Cause

  ResponsiveAlignControls renders a toolbar based on its type prop. This shared component only
  handled justify, vertical, orientation, and vertical-column. The Section block (kadence/column)
  also passes orientation-column, justify-align, justify-column, and justify-vertical — the last
  three honoring a reverse prop. Those types weren't recognized here, so they silently fell through
  to the default AlignmentToolbar, producing the 3-option text-alignment control for all of them.

  These branches previously lived in a local component inside kadence-blocks; they were lost when the
  block was migrated to consume this shared package, which never implemented them.

  Changes

  - Added the four missing type branches: orientation-column, justify-align, justify-column,
  justify-vertical.
  - Added the reverse prop and its inverted control ordering for the justify-* variants.
  - Added the required icon imports (justifyLeft, justifyCenter, justifyRight, justifySpaceBetween,
  justifyStretch).
  - Restored the per-type kb-responsive-align-${type} className so existing CSS hooks continue to
  apply.
  - Bumped version to 1.1.9.

  Behavior for the existing types (justify, vertical, orientation, vertical-column, default
  textAlign) is unchanged.

  Testing
Open an editor page. Add a Section and select it in the list view. View the Flex Settings in the General tab. They should now have the correct icons and choices for the Direction, Alignment, and Vertical Alignments as pictured below. 

https://www.loom.com/i/90dde1fa6a7444668097ea07b8adc39e
...

### Checklist


gilbert-hernandez:
There may be more to this because of the Linting. The linting deletes the package-lock.json file. Let me gather that information on why it does that.

glaubersilva (approved):

PR #998, opened 2026-06-01
Add snapshot test suite for DTCG documents

d4mation:
## Summary

Stands up a dedicated Codeception **`snapshot`** test suite so that snapshot tests can be written for generated DTCG (Design Tokens Community Group) documents as part of the Kadence AI -> Design System project.

This PR is **infrastructure only** -- no DTCG generation code exists in the repo yet, so no DTCG/example tests are included. Real snapshot tests get added once the generator lands.

## What's included

- **Dependency**: adds dev dependency `lucatume/wp-snaphot-assertions:^2` (a wp-browser/Codeception adapter around `spatie/phpunit-snapshot-assertions`). Resolves cleanly against the existing `lucatume/wp-browser:<3.5` constraint. Dev-only, so Strauss does not prefix it.
- **Suite config**: `tests/snapshot.suite.yml` (boots WordPress via `WPLoader`, actor `SnapshotTester`, theme `twentytwentyfour`).
- **Bootstrap glue**: `tests/snapshot/_bootstrap.php` translates Codeception's `--debug` into spatie's `--update-snapshots`, so `slic run snapshot --debug` (re)generates snapshots.
- **Actor + helper**: `tests/_support/SnapshotTester.php` and `tests/_support/Helper/Snapshot.php`.
- **Base class**: `Tests\Support\Classes\SnapshotTestCase` using `Spatie\Snapshots\MatchesSnapshots`, with:
  - A documented `assertMatchesJsonSnapshot()` override. Spatie's built-in version decodes then re-encodes JSON before comparing, and that round-trip is lossy -- e.g. an empty object `{}` gets normalized to an empty array `[]`, failing an otherwise-identical document. Our override encodes the value to JSON once with `JSON_PRETTY_PRINT` and compares it as text (so `{}` stays `{}`), producing stable, diff-friendly `.txt` golden files. A pre-serialized JSON **string** is passed straight through to Spatie's original (kept callable via a trait alias, since `parent::` cannot reach a trait method).
  - An `assertSnapshotExists()` helper (ported from sfwd-lms) to guard data-provider-driven tests against silently passing when no snapshot was ever written.
- **CI**: adds a `Run snapshot tests` step to `.github/workflows/tests.yml`.

## Verification

Ran locally via slic against a throwaway test (removed before commit):

- `slic run snapshot --debug` -> created the snapshot; confirmed `{}` is preserved (not coerced to `[]`).
- `slic run snapshot` -> matched the stored snapshot (OK, 1 test, 1 assertion).
- Empty suite (after removing the throwaway test) -> `No tests executed!`, exit 0, so the new CI step is green until real tests are added.

## How a future DTCG snapshot test will look

```php
namespace Tests\snapshot;

use Tests\Support\Classes\SnapshotTestCase;

class DtcgDocumentTest extends SnapshotTestCase {
    public function test_generated_dtcg_document(): void {
        $document = /* generator output as an array */;
        $this->assertMatchesJsonSnapshot( $document );
    }
}
```

### DIFF
--- a/tests/_support/Classes/SnapshotTestCase.php
+++ b/tests/_support/Classes/SnapshotTestCase.php
@@ -1,0 +1,99 @@
+<?php declare( strict_types=1 );
+
+namespace Tests\Support\Classes;
+
+use Codeception\TestCase\WPTestCase;
+use Spatie\Snapshots;
+use Spatie\Snapshots\MatchesSnapshots;
+
+/**
+ * Base test case for snapshot tests.
+ *
+ * Boots WordPress (via the `snapshot` Codeception suite) and asserts generated output
+ * against stored snapshots in sibling `__snapshots__/` directories. Run with
+ * `slic run snapshot --debug` to (re)generate snapshots.
+ */
+class SnapshotTestCase extends WPTestCase {
+
+	/*
+	 * Alias the trait's JSON assertion so the original implementation stays callable from
+	 * within our override below ( `parent::` cannot reach a trait method ).
+	 */
+	use MatchesSnapshots {
+		assertMatchesJsonSnapshot as protected spatieAssertMatchesJsonSnapshot;
+	}
+
+	/**
+	 * Assert that a value matches its stored JSON snapshot.
+	 *
+	 * Overrides Spatie\Snapshots\MatchesSnapshots::assertMatchesJsonSnapshot().
+	 *
+	 * The Spatie version expects a pre-serialized JSON string and its JSON driver
+	 * decodes then re-encodes both sides before comparing; that round-trip is
+	 * lossy -- e.g. an empty object {} is normalized to an empty array [] -- which
+	 * fails an otherwise-identical document.
+	 *
+	 * For a non-string value this version encodes it to JSON once with
+	 * JSON_PRETTY_PRINT and compares it as text, so structures like {} are preserved
+	 * exactly. A string is assumed to be already-serialized JSON and is passed through to
+	 * Spatie's original implementation ( via the aliased spatieAssertMatchesJsonSnapshot() ).
+	 *
+	 * @param mixed $data Data to snapshot. Non-strings are JSON-encoded; a string is
+	 *                    treated as pre-serialized JSON and handled by Spatie.
+	 *
+	 * @return void
+	 */
+	protected function assertMatchesJsonSnapshot( $data ): void {
+		if ( is_string( $data ) ) {
+			$this->spatieAssertMatchesJsonSnapshot( $data );
+
+			return;
+		}
+
+		$this->assertMatchesSnapshot( wp_json_encode( $data, JSON_PRETTY_PRINT ) );
+	}
+
+	/**
+	 * Fails the test if the snapshot does not already exist.
+	 *
+	 * Useful in scenarios where you may be using a Data Provider to generate snapshots and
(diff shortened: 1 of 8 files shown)


tests/_support/Helper/Snapshot.php
d4mation, line 1: Mirrors other `tests/_support/Helper/` files.

tests/_support/SnapshotTester.php
d4mation, line 1: Mirrors other `tests/_support/*Tester.php` files

tests/snapshot.suite.yml
```diff
@@ -0,0 +1,24 @@
...
+            wpRootFolder: %WP_ROOT_FOLDER%
+            dbName: %WP_TEST_DB_NAME%
+            dbHost: %WP_TEST_DB_HOST%
+            dbUser: %WP_TEST_DB_USER%
+            dbPassword: %WP_TEST_DB_PASSWORD%
+            tablePrefix: test_
+            domain: %WP_DOMAIN%
+            adminEmail: admin@kadence.test
+            title: 'Kadence Blocks Snapshot Tests'
+            theme: twentytwentyfour
```
d4mation, line 20: Because of the versions tested by CI, `twentytwentyfour` is the newest Theme we can utilize without manually installing a newer theme (and then potentially having incompatibility when used with an older WP version).

tests/_support/Classes/SnapshotTestCase.php
```diff
@@ -0,0 +1,99 @@
...
+	 */
+	protected function assertMatchesJsonSnapshot( $data ): void {
+		if ( is_string( $data ) ) {
+			$this->spatieAssertMatchesJsonSnapshot( $data );
+
+			return;
+		}
+
+		$this->assertMatchesSnapshot( wp_json_encode( $data, JSON_PRETTY_PRINT ) );
+	}
```
d4mation: This is to account for some weird behavior we noticed in LearnDash as part of the MCP project over there.

https://github.com/stellarwp/learndash-core/pull/2466#discussion_r2308332748

pauloiankoski (approved):
This looks good overall. The only suggestion I have would be to add a tiny self-test under `tests/snapshot/` that snapshots an array containing an empty object (for example `['meta' => new \stdClass()]`, or an empty array encoded as an object) and verifies the generated `.txt` keeps `{}` correctly.

Besides covering this edge case explicitly, it would also make the new CI step execute a real test instead of reporting "No tests executed!".

d4mation:
> This looks good overall. The only suggestion I have would be to add a tiny self-test under `tests/snapshot/` that snapshots an array containing an empty object (for example `['meta' => new \stdClass()]`, or an empty array encoded as an object) and verifies the generated `.txt` keeps `{}` correctly.
> 
> Besides covering this edge case explicitly, it would also make the new CI step execute a real test instead of reporting "No tests executed!".

@pauloiankoski I could add a test for it, but it seems strange to have a test to test a test method 😅 

We effectively have it tested in [this PR](https://github.com/stellarwp/learndash-core/pull/3062) so I'm not worried about it.

We'll have tests for the DTCG output using this before long, so I'm not concerned about it finding no tests at the moment.

I'm going to merge this one in particular to the Feature branch rather than `master` as I think it is less likely to be needed for maintenance. But we could always easily port it out later if I end up being wrong about that :)

PR #999, opened 2026-06-01
[DTM] Added design tokens table and Token_Store

dave-green-uk:
🎫 SOFT-3375

## Summary

Adds the persistence layer for the Design Tokens system: a kb_design_tokens database table and a Token_Store gateway class.

The store holds an overrides-only DTCG document per token set — an empty or absent row means the site renders entirely from baseline tokens. Each write bumps a cache-busting version hash and fires a change action so downstream caches/projectors can react.

## Changes

- Token_Table — new custom table (kb_design_tokens). Unlike the Optimizer tables, it holds authored user data, so it uses the base class's non-destructive dbDelta migration (no drop-on-schema-change).
- Token_Store — the sole gateway to the table: get_document(), get_version(), save_document(), bump_version(). Validation is deliberately left to the REST layer; this class only persists and signals change.
- Wiring — Design_Tokens_Provider + Database\Provider register the store in the container; the table is registered centrally in Database_Provider.
- Write integrity — save_document() / bump_version() return void and fire the change action only after a successful write. StellarWP's DB layer throws on a failed write, so the action is never reached on error — a failure is never advertised as a change.
- Tests — Token_StoreTest covers read/write, version bumping, slug isolation, title preservation, the change action, and the no-fire-on-error failure path.

## Test plan (for the future)

- [ ] On a fresh install, the kb_design_tokens table is created on
  activation.
- [ ] Saving a document persists it and changes version; saving
  again updates the same row with a new version.
- [ ] Bumping SCHEMA_VERSION triggers a clean dbDelta ALTER (no
  data loss, no repeated ALTER attempts).
- [ ] kadence_blocks_design_tokens_changed fires on a successful
  save/bump only.


includes/resources/Design_Tokens/Database/Token_Table.php
```diff
@@ -0,0 +1,61 @@
...
+	/**
+	 * @inheritDoc
+	 */
+	protected function get_definition(): string {
+		global $wpdb;
+		$table_name      = self::table_name();
+		$charset_collate = $wpdb->get_charset_collate();
+
+		return "
+			CREATE TABLE `$table_name` (
```
dave-green-uk, line 66: Any changes needed here?

pauloiankoski: No, the definition is correct. The only implicit contract is that `document`/`updated_at` are NOT NULL with no default, which is fine as long as `Token_Store` stays the sole writer (which it's documented to be).

pauloiankoski (approved):
The code looks good. There may be only one issue with the PR description, though. Where it says:

> Write integrity — save_document() / bump_version() return bool and only fire CHANGED_ACTION on a successful write, so a failed DB write is never advertised as a change.

Those methods actually return `void`, not `bool`.

includes/resources/Design_Tokens/Database/Design_Tokens_Table.php
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * The Design Tokens database table.
+ *
+ * Stores the overrides-only DTCG document per token set. An empty (or absent)
+ * document means the site renders entirely from the baseline tokens.
+ *
+ * Unlike the Optimizer tables, this holds authored user data, so it must NOT
+ * drop itself on a schema change — the base class migration is non-destructive.
+ */
+final class Design_Tokens_Table extends Table {
```
d4mation: I left it for `Design_Tokens_Provider` because that's how the rest of the plugin is written, but does this _need to be_ `Design_Tokens_Table`? It is already getting `Design_Tokens` from the Namespace. 

`Token_Table` could be fine if we want to be a little more specific, but having all of `Design_Tokens_` just feels like repeating ourselves 😅

includes/resources/Design_Tokens/Database/Token_Table.php
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Stores the overrides-only DTCG document per token set. An empty (or absent)
+ * document means the site renders entirely from the baseline tokens.
+ *
+ * Unlike the Optimizer tables, this holds authored user data, so it must NOT
+ * drop itself on a schema change — the base class migration is non-destructive.
+ */
+final class Design_Tokens_Table extends Table {
+
+	public const SCHEMA_VERSION = '1.0.0';
```
d4mation, line 25: We should add a variable doc here.

It is worthwhile to have `@since TBD` in variable docs so we can easily know when they were introduced.

includes/resources/Design_Tokens/Database/Token_Table.php
```diff
@@ -0,0 +1,61 @@
...
+
+	/**
+	 * @var string The field that uniquely identifies a row in the table.
+	 */
+	protected static $uid_column = 'id';
+
+	/**
+	 * @inheritDoc
+	 */
+	protected function get_definition(): string {
```
d4mation, line 60: `@inheritDoc` is fine, but we should add a `@since TBD`

includes/resources/Design_Tokens/Database/Provider.php
```diff
@@ -0,0 +1,26 @@
...
+		$this->register_token_store();
+	}
+
+	private function register_token_store(): void {
+		$this->container->singleton( Token_Store::class, Token_Store::class );
+
+		$this->container->when( Token_Store::class )
+						->needs( '$table' )
+						->give( static fn(): string => Design_Tokens_Table::table_name( false ) );
+	}
```
d4mation, line 39: We need to add phpdocs to these methods.

includes/resources/Design_Tokens/Database/Token_Store.php
```diff
@@ -0,0 +1,169 @@
...
+
+	/**
+	 * Fires after any write so projectors and caches can react.
+	 */
+	public const CHANGED_ACTION = 'kadence_blocks_design_tokens_changed';
+
+	/**
+	 * The default token set slug. v1 ships a single set under this slug.
+	 */
+	public const DEFAULT_SLUG = 'default';
```
d4mation: We should avoid `public const`. They cannot be deprecated cleanly if we ever need to in the future. If we need these publicly accessible, then changing these to `protected`/`private` and then having a `public static` method to retrieve them would be better.

Also, we need to ensure we add `@since TBD` and otherwise complete the variable doc here so that the type is defined, etc.

(I did notice that for the DB Table class we are inheriting the `public const` so I grumbled about it and left it alone, but we should avoid it where we can).

includes/resources/Design_Tokens/Database/Token_Store.php
```diff
@@ -0,0 +1,169 @@
...
+
+	/**
+	 * Read the raw overrides-only DTCG document for a token set.
+	 *
+	 * @param string $slug The token set slug.
+	 *
+	 * @return string The raw DTCG JSON, or an empty string when no row exists
+	 *                (caller should then render entirely from baseline).
+	 */
```
d4mation, line 69: We need to ensure all methods have `@since TBD`

includes/resources/Design_Tokens/Database/Design_Tokens_Table.php
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * The Design Tokens database table.
+ *
+ * Stores the overrides-only DTCG document per token set. An empty (or absent)
+ * document means the site renders entirely from the baseline tokens.
+ *
+ * Unlike the Optimizer tables, this holds authored user data, so it must NOT
+ * drop itself on a schema change — the base class migration is non-destructive.
+ */
```
d4mation, line 17: `@since TBD` for classes is a good idea too.

d4mation (approved):

PR #1001, opened 2026-06-01
Set up PHPStan static analysis

d4mation:
## Summary

Sets up PHPStan static analysis for the repository, following a configuration similar to `kadence-shop-kit`. Targets `feature/design-tokens-module`.

### What's included

- **`phpstan.neon`** — level `max`, analysing `includes/`, `kadence-blocks.php` and `uninstall.php`. Excludes `includes/assets/*` and the generated `includes/*-array.php` icon/font data files (large literal arrays that choke analysis at level max with nothing meaningful to check).
- **`.phpstan/`** — stubs for runtime-defined constants (`KADENCE_BLOCKS_*`) and the WordPress string polyfills (`str_contains` / `str_starts_with` / `str_ends_with`) that ship in WP core but are missing from `php-stubs/wordpress-stubs`.
- **`phpstan-baseline.neon`** — generated against the **`master`** branch state (~4,900 pre-existing items baselined) so only legacy debt is suppressed and new code is held to the standard.
- **Composer** — adds `phpstan/extension-installer` and `szepeviktor/phpstan-wordpress` dev deps, plus `phpstan` / `phpstan-baseline` scripts.
- **CI** — `.github/workflows/phpstan.yml` runs PHPStan on pull requests.
- **`.distignore` / `.gitignore`** — exclude the new config from the built plugin zip and ignore `phpstan-cache/`.

### Bugfixes (separate commits)

PHPStan reports a small class of **non-ignorable** errors that cannot be placed in the baseline and therefore *must* be fixed in code before analysis can pass. It surfaced two genuine latent bugs, each committed on its own:

1. **`current_time()` called with no arguments** in `class-kadence-blocks-prebuilt-library.php` — `current_time( $type, $gmt = 0 )` requires `$type`, so this would throw an `ArgumentCountError` at runtime. Fixed to use the same format string as the call a few lines above.
2. **`$classes[] .= ' ' . $attributes['className']`** in `class-kadence-blocks-posts-block.php` — `$classes[]` appends a new array element and `.=` immediately reads from it, which is invalid ("Cannot use [] for reading"). The effect was that the custom CSS class was never actually added. Fixed to assign the class name as a new element (it is imploded with spaces downstream).

**IMPORTANT** These are very minor and innocuous changes and the `current_time()` fix is required to prevent phpstan from straight up crashing, so these are going to be merged directly to `master` instead of a maintenance release to avoid possible issues.

## Notes

- The baseline was generated against `master` rather than this branch, so any PHPStan errors introduced by the design-tokens work surface in CI rather than being silently baselined.

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -34,3 +34,7 @@ codeception.slic.yml
 .wordpress-org/
 .npmrc
 bunfig.toml
+.phpstan/
+phpstan-baseline.neon
+phpstan-cache/
+phpstan.neon
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -1,0 +1,59 @@
+name: Run PHPStan on pull requests
+on: pull_request
+jobs:
+  phpstan:
+    runs-on: ubuntu-latest
+    env:
+        COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0 # fetch all history.
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+          coverage: none
+          tools: composer
+
+      - name: Configure Composer Auth
+        run: composer config --global github-oauth.github.com ${{ secrets.GH_BOT_TOKEN }}
+
+      - name: Cache Dependencies
+        id: deps-cache
+        uses: actions/cache@v5
+        with:
+          path: |
+            vendor
+            vendor/vendor-prefixed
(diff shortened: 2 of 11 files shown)


pauloiankoski (approved):
Looks good to me overall. I only left a few small non-blocking workflow suggestions around the new CI file, but nothing that should block merging this.

- Every other workflow in this repo uses `actions/checkout@v4`/`@v6`; v2 runs on a deprecated Node runtime and emits GitHub deprecation warnings. Bump to @v6 for consistency and longevity.
- `env.COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}` is declared but never referenced; auth is actually configured via `composer config --global github-oauth.github.com ${{ secrets.GH_BOT_TOKEN }}`. Remove the unused env to avoid confusion.
- The workflow triggers on every `pull_request`; `tests.yml` restricts to `**.php` / `composer.lock` / its own workflow file. Adding a similar `paths:` filter avoids running static analysis on PRs that only touch JS/CSS/docs and saves CI minutes.

pauloiankoski (approved):

PR #1002, opened 2026-06-01
feat: PHPCS CI workflow for changed lines

d4mation:
## Summary

Adds a PHPCS CI check that runs against **only the lines changed** by a branch, modeled on LearnDash's (`sfwd-lms`) `ci-standards.yml` approach. A branch is held to the coding standard for what it touches, without failing on pre-existing violations elsewhere in a touched file.

Deliberately **does not** use ReviewDog (as `kadence-shop-kit` does), since ReviewDog makes it hard to see all issues in CI once there are more than ~30 of them. Here every remaining issue is printed in full to the job log.

> **Note:** This PR is expected to show the PHPCS check as *not run* - it modifies no `*.php` files, and the workflow is gated to changes touching `**.php`. It will run on subsequent PRs that change PHP.

## Changes

- **`.github/workflows/phpcs.yml`** - runs on every PR (and on pushes) touching `**.php`. Fetches the base branch and runs `composer phpcs-ci "origin/$BASE_REF"`. Uses the same PHP/Composer setup as the existing `compatibility.yml` / `tests.yml` workflows.
- **`dev_scripts/partial-phpcs.sh`** - runs phpcs (JSON report) on the changed `.php` files, then `vendor/bin/diffFilter --phpcs` narrows the report down to the lines present in the diff. Exits non-zero on any remaining issue.
- **`composer.json` / `composer.lock`** - adds the `exussum12/coverage-checker` dev dependency (provides `diffFilter`, same as LearnDash) and `phpcs-ci`, `phpcs-partial`, and `phpcs-autofix` scripts.
- **`.gitignore` / `.distignore`** - ignore the generated `diff.txt`, `diffcs.txt`, and `phpcs.json` artifacts (and keep `dev_scripts/` out of the dist zip).

## Notes

- Installing `exussum12/coverage-checker` required downgrading the transitive dev dependency `nikic/php-parser` from v5.7.0 to v4.19.5 (the package only supports v3/v4). This is safe: php-parser is only consumed by `phpunit/php-code-coverage`, which supports v4, and PHPStan uses its own bundled parser.
- Added `-q` to the phpcs invocation so deprecation notices from the ruleset don't pollute the JSON report.

## Testing

Verified locally against the branch diff:
- Clean changed lines -> exit 0 (`100.00% Covered`).
- Injected a violation on a changed line -> exit non-zero, with each issue printed per line; then reverted.


dev_scripts/partial-phpcs.sh
pauloiankoski, line 1: I think there’s one edge case here that could end up reporting a false pass.

Right now the script does not use `set -euo pipefail`, so if `git diff "$BASE_BRANCH"` fails (for example because the base ref was not fetched, there is a typo, or due to a shallow-fetch issue), the error gets printed but the script keeps running.

That means this line:

```bash id="8dw0lu"
git diff "$BASE_BRANCH" --name-only ... > diffcs.txt
```

can produce an empty `diffcs.txt`, and then the next guard hits:

```bash id="c8t9iv"
if [ ! -s "$DIFF_PHPCS_FILE" ]
```

which exits with `"No php files to check."` and status `0`.

So in practice, a broken or missing base ref can still produce a green check, which feels risky for a guardrail script like this.

Maybe we could add `set -euo pipefail` near the top and also verify the base ref before running the diff, something like:

```bash id="h4prjk"
git rev-parse --verify --quiet "$BASE_BRANCH" >/dev/null \
  || { echo "Base ref '$BASE_BRANCH' not found"; exit 1; }
```

I think the existing `|| true` on the `phpcs` command still makes sense to keep, since `phpcs` returning non-zero for detected issues is expected behavior, and the final `diffFilter` exit code can still remain the script result.

composer.json
```diff
@@ -133,6 +134,9 @@
 			"@compatibility:php-8.4"
 		],
 		"phpcs": "phpcs --standard=phpcs.xml",
+		"phpcs-autofix": "phpcbf --standard=phpcs.xml",
+		"phpcs-partial": "dev_scripts/partial-phpcs.sh selected",
```
pauloiankoski: In selected mode the script requires `$3` (the newline-separated file list) and exits with status `1` otherwise. Running `composer phpcs-partial` with no extra args always fails with `"File list is required..."`. Either document the required invocation (`composer phpcs-partial origin/master $'file1\nfile2'`) in the script header / PR, or reconsider whether selected mode belongs in a composer alias at all. As shipped, the alias only ever errors.

dev_scripts/partial-phpcs.sh
d4mation: https://github.com/stellarwp/kadence-blocks/pull/1002/commits/ee0862b93b58ab3b011f09ea7e3121fce1166324

composer.json
d4mation: https://github.com/stellarwp/kadence-blocks/pull/1002/commits/ee0862b93b58ab3b011f09ea7e3121fce1166324

pauloiankoski (approved):

PR #1000, opened 2026-06-01
Add CONTRIBUTING.md file with setup instructions for local development

gilbert-hernandez:
[[SMTNC-1435]](https://linear.app/nexcess/issue/SMTNC-1435/complete-the-bun-migration-unify-dependency-tooling-and-ship-a)

This PR migrates the Kadence Blocks JavaScript install/build workflow from npm to Bun.
- Replaces npm-based install/build steps in CI, release, plugin check, POT generation, Ghost Inspector, and PUP zip builds with `bun install --frozen-lockfile` and `bun run build`.
- Adds `bun.lock` and `bunfig.toml` so dependency installs are locked and the `@stellarwp` scope resolves through GitHub Packages using `BUN_AUTH_TOKEN`.
- Adds a root `postinstall` step that builds the fetched `@kadence/helpers`, `@kadence/icons`, and `@kadence/components` packages when their compiled output is missing.
- Removes reliance on `package-lock.json` and documents the Bun-based setup flow in `CONTRIBUTING.md`.
## Notes
Bun does not build the fetched `@kadence/*` GitHub packages the same way npm previously did, so the package build loop remains in `package.json` and uses npm inside those package directories. CI/release workflows now rely on the root `postinstall` to run that step consistently after `bun install`.
## Testing
- Confirmed workflows and PUP config now use Bun for dependency installation and builds.
- Confirmed local development docs were updated to use `bun install`, `bun run start`, and `bun run build`.
- CI should verify the full build, lint, plugin check, release, and packaging paths....

### Checklist
- [x] I have performed a self-review.
- [x] No unrelated files are modified.
- [x] No debugging statements exist (Ex: console.log, error_log).
- [x] There are no warnings or notices in the wordpress error log.
- [ ] Passes all tests (linting, acceptance, & unit)

### Block specific checklist (where relevant)
- [ ] Tested with an existing instance of this block .
- [ ] Tested creating a new instance of this block.
- [ ] Tested with Dynamic content & Elements.

PR #1003, opened 2026-06-01
feat: add cSpell CI check for changed files

d4mation:
## Summary

Adds a cSpell CI check that spell-checks **only the files changed in a PR**, so the repo's pre-existing spelling issues don't block anything, but new code is caught.

Referenced the `spell-check` job from sfwd-lms's `ci-standards.yml` and the cspell config from `kadence-shop-kit`.

## Changes

- **`.github/workflows/cspell.yml`** - new workflow, triggered on `pull_request`. Diffs against the base branch and runs cspell only on changed files (`--diff-filter=ACMRTUXB`). Uses Bun + the runner's default Node 22.
- **`.cspell.json`** - config seeded from kadence-shop-kit, plus the project/dependency terms used in the files this PR touches.
- **`package.json`** - adds `cspell-cmd` / `cspell-full` scripts and `cspell` devDependency.
- **`.nvmrc`** - `22`, since cspell 10 requires Node >= 22.18 (for local dev consistency).
- **`.distignore`** - excludes `.cspell.json` and `.nvmrc` from the distributed plugin zip.

## Notes

- It is expected that there are many pre-existing spelling issues; this intentionally only checks changed files to avoid blocking on legacy problems while preventing **new** issues.
- Verified locally on Node 22: cspell 10 runs cleanly, and all files changed in this PR pass with 0 issues.


package.json
```diff
@@ -67,6 +69,7 @@
 		"@wordpress/scripts": "^30.19.0",
 		"ajv": "^8.17.1",
 		"concurrently": "^9.2.0",
+		"cspell": "^10.0.1",
```
pauloiankoski, line 72: Ideally, whenever we add or update a package here, we should also run `npm install` to keep the lock file in sync, especially because `npm ci` strictly follows `package-lock.json`.

It may also be worth taking a look at [PR #1000](https://github.com/stellarwp/kadence-blocks/pull/1000), which is replacing `npm` with `bun`. If the project fully migrates to Bun, it might make sense to remove `package-lock.json` entirely and rely on Bun's lockfile instead.

.cspell.json
```diff
@@ -0,0 +1,28 @@
+{
+    "version": "0.2",
+    "words": [
```
pauloiankoski, line 17: We could also add an `ignorePaths` section to `.cspell.json` to avoid spell-checking generated, binary, vendor, and lock files. Otherwise, routine PRs that touch those files will likely fail the check with false positives, making the workflow noisy and unreliable over time.


```suggestion
    "version": "0.2",
    "ignorePaths": [
        "package-lock.json",
        "bun.lockb",
        "composer.lock",
        "build/**",
        "**/*.min.js",
        "**/*.asset.php",
        "**/*.pot",
        "languages/**",
        "vendor/**",
        "node_modules/**",
        "**/*.svg"
    ],
    "words": [
```

.github/workflows/cspell.yml
```diff
@@ -0,0 +1,58 @@
...
+          restore-keys: |
+            ${{ runner.os }}-bun-
+
+      - name: Remove npm lockfile to prevent bun git-clone issues
+        run: rm -f package-lock.json
+
+      - name: Install dependencies
+        run: bun install
+        env:
+          BUN_AUTH_TOKEN: ${{ secrets.GH_BOT_TOKEN }}
```
pauloiankoski: If we are only checking for spelling errors here, we may not need the full dependency installation step nor the `GH_BOT_TOKEN`. A simple `bunx cspell lint $FILES` in the next step should be enough for this workflow. This could potentially reduce runtime.

.github/workflows/cspell.yml
```diff
@@ -0,0 +1,58 @@
+# Checks spelling on the lines/files changed in a pull request.
+# The repository is expected to have pre-existing spelling issues, so this
+# only checks files touched by the PR to avoid blocking on legacy problems
+# while still preventing new issues from being introduced.
+# cSpell:ignore ACMRTUXB
+name: cSpell
+run-name: '[cSpell] ${{ github.event.head_commit.message }}'
```
pauloiankoski: Since this workflow only runs on `pull_request`, `github.event.head_commit.message` will always be empty here because `head_commit` is only available for `push` events. Looks like this was copied from `eslint.yml`, where it makes sense.

Maybe we could switch this to something PR-specific instead, like:
 
```suggestion
run-name: '[cSpell] ${{ github.event.pull_request.title }}'
```

package.json
d4mation: This is bizarre honestly. `package-lock.json` is not versioned and the `bun.lock*` file patterns are explicitly `.gitignore`'d. 

CI technically calls `bunx` for this (via `bun run` in our npm script) so it doesn't _need_ to be installed, so this shouldn't block this PR as-is (the eslint CI check does the same thing) so I think we can technically ignore it for now. But once #1000 is merged we will want to ensure that cspell makes it into the lockfile. A small follow-up PR can do that :)

pauloiankoski:
@d4mation We just need to resolve the conflicts then I can approve the PR.

d4mation:
@pauloiankoski resolved!

pauloiankoski (approved):

pauloiankoski (approved):

PR #1006, opened 2026-06-01
[DTM] Token Registry + register_design_token() helper

pauloiankoski:
🎫 [SOFT-3379](https://linear.app/nexcess/issue/SOFT-3379/token-registry-register-design-token-helper)

Adds the **Token Registry** — the single declaration point for design tokens — and the `kadence_blocks_register_design_token()` helper, building on the Design Tokens Module (DTM) scaffold.

The Registry holds **structure only, never values**. Each token is declared once in PHP and the Registry records where it must surface: its `type`, `label`, UI `group`, and `projections` (WordPress preset slug, Kadence palette slot, dedicated `--kb-token--*` CSS var, Site Editor visibility). Because every projector and the admin UI read this one Registry, adding a token is a single declaration that automatically reaches all of them.

### What's included

- `Css_Var` — the deterministic `id → --kb-token--*` derivation rule (`semantic.color.button-bg → --kb-token--semantic--color--button-bg`).
- `Token_Definition` — immutable VO for a single token declaration (derives the css-var, supports an explicit override).
- `Token_Registry` — the service: `all() / get() / has() / by_type() / by_projection() / for_block() / css_var_for() / to_ui_schema()`, plus the `is_active()` projection gate.
- `kadence_blocks_register_design_token()` + `kadence_blocks_register_design_variant_set()` global helpers.
- `Baseline_Document` contract + a **fail-closed `Baseline_Guard`**: under `WP_DEBUG` it throws on a token missing a baseline entry; in production it logs, deactivates projection, and shows an admin notice (never white-screens a live site).
- `Variant_Set` skeleton recording which blocks accept variants.
- Full wpunit coverage (40 tests).

### Downstream interfaces exposed

| Consumer | Reads |
| --- | --- |
| SOFT-3380 Resolver | `all()`, `by_type()`, `css_var_for($id)` — must gate on `is_active()` |
| SOFT-3385 Admin UI feed | `to_ui_schema()` |
| SOFT-3393 Variant data model | `for_block()`, `register_variant_set()` |
| Projectors | `by_projection('wp_preset' \| 'kadence_slot' \| 'site_editor')` |

---

### ⚠️ NOTE — Temporary scaffolding in this PR

Everything below is intentionally temporary so the path can be proven end-to-end before its owning ticket lands. Each row names the issue that will fix/replace/remove it and the action that ticket will take.

| Temporary item | Where | Owning issue | Action to be taken |
| --- | --- | --- | --- |
| **`Always_Present_Baseline_Document` is bound as the baseline** (its `has()` always returns `true`), so the example tokens flow through registration + the guard now. The guard is effectively a no-op at runtime until this is swapped. | `Registry/Provider.php`, `Registry/Baseline/Always_Present_Baseline_Document.php` | **SOFT-3377** | Swap the binding to the real `Baseline_Document`. At that point every declared token must have a genuine baseline entry or the guard fires — the intended fail-closed behaviour. |
| **`Empty_Baseline_Document` stub** kept as the fail-closed reference (`has()` always `false`). | `Registry/Baseline/Empty_Baseline_Document.php` | **SOFT-3377** | Superseded by the real baseline document; remove once SOFT-3377's `has()` exists. |
| **Example token declarations** (`semantic.color.button-bg`, `semantic.color.button-text`) — a small representative set, not the real catalog. | `Registry/declarations.php` | **SOFT-3377** | Replace with the full token catalog supplied by the Baseline document. |
| **Guard runs inline during `register()`.** Works because the current stub is in-memory. | `Registry/Provider.php` (`guard_baseline()`) | **SOFT-3377** | When the real baseline loads from a store/file that needs WP loaded, move the guard invocation to a late `init` / `plugins_loaded` hook. The guard logic itself does not change — only *when* it is invoked. |
| **`Variant_Set` is a skeleton** — holds only `block` + `variants` names. The example Button variant set is a placeholder. | `Registry/Variant_Set.php`, `Registry/declarations.php` | **SOFT-3393** | Extend with the rich `$extensions…variants.<block>` payload shape, `$default` handling, and validation. |

### To confirm during review

- **CSS-var prefix:** implemented the **double-dash** `--kb-token--` per the design-doc §6 worked example (the prose says single dash). Flag if single-dash was intended.
- **`to_ui_schema()` shape** (`{ groups: { <group>: [ {id,type,label,cssVar,projections} ] } }`) is a proposal for SOFT-3385 — confirm it matches what the React app expects, or let SOFT-3385 finalize it.

### DIFF
--- a/includes/resources/Design_Tokens/Registry/Baseline_Guard.php
+++ b/includes/resources/Design_Tokens/Registry/Baseline_Guard.php
@@ -114,6 +114,12 @@ final class Baseline_Guard {
 		add_action(
 			'admin_notices',
 			static function () use ( $missing ): void {
+				// Only surface the misconfiguration to users who could act on it, and avoid leaking
+				// internal token ids to lower-privileged admin-area users.
+				if ( ! current_user_can( 'manage_options' ) ) {
+					return;
+				}
+
 				printf(
 					'<div class="notice notice-error"><p>%s</p></div>',
 					esc_html(
--- a/includes/resources/Design_Tokens/Registry/Baseline_Guard.php
+++ b/includes/resources/Design_Tokens/Registry/Baseline_Guard.php
@@ -1,0 +1,130 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Registry;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Contracts\Baseline_Document;
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Exception\Missing_Baseline_Entry;
+use KadenceWP\KadenceBlocks\Psr\Log\LoggerInterface;
+
+/**
+ * Enforces "every declared token must have a baseline entry."
+ *
+ * Policy:
+ *   - Always log the missing ids.
+ *   - Under WP_DEBUG: throw Missing_Baseline_Entry so dev/CI cannot miss the misconfiguration.
+ *   - In production: deactivate token projection (fall back to existing KB behaviour) + admin notice.
+ *     Never white-screen a live site.
+ *
+ * @since TBD
+ */
+final class Baseline_Guard {
+
+	private Token_Registry $registry;
+	private Baseline_Document $baseline;
+	private LoggerInterface $logger;
+
+	/**
+	 * Whether a missing entry throws. Resolves to WP_DEBUG when null (the production/dev default).
+	 *
+	 * @var bool|null
+	 */
+	private ?bool $throw_on_missing;
+
+	/**
+	 * @since TBD
+	 *
+	 * @param Token_Registry    $registry         The registry whose declared tokens are validated.
+	 * @param Baseline_Document $baseline         The baseline to validate the declared tokens against.
+	 * @param LoggerInterface   $logger           Logger for the missing-entry diagnostics.
+	 * @param bool|null         $throw_on_missing Force the throw-vs-degrade decision; null defers to
+	 *                                            WP_DEBUG. Exposed for testability; the container
+	 *                                            leaves it null.
+	 */
+	public function __construct(
+		Token_Registry $registry,
+		Baseline_Document $baseline,
+		LoggerInterface $logger,
+		?bool $throw_on_missing = null
+	) {
+		$this->registry         = $registry;
+		$this->baseline         = $baseline;
+		$this->logger           = $logger;
+		$this->throw_on_missing = $throw_on_missing;
+	}
+
+	/**
+	 * Run the guard: log, then either throw (debug) or fail closed (production) on missing entries.
+	 *
+	 * @since TBD
+	 *
+	 * @throws Missing_Baseline_Entry When WP_DEBUG is on and any declared token lacks a baseline entry.
+	 *
+	 * @return void
+	 */
+	public function run(): void {
+		$missing = $this->registry->missing_from_baseline( $this->baseline );
+
+		if ( $missing === [] ) {
+			return;
+		}
+
+		$message = sprintf(
+			'Design tokens disabled: %d declared token(s) have no baseline entry: %s',
+			count( $missing ),
+			implode( ', ', $missing )
+		);
+
+		$this->logger->error( $message );
+
+		if ( $this->should_throw() ) {
+			throw new Missing_Baseline_Entry( $message );
+		}
+
+		// Production: fail closed but stay up.
+		$this->registry->deactivate();
+		$this->register_admin_notice( $missing );
+	}
+
+	/**
+	 * Whether to throw on a misconfiguration. Honours an injected override, else defers to WP_DEBUG —
+	 * the seam lets tests exercise both branches without redefining the WP_DEBUG constant.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	private function should_throw(): bool {
+		if ( $this->throw_on_missing !== null ) {
+			return $this->throw_on_missing;
+		}
+
+		return defined( 'WP_DEBUG' ) && WP_DEBUG;
+	}
+
+	/**
+	 * Register the production admin notice explaining why token projection is disabled.
+	 *
+	 * @since TBD
+	 *
+	 * @param string[] $missing The token ids missing a baseline entry.
+	 *
+	 * @return void
+	 */
+	private function register_admin_notice( array $missing ): void {
+		add_action(
+			'admin_notices',
+			static function () use ( $missing ): void {
+				printf(
+					'<div class="notice notice-error"><p>%s</p></div>',
+					esc_html(
+						sprintf(
+							/* translators: %s: comma-separated token ids. */
+							__( 'Kadence Blocks design tokens are disabled because these tokens have no baseline entry: %s', 'kadence-blocks' ),
+							implode( ', ', $missing )
+						)
+					)
+				);
+			}
+		);
+	}
+}
--- a/includes/resources/Design_Tokens/Registry/Baseline_Guard.php
+++ b/includes/resources/Design_Tokens/Registry/Baseline_Guard.php
@@ -74,6 +74,10 @@ final class Baseline_Guard {
 			implode( ', ', $missing )
 		);
 
+		// TODO (SOFT-3377): the guard moves off `init` and runs against the real baseline, so this would
+		// otherwise log once per front-end/admin/REST/AJAX/cron request on a misconfigured site. A missing
+		// baseline entry is a deterministic deploy-time error (identical every request), so throttle to
+		// once per process with a static flag before this stops being a no-op.
 		$this->logger->error( $message );
 
 		if ( $this->should_throw() ) {
--- a/includes/resources/Design_Tokens/Registry/Css_Var.php
+++ b/includes/resources/Design_Tokens/Registry/Css_Var.php
@@ -1,0 +1,35 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Registry;
+
+/**
+ * Derives the canonical CSS custom-property name from a token id.
+ *
+ * The id is a DTCG dot-path (e.g. "semantic.color.button-bg"). The variable name is never
+ * declared — it is produced by a single deterministic rule so ids and var names cannot drift:
+ *
+ *   prefix "--kb-token--", then replace each "." with "--".
+ *
+ *   semantic.color.button-bg  →  --kb-token--semantic--color--button-bg
+ *
+ * The --kb-token-- prefix keeps the variable inside KB's --global-kb-* family.
+ *
+ * @since TBD
+ */
+final class Css_Var {
+
+	public const PREFIX = '--kb-token--';
+
+	/**
+	 * Derive the canonical CSS custom-property name from a token id.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $id The DTCG dot-path id.
+	 *
+	 * @return string The derived CSS custom-property name.
+	 */
+	public static function from_id( string $id ): string {
+		return self::PREFIX . str_replace( '.', '--', $id );
+	}
+}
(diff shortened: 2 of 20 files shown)


tests/wpunit/Resources/Design_Tokens/Registry/RegistrationHelperTest.php
```diff
@@ -0,0 +1,68 @@
...
+use Tests\Support\Classes\TestCase;
+
+final class RegistrationHelperTest extends TestCase {
+
+	private Token_Registry $registry;
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->registry = $this->container->get( Token_Registry::class );
```
dave-green-uk, line 15: `Token_Registry` is a container singleton and `TestCase` never resets it, so the tokens registered in these tests (`semantic.color.test-only`, `kadence/testblock`) leak into every later test. It passes today only because the `Always_Present` stub no-ops the guard — but once SOFT-3377 binds a real baseline, a leaked token with no baseline entry trips the guard, giving an order-dependent failure. Suggest resetting the registry in `tearDown()`, or using `new Token_Registry()` where the container isn't required.

includes/resources/Design_Tokens/Registry/Baseline_Guard.php
```diff
@@ -0,0 +1,136 @@
...
+			return;
+		}
+
+		$message = sprintf(
+			'Design tokens disabled: %d declared token(s) have no baseline entry: %s',
+			count( $missing ),
+			implode( ', ', $missing )
+		);
+
+		$this->logger->error( $message );
```
dave-green-uk, line 81: This runs on `init`, so it fires on every front-end/admin/REST/AJAX/cron request. With the real baseline + any missing token, that's an `error_log()` write on *every* request. Could we log once (static flag or short transient) then degrade quietly? Otherwise a misconfigured prod site floods its log.

includes/resources/Design_Tokens/Registry/Token_Definition.php
```diff
@@ -0,0 +1,110 @@
...
+		$id = $definition['id'];
+
+		return new self(
+			$id,
+			$definition['type'],
+			$definition['label'],
+			$definition['group'] ?? '',
+			// css_var override is rare; default is derived and impossible to drift from the id.
+			isset( $definition['css_var'] ) ? (string) $definition['css_var'] : Css_Var::from_id( $id ),
+			$definition['projections'] ?? []
```
dave-green-uk, line 117: Only `id`/`type`/`label` are validated; `group`/`projections`/`css_var` fall through to a raw `TypeError`/`Error`, which breaks the `@throws InvalidArgumentException` contract for what's a public helper. A couple of `is_string`/`is_array` guards here would make the failure mode match the docblock.

includes/resources/Design_Tokens/Registry/Css_Var.php
```diff
@@ -0,0 +1,35 @@
...
+	 * Derive the canonical CSS custom-property name from a token id.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $id The DTCG dot-path id.
+	 *
+	 * @return string The derived CSS custom-property name.
+	 */
+	public static function from_id( string $id ): string {
+		return self::PREFIX . str_replace( '.', '--', $id );
```
dave-green-uk, line 33: `str_replace('.', '--', $id)` has no charset check — an id containing a space or slash yields an invalid CSS custom-property name. Worth an id-format guard in `Token_Definition::from_array()` since the registration helper is third-party-facing.

includes/resources/Design_Tokens/Registry/Provider.php
```diff
@@ -0,0 +1,86 @@
...
+		$this->load_helper();
+
+		// Declarations carry __() labels/groups, so they must not be evaluated before init or WordPress
+		// fires the _load_textdomain_just_in_time notice. register() runs on plugins_loaded, so defer the
+		// declaration load (and the guard that runs once tokens are in) to init. Early priority keeps the
+		// registry populated before any consumer (admin UI, projectors, resolver) reads it.
+		//
+		// The guard validates the registered tokens, so it must run after register_declarations. The
+		// later priority makes that ordering explicit rather than relying on same-priority insertion order.
+		add_action( 'init', [ $this, 'register_declarations' ], 0 );
```
dave-green-uk, line 43: The guard is a one-shot at `init:1`, but the helper docblock invites third-party callers "on their own hooks". Anything registered at `init:2`+ skips baseline validation entirely and `is_active()` is never re-checked. Fine if v1 is central-declarations-only — could we add a doc note ("register before `init:1`") to make that explicit?

includes/resources/Design_Tokens/Registry/Variant_Set.php
```diff
@@ -0,0 +1,48 @@
...
+	 * @since TBD
+	 *
+	 * @param array<string, mixed> $set The variant-set declaration.
+	 *
+	 * @throws InvalidArgumentException When "block" is missing.
+	 *
+	 * @return self
+	 */
+	public static function from_array( array $set ): self {
+		if ( empty( $set['block'] ) || ! is_string( $set['block'] ) ) {
```
dave-green-uk: `empty( $set['block'] )` rejects a `"0"` block name, which contradicts the `Token_Definition` tightening in this same PR (692d2e0d8) that deliberately avoids `empty()` to preserve `"0"`. Low-stakes, but worth matching: `! isset() || ! is_string() || === ''`.

dave-green-uk (changes requested):
Really clean module and great test coverage — no correctness bugs on the happy path. Comments are all input-robustness / test-isolation / forward-looking. The only one I'd treat as pre-merge is the singleton registry leak in the helper tests (cheap fix, prevents an order-dependent flake later); the rest are fair to defer.

tests/wpunit/Resources/Design_Tokens/Registry/RegistrationHelperTest.php
pauloiankoski: Fixed: rebind a fresh registry repopulated from declarations in tearDown. 49a5a73df

includes/resources/Design_Tokens/Registry/Baseline_Guard.php
pauloiankoski: Added a TODO (SOFT-3377) to throttle to once-per-process before it goes live. c9c84313b08d416a88c21bd0dea136ff98e43a59

includes/resources/Design_Tokens/Registry/Token_Definition.php
pauloiankoski: Fixed — is_string/is_array guards so it raises InvalidArgumentException, not a raw TypeError. Added coverage. d79b1fba7

includes/resources/Design_Tokens/Registry/Css_Var.php
pauloiankoski: Fixed at declaration time (keeps Css_Var a pure rule): `^[a-z0-9]+([.-][a-z0-9]+)*$`. Flag if uppercase ids should be allowed per the design doc. 1345a2eb7

includes/resources/Design_Tokens/Registry/Provider.php
pauloiankoski: Doc note added to the helper + Provider: register before init:1 to be covered. Re-validating late registrations is  SOFT-3377 scope. https://github.com/stellarwp/kadence-blocks/commit/c9c84313b08d416a88c21bd0dea136ff98e43a59

includes/resources/Design_Tokens/Registry/Variant_Set.php
pauloiankoski: Fixed — now matches the Token_Definition guard and preserves \"0\". 6bb349bdd

includes/resources/Design_Tokens/Registry/Baseline_Guard.php
```diff
@@ -0,0 +1,140 @@
...
+	private Token_Registry $registry;
+	private Baseline_Document $baseline;
+	private LoggerInterface $logger;
+
+	/**
+	 * Whether a missing entry throws. Resolves to WP_DEBUG when null (the production/dev default).
+	 *
+	 * @var bool|null
+	 */
+	private ?bool $throw_on_missing;
```
d4mation, line 31:
```suggestion
	/**
	 * Whether a missing entry throws. Resolves to WP_DEBUG when null (the production/dev default).
	 *
	 * @since TBD
	 *
	 * @var bool|null
	 */
	private ?bool $throw_on_missing;
```

includes/resources/Design_Tokens/functions.php
```diff
@@ -0,0 +1,39 @@
...
+	 * Token_Registry::register() that mirrors KB's existing global helpers. The declaration accepts
+	 * 'id' (DTCG dot-path), 'type' (DTCG $type), 'label', optional 'group', optional 'projections'
+	 * and an optional 'css_var' override. See Token_Definition::from_array().
+	 *
+	 * Register before `init` priority 1: the baseline guard runs once there and does not re-validate
+	 * tokens declared later, so a token registered on a later hook skips baseline validation entirely.
+	 *
+	 * @since TBD
+	 *
+	 * @param array<string, mixed> $definition The token declaration.
```
d4mation, line 17: We should have a clearer type definition here. It'll help a lot when phpstan support is merged and it'll make it clearer for anyone calling this what the expected shape is.

includes/resources/Design_Tokens/functions.php
```diff
@@ -0,0 +1,39 @@
...
+	}
+}
+
+if ( ! function_exists( 'kadence_blocks_register_design_variant_set' ) ) {
+	/**
+	 * Declare which block accepts variants (skeleton; data model in SOFT-3393).
+	 *
+	 * @since TBD
+	 *
+	 * @param array<string, mixed> $set See Variant_Set::from_array().
```
d4mation, line 32: Ideally we should define this better too if we can.

dave-green-uk (approved):
LGTM

includes/resources/Design_Tokens/Registry/Token_Definition.php
```diff
@@ -0,0 +1,133 @@
...
+ * Immutable description of a single design token — structure only, no value.
+ *
+ * Built from a declaration array via from_array(); the css-var is derived from the id by default
+ * (Css_Var) and only overridden when a token must match a pre-existing variable name.
+ *
+ * @since TBD
+ */
+final class Token_Definition {
+
+	/** @var string DTCG dot-path, e.g. "semantic.color.button-bg". */
```
d4mation, line 17: This will pass phpstan, but we should expand this so that we can utilize `@since`

includes/resources/Design_Tokens/Registry/Variant_Set.php
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Immutable record that a block accepts variants. Skeleton for SOFT-3379: holds the block name and
+ * the declared variant names. The $extensions…variants.<block> payload shape, $default handling and
+ * validation are SOFT-3393.
+ *
+ * @since TBD
+ */
+final class Variant_Set {
+
+	public string $block;
```
d4mation, line 16: We'll need a vardoc here as well.

includes/resources/Design_Tokens/Registry/Baseline_Guard.php
```diff
@@ -0,0 +1,140 @@
...
+	 * @since TBD
+	 *
+	 * @throws Missing_Baseline_Entry When WP_DEBUG is on and any declared token lacks a baseline entry.
+	 *
+	 * @return void
+	 */
+	public function run(): void {
+		$missing = $this->registry->missing_from_baseline( $this->baseline );
+
+		if ( $missing === [] ) {
```
d4mation, line 67: While accurate, this comparison feels odd to look at. Why not `empty( $missing )`?

includes/resources/Design_Tokens/Registry/Baseline_Guard.php
```diff
@@ -0,0 +1,140 @@
...
+	 * Run the guard: log, then either throw (debug) or fail closed (production) on missing entries.
+	 *
+	 * @since TBD
+	 *
+	 * @throws Missing_Baseline_Entry When WP_DEBUG is on and any declared token lacks a baseline entry.
+	 *
+	 * @return void
+	 */
+	public function run(): void {
+		$missing = $this->registry->missing_from_baseline( $this->baseline );
```
d4mation, line 65: I wonder if performance could be a problem here in the future. Maybe we need a TODO to come back to this. I wonder if we should only check against this once each time a Token is added/updated/removed and then cache the result. Otherwise we're looking at a O(n) problem here every since time this runs.

PR #1007, opened 2026-06-02
[DTM] Added baseline DTCG document and Json_Baseline_Document loader

dave-green-uk:
🎫  SOFT-3377

## Summary

Ships the read-only baseline DTCG document the whole module renders from, and binds it as the real Baseline_Document — replacing the permissive stub from SOFT-3379. The baseline gives every token a value, so a site that's never been authored still renders on-brand, and the activation guard now validates declared tokens against real entries (fail-closed if any are missing). Per §4a of the eng doc.

Scope is deliberately baseline + loader only; registering the full token catalog stays with the per-block wiring tickets (SOFT-3401–3406) and the variant work (SOFT-3393).

## Changes

- baseline.json — the shipped document: primitive + semantic token layers, foundation presets (4 type scales + 3 starter palettes), a $default block preset for each of the 6 Alpha blocks, and the Button primary/secondary/ghost variant set. Block set follows Donaldo's updated Alpha-6 (Icon, not Info Box).
- Json_Baseline_Document — file-backed loader implementing the contract. document() returns the full decoded baseline (for the Resolver); has($id) answers the guard. Loaded once per request, cached in the object cache keyed on the plugin version; failed loads aren't cached so a transient miss self-heals.
- Baseline_Document contract — extended with document() alongside has(); ripple applied to the two stubs + a test fake.
- Registry\Provider — binds Json_Baseline_Document in place of the stub.
- Docs — corrected the declarations.php breadcrumb (the full catalog lands via the wiring/variant tickets, not here).

## Test plan

- [ ] Unit (Json_Baseline_DocumentTest, 12 cases): has() true for registered tokens, primitives, and composite (shadow/typography) leaves; false for group nodes, unknown paths, and the $extensions layer; document() returns the full decoded baseline; document is cached across instances on a version; a failed load isn't cached and recovers; missing/malformed files fail closed.
- [ ] Invariant: a test loads the real declarations.php and asserts every registered token has a baseline entry — adding a token without a baseline leaf turns the suite red.
- [ ] Integration: the guard test resolves the real binding through the container and confirms projection stays active.
- [ ] Full module suite: 53 tests / 102 assertions green; phpcs clean on all authored files.
- [ ] Manual: N/A — no runtime UI/behaviour yet; consumed by later projector/resolver tickets.


includes/resources/Design_Tokens/Registry/Baseline/baseline.json
```diff
@@ -0,0 +1,457 @@
+{
+	"$description": "Shipped, read-only baseline DTCG document for the Kadence Blocks Design Tokens module. Defines a value for every token, a curated pool of foundation presets, a $default block preset for each alpha block, and the Button lead variant set. The stored document (kb_design_tokens table) is sparse overrides layered on top of this baseline; this file is versioned with the plugin and is NEVER mutated at runtime. See Engineering Design Document section 4a.",
```
d4mation: I don't know if we want to reference the EngDoc here. It could easily be forgotten to be removed later 😅 

In PHP code we can easily leave TODO comments, etc., but not in JSON.

d4mation (approved):
Minor thing, but the code looks good!

includes/resources/Design_Tokens/Registry/Baseline/baseline.json
pauloiankoski: Resolved on https://github.com/stellarwp/kadence-blocks/pull/1007/commits/13c70706e4fbdb8079acc57b6c4af691f717d712

dave-green-uk: @d4mation Dang, I missed this in the self-review. thanks for the commit @pauloiankoski !

d4mation:
@dave-green-uk Hold off on merging this for a few, please :) Working on fixing CI issues in the base branch

d4mation:
@dave-green-uk The PR to fix the existing CI issues is here and is ready for review: https://github.com/stellarwp/kadence-blocks/pull/1010

Once it is merged, this branch should be synced with its base branch again so that it can flag any possible CI issues.

dave-green-uk:
@d4mation still getting some issues post-sync

d4mation:
> @d4mation still getting some issues post-sync

@dave-green-uk it looks like the names of different blocks should probably be added to the global ignore list to avoid them getting flagged. But it could also be worth considering ignoring the baseline file in its entirety. 

The Tests CI issues should be unrelated, but I plan to look more closely at them if they're still happening tomorrow.

dave-green-uk:
Merging as the cspell update has resolved that specific failure.

PR #1008, opened 2026-06-02
[DTM] Address review comments on Token Registry (#1006)

pauloiankoski:
Follow-up to #1006 addressing @d4mation's review comments. Each fix is its own commit.

## Changes

- **Baseline_Guard** — add `@since` to the `$throw_on_missing` vardoc.
- **functions.php** — give both registration helpers explicit array-shape param types (`array{id: string, ...}`) so the expected declaration shape is clear and PHPStan-friendly.
- **Token_Definition** — expand the single-line property vardocs to full blocks so each carries `@since`.
- **Variant_Set** — add property vardocs to `$block` / `$variants`.
- **Baseline_Guard** — early-return on `empty( $missing )` instead of `=== []` (array, no `"0"` concern).
- **Baseline_Guard** — add a `TODO (SOFT-3377)` to cache the O(n) `missing_from_baseline()` check rather than recomputing every run.

### DIFF
--- a/includes/resources/Design_Tokens/Registry/Baseline_Guard.php
+++ b/includes/resources/Design_Tokens/Registry/Baseline_Guard.php
@@ -66,7 +66,7 @@ final class Baseline_Guard {
 	public function run(): void {
 		$missing = $this->registry->missing_from_baseline( $this->baseline );
 
-		if ( $missing === [] ) {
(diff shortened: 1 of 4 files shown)


d4mation (approved):
Thanks!

PR #1009, opened 2026-06-02
DTCG JSON Schema + validation rules

d4mation:
🎫  https://linear.app/nexcess/issue/SOFT-3378/dtcg-json-schema-validation-rules

Adds the DTCG JSON Schema + validation rules for the Design Tokens module. A `Dtcg_Validator` walks a decoded DTCG document and collects every error (with dot-path locations and stable `get_code_*()` codes) rather than throwing, so the REST writes can return them all at once. It enforces:

- the `$type` enum (`color`, `dimension`, `fontFamily`, `shadow`, `typography`);
- the whole-string alias `{dot.path}` pattern, accepted anywhere a literal is expected — including composite sub-fields;
- per-type value grammar and composite shapes (shadow/typography walked field by field);
- leaf shape — a token leaf only carries `$`-prefixed keys, matching the published schema's `additionalProperties: false`;
- two contexts — `baseline` (full document, sentinels rejected) and `overrides` (sparse store writes that are sparse in PATHS, not keys: every concrete leaf is still self-describing with `$type + $value`, and the only sparse leaf forms are the `$value:null` / `$disabled` sentinels).

Also ships a published draft-07 `dtcg.schema.json` generated from the same `Token_Type`/`Alias` source of truth (so it can't drift) and held in lockstep by a conformance test that exercises both engines on the same structural-invalid fixtures. The code is organized into a `Vocabulary\` layer (the grammar primitives the Resolver will also read) and a `Validation\` layer (the engine, result model, and per-type validators); the schema generator lives as dev-only tooling behind `composer generate-dtcg-schema`. Alias-cycle detection (Resolver dry run) and CSS rendering are intentionally out of scope. The conformance test validates against the real shipped baseline.

> [!NOTE]
> **Most of the diff is generated output, reference data, or boilerplate — not novel logic.** Of the ~3,950 added lines, roughly half is low-review-effort:
> - the generated `dtcg.schema.json` (~327 lines — emitted by the generator, don't hand-review it);
> - curated reference data in `Literals` (the ~150-entry CSS named-color allowlist plus the color/font-weight keyword and length-unit lists, ~165 lines) — lookup tables, not logic;
> - accessor + const boilerplate — 22 `public static function get_*()` accessors wrapping private consts (repo convention: a public const can't be deprecated cleanly), each accessor and const carrying its own `@since`/`@var` docblock, repeated across the vocabulary and error-code classes;
> - the ~1,250-line test suite is largely declarative — named-yield `Generator` data providers and the per-method `@param` / `@return void` docblocks the repo standard requires in `tests/`.
>
> The genuinely new validation logic — the document walker, the per-type value checks, the leaf-shape enforcement, and the alias/composite rules — is the smaller remainder.

### Checklist


composer.json
```diff
@@ -136,7 +140,7 @@
 		],
 		"phpcs": "phpcs --standard=phpcs.xml",
 		"phpcs-autofix": "phpcbf --standard=phpcs.xml",
-		"phpcs-ci": "dev_scripts/partial-phpcs.sh diff",
+		"phpcs-ci": "dev_scripts/bin/partial-phpcs.sh diff",
```
d4mation, line 143: Moved to `dev_scripts/bin` to match the new structure added by the Schema Generator which has no reason to be shipped in production code.

dev_scripts/bin/generate-dtcg-schema.php
d4mation, line 1: Separated from `dev_scripts/src/Design_Tokens/Schema/Dtcg_Schema_Generator.php` intentionally as tests call `dev_scripts/src/Design_Tokens/Schema/Dtcg_Schema_Generator.php` to check that it has been generated properly at the expected location: `includes/resources/Design_Tokens/Schema/dtcg.schema.json`.

includes/resources/Design_Tokens/Schema/Vocabulary/Literals.php
```diff
@@ -0,0 +1,347 @@
...
+
+	/**
+	 * The CSS named colors (CSS Color Module Level 4 extended set, lower-cased). A curated allowlist is
+	 * what lets "not-a-color" be rejected while "rebeccapurple" is accepted.
+	 *
+	 * @since TBD
+	 *
+	 * @var string[]
+	 */
+	private const NAMED_COLORS = [
```
d4mation, line 65: I _hate_ this, but we should validate these and an allowlist is the only way to not just match random text as if it were a known color 🫠

includes/resources/Design_Tokens/Schema/Validation/Values/Composite_Value.php
```diff
@@ -0,0 +1,101 @@
...
+
+		if ( ! is_array( $value ) ) {
+			return [
+				new Validation_Error(
+					$path,
+					Validation_Error::get_code_malformed_node(),
+					sprintf( 'A %s value must be an object of sub-fields or an alias.', $this->type() )
+				),
+			];
+		}
```
pauloiankoski, line 66: When a composite `$value` is a bare scalar (for example, someone passes `5` for a shadow value), we return `CODE_MALFORMED_NODE`. But for the same class of problem on a scalar type, such as an integer passed as a color, `Kind::validate()` falls through to `Literals` and returns `CODE_VALUE_INVALID`.

Since these error codes are part of a cross-component contract (REST clients and MCP clients branch on them), a REST consumer would currently need to handle these two cases differently even though both effectively mean “wrong shape for `$value`”. `CODE_VALUE_INVALID` also feels more descriptive here because a non-object composite is an invalid value, not a malformed document node.

One-line fix: swap `get_code_malformed_node()` for `get_code_value_invalid()` in that branch and update `testShadowRejectsNonObjectLiteral` accordingly.

includes/resources/Design_Tokens/Schema/Validation/Dtcg_Validator.php
```diff
@@ -0,0 +1,448 @@
...
+		if ( $context === self::CONTEXT_BASELINE ) {
+			if ( $has_disabled || $is_reset ) {
+				$errors[] = new Validation_Error(
+					$path,
+					Validation_Error::get_code_sentinel_not_allowed(),
+					'Override sentinels ("$value": null, "$disabled") are not allowed in the baseline document.'
+				);
+
+				return $errors;
+			}
```
pauloiankoski, line 239: When a sentinel appears in the baseline context, the error path points at the leaf (`primitive.x`) instead of the specific key (`primitive.x.$value` or `primitive.x.$disabled`). Every other error in this PR (`sentinel_invalid`, `type_unknown`, `leaf_field_unknown`, and the composite field errors) points at the exact key that's wrong.

This one is the odd one out, and since the path is what a REST client uses to highlight the correct field in the editor, it seems worth making it precise.

The fix is just threading the right suffix through:

```php
// $is_reset:
$errors[] = new Validation_Error(
    $path . '.' . Sentinels::get_value_key(),
    Validation_Error::get_code_sentinel_not_allowed(),
    ...
);

// $has_disabled:
$errors[] = new Validation_Error(
    $path . '.' . Sentinels::get_disabled_key(),
    Validation_Error::get_code_sentinel_not_allowed(),
    ...
);
```

And update the “reset sentinel baseline” provider case so the expected `path` changes from `primitive.x` to `primitive.x.$value`.

includes/resources/Design_Tokens/Schema/Vocabulary/Literals.php
```diff
@@ -0,0 +1,375 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string $value The candidate function string.
+	 *
+	 * @return bool
+	 */
+	private static function is_function( string $value ): bool {
+		return (bool) preg_match( '/^[a-z][a-z0-9-]*\(.*\)$/is', $value );
+	}
```
pauloiankoski, line 374: Tiny one: the `is_function` regex uses the `s` (`DOTALL`) modifier, which makes `.` match newlines. That means a string like `rgb(\n10,20,30)` would currently pass as a valid CSS function.

CSS values decoded from JSON probably will not contain embedded newlines in practice, but since the docblock already carefully explains the greedy `.*` trade-off, it seems better not to silently widen the match further.

Dropping the `s` flag from the regex is a one-character fix:

```php id="g8t2qp"
/^[a-z][a-z0-9-]*\(.*\)$/i
```

Non-blocking tho.

tests/wpunit/Resources/Design_Tokens/Schema/Validation/Dtcg_ValidatorTest.php
```diff
@@ -0,0 +1,278 @@
...
+						'$type'  => 'color',
+						'$value' => '#fff',
+						'extra'  => 'oops',
+					],
+				],
+			],
+			'context'  => Dtcg_Validator::get_context_baseline(),
+			'code'     => Validation_Error::get_code_leaf_field_unknown(),
+			'path'     => 'primitive.x.extra',
+		];
```
pauloiankoski, line 177: Minor test coverage gap: the “unknown leaf field” provider case currently asserts `assertCount(1, $errors)`, which locks in single-error behaviour but does not verify that multiple unknown fields on the same leaf produce independent errors.

`testItCollectsEveryErrorInOnePass()` already covers the multi-leaf case nicely, so it would be great to have a parallel case for multiple unknown fields on a single leaf. Something like:

```php id="y9k4dn"
[
    '$type'  => 'color',
    '$value' => '#fff',
    'foo'    => 'bar',
    'baz'    => 'qux',
]
```

with an assertion like:

```php id="a2w8rp"
assertCount( 2, $errors );
```

Totally non-blocking, just a small gap worth filling.

pauloiankoski (approved):

PR #1011, opened 2026-06-03
[DTM] Record previous document on token save in kb_design_tokens_history

dave-green-uk:
🎫  SOFT-3376

Note: the CI errors are unrelated to these changes, but the fix is up for review here: https://github.com/stellarwp/kadence-blocks/pull/1012

## Summary

Adds the kb_design_tokens_history table so the Design Tokens module owns its own undo/audit trail: on every successful token-set save, the previous DTCG document is archived. This gives controlled undo/audit of UI- and MCP-authored changes without relying on WordPress revisions — which several managed hosts (e.g. WP Engine) disable.

Stacked on feature/design-tokens-module (builds directly on the Token_Store/Token_Table from SOFT-3375).

###  New persistence layer
- Token_History_Table — the kb_design_tokens_history table (id, slug, document, version, created_at, KEY slug). Non-destructive migration; registered in Database_Provider's Register::tables([...]) list.
- Token_History_Store — the sole gateway to the table: record(), latest(), for_set(), count(). Append-only; each archive prunes the set's trail to the most-recent N snapshots (default 5).

### Capture hook in Token_Store
- save_document() reads the existing row before the upsert, then — only after the write commits — fires a new kadence_blocks_design_tokens_superseded action carrying the now-previous document and version. A failed write throws first, so nothing is archived for a save that didn't happen, and first saves (no prior row) archive nothing.
- Design_Tokens\Database\Provider binds the history store and subscribes it to that action via the container callback() idiom.

## Scope notes
- Minimal schema — no source/author column yet; that lands with the REST/MCP layer that knows the origin.
- No dedup of identical saves, mirroring Token_Store's existing "persist, don't judge" behaviour (it re-versions on every save).
- Retention — ships pruning to the most-recent 5 snapshots per set, pruned on each archive. Two independent per-set filters: `kadence_blocks_design_tokens_history_limit` sets the cap (clamped to a minimum of 1 — it's a cap, not an off switch), and `kadence_blocks_design_tokens_history_enabled` (bool, default true) toggles recording entirely. Pruning deletes surplus rows only — no schema migration.

## Test plan

- [ ] Save a token set twice in the admin → confirm one history row holding the first document; save again → two rows, newest first.
- [ ] On a fresh site (no row), first save creates no history row.
- [ ] Save a set more than 5 times → confirm only the newest 5 snapshots remain.
- [ ] Filter `kadence_blocks_design_tokens_history_enabled` to false → confirm new saves stop archiving while any existing snapshots remain in place.


includes/resources/Design_Tokens/Database/Token_Store.php
```diff
@@ -214,4 +253,29 @@ private function changed( string $slug ): void {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string $slug     The token set slug that was overwritten.
+	 * @param string $document The now-previous document that was replaced.
+	 * @param string $version  The now-previous version hash.
+	 *
+	 * @return void
+	 */
+	private function saving( string $slug, string $document, string $version ): void {
```
d4mation: If it has already been overwritten, do we want to call it _saving_?

I would expect it to happen concurrently with the save with that name. Perhaps `saved` would be better?

But since we have `changed`, maybe that is confusing? Maybe `changed` should just fire this action?

d4mation (commented):
I wonder: Maybe we should provide a way to limit the number of previous tokens to save, and then we delete the oldest once that limit is hit? It could be a nice feature to help avoid this table getting really large.

That can be a follow-up though. Otherwise, the PR looks pretty good :)

includes/resources/Design_Tokens/Database/Token_Store.php
dave-green-uk: Good catch — renamed it to superseded (superseded_action() / kadence_blocks_design_tokens_superseded), since it fires after the write and carries the prior document being replaced, so the old name was misleading on both tense and direction. Kept it separate from changed on purpose: changed fires on every write (including first saves and version bumps) with just the slug for cache invalidation, whereas this one fires only on an overwrite and carries the previous document/version for the history store to archive.

Changed in https://github.com/stellarwp/kadence-blocks/pull/1011/commits/ae223c0ea8836d3638733959516a63a1fcb1ae3c

dave-green-uk:
@d4mation pruning/retention change is in https://github.com/stellarwp/kadence-blocks/pull/1011/commits/1e2da223a015c2ec584bd422eaad44dc2e05b281

I went with `5` as after looking around, WPE and a few other managed hosts suggest that as a sensible value. WP core doesn't actually set a default!

includes/resources/Design_Tokens/Database/Token_History_Store.php
```diff
@@ -0,0 +1,219 @@
...
+	 *
+	 * @param string $slug The token set slug the limit applies to.
+	 *
+	 * @return int
+	 */
+	private function history_limit( string $slug ): int {
+		/**
+		 * Filters how many design-token history snapshots are kept per token set.
+		 *
+		 * Return 0 or a negative number to disable pruning and keep the full trail.
```
d4mation: I don't know if I like the idea of `0` or a negative number _both_ disabling pruning. In my head, `0` should disable history altogether. But with the name of the filter, `history_limit === 0` also could be confusing, as it _could_ imply disabling pruning 🤔 

Maybe we need a separate filter elsewhere to disable history altogether, and then this filter can exist as it does now.

includes/resources/Design_Tokens/Database/Token_History_Store.php
```diff
@@ -0,0 +1,219 @@
...
+	 * The shipped per-set history retention limit, before any filtering.
+	 *
+	 * @since TBD
+	 *
+	 * @return int
+	 */
+	public static function default_history_limit(): int {
+		return self::DEFAULT_HISTORY_LIMIT;
+	}
+
```
d4mation: Does this need to be public? Is there a reason a 3rd party/external consumer would need to know the default limit? I figure the only thing they would need access to is the filtered limit.

d4mation: If we don't need to expose it, we could remove this method altogether.

includes/resources/Design_Tokens/Database/Token_History_Store.php
dave-green-uk: Went with the separate filter. I didn't feel strongly about it either way.

includes/resources/Design_Tokens/Database/Token_History_Store.php
dave-green-uk: Removed!

d4mation (approved):

PR #1013, opened 2026-06-03
[DTM] Token Resolver — effective-document merge, alias flattening, CSS renderers

pauloiankoski:
## Summary

The Token Resolver — the single place where DTCG aliases flatten. It takes the raw overrides-only document from `Token_Store`, deep-merges it onto the shipped baseline, walks the merged ("effective") document depth-first, collapses alias chains to literals, and renders each `$type` to a CSS string. Output is two flat, alias-free maps that every downstream projector consumes directly:

- `token-id → resolved value`
- `css-var → value`

This keeps alias interpretation in exactly one component so projectors never re-implement it. It is the convergence point of the foundation spine — the downstream projector and endpoint work all build on these maps.

## What's included

- **`Alias`** — DTCG reference micro-syntax (`{dot.path}`): `is_alias()` / `path_of()`.
- **`Css_Renderer`** — per-`$type` rendering: `fontFamily` lists join (quoting multi-word families), `shadow` → `offsetX offsetY blur spread color`, `typography` → `font` shorthand, `color`/`dimension` pass through.
- **`Effective_Document`** — deep-merges stored overrides onto the baseline (override leaves win); `$value: null` and `"$disabled": true` sentinels remove a baseline entry. Only `primitive`/`semantic` layers merge; `$extensions` is excluded.
- **`Resolved_Tokens`** — immutable value object holding the two maps.
- **`Token_Resolver`** — depth-first walk + cycle-safe alias resolution (composite sub-fields resolve field-by-field). `resolve()` memoises per request on the store version (bumped on every write, so the memo self-invalidates); `dry_run()` lets the write layer reject aliasing cycles before committing.
- **`Resolver\Provider`** — binds the above as singletons; registered from `Design_Tokens_Provider`.

## Caching

Request-scoped memoisation only, keyed on the store version. Cross-request object-cache persistence is deliberately deferred to its follow-up — the version-keyed design makes that purely additive, no change to resolution logic.

## Tests

45 tests (56 assertions) passing under wpunit: unit coverage for `Alias`, `Css_Renderer`, `Resolved_Tokens`, and `Effective_Document`, plus `Token_Resolver` dry-run resolution (alias chains, composite walk, dangling references, cycle detection) and an integration test against the real `baseline.json` proving store-backed resolution and memo invalidation on a version bump.

The integration test surfaced and fixed a rendering bug: a typography token whose `fontFamily` field resolves to a list was being stringified to `"Array"`; it now renders through the family-list path.

## Notes for review

- Dangling alias (reference to a missing path) passes the `{…}` string through unchanged rather than emitting empty, so the failure stays visible.
- Composite emission for `typography` (single shorthand vs. split longhand custom properties) is left as a follow-up decision for the CSS-var projector; the renderer is the only place that would change.


includes/resources/Design_Tokens/Resolver/Alias.php
```diff
@@ -0,0 +1,36 @@
...
+ * The DTCG alias micro-syntax: a value that is a single reference, e.g. "{primitive.color.brand.primary}".
+ *
+ * @since TBD
+ */
+final class Alias {
+
+	/**
+	 * A $value is an alias only when the whole string is a single {dot.path} reference.
+	 */
+	private const PATTERN = '/^\{([\w.-]+)\}$/';
```
d4mation: All variable docs need `@since TBD` and in the case of `const` where we can't add a native type, it should have `@var <type>` as well.

includes/resources/Design_Tokens/Resolver/Alias.php
```diff
@@ -0,0 +1,36 @@
...
+
+	/**
+	 * @param mixed $value
+	 */
+	public static function is_alias( $value ): bool {
```
d4mation: All methods need `@since TBD` and they should also have `@return <type>` even when a native return type is defined.

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
```diff
@@ -0,0 +1,100 @@
...
+			case 'fontFamily':
+				return $this->font_family( $value );
+			case 'shadow':
+				return $this->shadow( $value );
+			case 'typography':
+				return $this->typography( $value );
+			case 'color':
+			case 'dimension':
+			default:
+				return is_scalar( $value ) ? (string) $value : '';
```
d4mation: I wonder if we should just bring over the `Cast` class from Harbor 🤔 

https://github.com/stellarwp/harbor/blob/main/src/Harbor/Utils/Cast.php

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
```diff
@@ -0,0 +1,100 @@
...
+
+	/**
+	 * Render a shadow composite to "<offsetX> <offsetY> <blur> <spread> <color>".
+	 *
+	 * v1 supports a single shadow object. DTCG also permits a $value that is an array of
+	 * shadow objects (stacked box-shadows); that shape is intentionally not handled here —
+	 * Token_Resolver passes a list through untouched, and supporting it is a follow-up that
+	 * would change this method (join each layer with ", ") and the resolver's list handling.
+	 *
+	 * @param mixed $value
```
d4mation: We should probably do `mixed|array{offsetX: string, offsetY: string, <rest of the shape>}` so that the expected shape is documented while still allowing in non-array values for if a Token is somehow provided incorrectly.

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
```diff
@@ -0,0 +1,100 @@
...
+			$value['color']   ?? ''
+		) );
+	}
+
+	/**
+	 * Render a typography composite to the CSS `font` shorthand
+	 * ("<weight> <size>/<line-height> <family>"). Emitted as one custom property;
+	 * projectors may also split it into longhand properties if needed.
+	 *
+	 * @param mixed $value
```
d4mation: Same here with defining the array shape

includes/resources/Design_Tokens/Resolver/Effective_Document.php
```diff
@@ -0,0 +1,113 @@
...
+			}
+
+			// A leaf ($value present) replaces wholesale — never field-merge a token value.
+			if ( is_array( $value ) && array_key_exists( '$value', $value ) ) {
+				$base[ $key ] = $value;
+				continue;
+			}
+
+			// Two groups: descend. Otherwise the override introduces a new branch.
+			if ( is_array( $value ) && isset( $base[ $key ] ) && is_array( $base[ $key ] ) ) {
```
d4mation: This can potentially corrupt a leaf. It needs to check if the baseline node is a token leaf, because if the baseline holds a leaf at a key and the override tries to put a group there the children get merged in as sublings of `$type` and `$value` resulting in a weird hybrid node and it'll drop the child tokens.

Adding `&& ! array_key_exists( '$value', $base[ $key ] )` to the condition should account for this.

includes/resources/Design_Tokens/Resolver/Alias.php
d4mation, line 1: This is effectively a duplicate of `Design_Tokens\Schema\Vocabulary\Alias`, right? Maybe we should reuse it instead.

includes/resources/Design_Tokens/Resolver/Effective_Document.php
```diff
@@ -0,0 +1,113 @@
...
+	private function is_removal( $value ): bool {
+		if ( ! is_array( $value ) ) {
+			return false;
+		}
+
+		if ( array_key_exists( '$value', $value ) && $value['$value'] === null ) {
+			return true;
+		}
+
+		return ( $value['$disabled'] ?? false ) === true;
```
d4mation: We should utilize `Sentinels::is_disabled()` here I think rather than hardcoding it.

tests/wpunit/Resources/Design_Tokens/Resolver/AliasTest.php
```diff
@@ -0,0 +1,48 @@
...
+	 * @param mixed $value
+	 */
+	public function testItRecognizesAliases( $value, bool $expected ): void {
+		$this->assertSame( $expected, Alias::is_alias( $value ) );
+	}
+
+	/**
+	 * @return array<string, array{0: mixed, 1: bool}>
+	 */
+	public function aliasProvider(): array {
```
d4mation: Not a _huge_ deal breaker, but I'd rather us use `Generator` data providers personally. Claude always seems to default to array-based ones.

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
```diff
@@ -0,0 +1,100 @@
...
+	 * shadow objects (stacked box-shadows); that shape is intentionally not handled here —
+	 * Token_Resolver passes a list through untouched, and supporting it is a follow-up that
+	 * would change this method (join each layer with ", ") and the resolver's list handling.
+	 *
+	 * @param mixed $value
+	 */
+	private function shadow( $value ): string {
+		if ( ! is_array( $value ) ) {
+			return '';
+		}
```
d4mation: The Schema validation _should_ prevent it from ever happening, but if the Array were nested somehow we could encounter an issue here. 

```php
if ( ! is_array( $value ) || $value === [] || array_keys( $value ) === range( 0, count( $value ) - 1 ) ) {
    return '';
}
```

However, it is technically valid DTCG to have `$value` be a list of this object (stacked box-shadows) which our code doesn't support at all right now.

Example: 

```json
"shadow": {
  "card-elevated": {
    "$type": "shadow",
    "$value": [
      {
        "color": "{primitive.color.neutral.900}",
        "offsetX": "0px",
        "offsetY": "1px",
        "blur": "2px",
        "spread": "0px"
      },
      {
        "color": "{primitive.color.neutral.700}",
        "offsetX": "0px",
        "offsetY": "4px",
        "blur": "12px",
        "spread": "-2px"
      }
    ]
  }
}
```

Maybe we will want to revisit it, but for now the guard may be best.

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
```diff
@@ -0,0 +1,100 @@
...
+ * @since TBD
+ */
+final class Css_Renderer {
+
+	/**
+	 * @param mixed $value Literal scalar, list, or composite array.
+	 */
+	public function render( string $type, $value ): string {
+		switch ( $type ) {
+			case 'fontFamily':
```
d4mation: We should use `Token_Type` to get these strings.

includes/resources/Design_Tokens/Resolver/Effective_Document.php
```diff
@@ -0,0 +1,113 @@
...
+ */
+final class Effective_Document {
+
+	/**
+	 * The DTCG layers that carry resolvable tokens. Mirrors Json_Baseline_Document
+	 * so $extensions (presets/variants) is never treated as a token tree.
+	 *
+	 * @var string[]
+	 */
+	private const TOKEN_LAYERS = [ 'primitive', 'semantic' ];
```
d4mation: We now have this exact same array in three locations: 

- `Json_Baseline_Document`
- `Effective_Document`
- `Token_Resolver`

Maybe we should put this somewhere centralized?

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
```diff
@@ -0,0 +1,243 @@
...
+	/**
+	 * Dry-run resolution against an ad-hoc overrides array — used by the REST write layer
+	 * to reject aliasing cycles before committing. Never touches the store/memo.
+	 *
+	 * @param array<string,mixed> $overrides Decoded candidate overrides.
+	 *
+	 * @throws Alias_Cycle_Exception    When the candidate introduces an unresolvable cycle.
+	 * @throws Dangling_Alias_Exception When the candidate aliases a path with no token leaf.
+	 */
+	public function dry_run( array $overrides ): Resolved_Tokens {
```
d4mation: This method seems really strange. 

The goal is so that we check the resolution before storing, but nothing actually _calls this_ outside of tests.

Maybe it should be renamed to `resolve_overrides()`? Not a huge blocker as I know we'll use it more later. But just trying to make sure the overall intent of it is clear.

includes/resources/Design_Tokens/Resolver/Alias.php
pauloiankoski: Resolved by reusing `Schema\Vocabulary\Alias`; this duplicate class was removed: e1c195f

includes/resources/Design_Tokens/Resolver/Alias.php
pauloiankoski: Resolved by reusing `Schema\Vocabulary\Alias`; this duplicate class was removed: e1c195f

includes/resources/Design_Tokens/Resolver/Alias.php
pauloiankoski: Agreed — reused it and removed the duplicate. Resolved: e1c195f

tests/wpunit/Resources/Design_Tokens/Resolver/AliasTest.php
pauloiankoski: Resolved: e1c195f — the duplicate `Alias` and this test were removed (reusing `Schema\Vocabulary\Alias`, which has its own coverage).

includes/resources/Design_Tokens/Resolver/Effective_Document.php
pauloiankoski: Resolved: 09fb1d2 — now uses `Sentinels::is_reset()` / `Sentinels::is_disabled()` instead of the hardcoded keys.

includes/resources/Design_Tokens/Resolver/Effective_Document.php
pauloiankoski: Good catch. Resolved: d5312d1 — added the `! array_key_exists( '$value', $base[ $key ] )` guard plus a regression test.

includes/resources/Design_Tokens/Resolver/Effective_Document.php
pauloiankoski: Resolved: bb5240a — extracted into a `Schema\Vocabulary\Layers` class now used by all three consumers.

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
pauloiankoski: Resolved: 8c7023e — dispatches on `Token_Type` getters now.

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
pauloiankoski: Resolved: 3da5655 — string coercion now goes through Harbor's `Cast::to_string()`.

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
pauloiankoski: Resolved: 946cbe9 — documented the array shape.

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
pauloiankoski: Resolved: 946cbe9 — documented the array shape.

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
pauloiankoski: Resolved: 9f575be — added the empty/list-shaped guard (stacked box-shadows remain a follow-up).

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
pauloiankoski: Resolved: 5e8b830 — renamed to `resolve_overrides()`.

dave-green-uk (changes requested):
is_removal() / Sentinels.php — RESET sentinel doesn't match its docs

Sentinels.php says $value: null is RESET → "fall back to the baseline value." But is_removal() treats it like DISABLE and unset()s the entry, so the token is dropped entirely. The two sentinels end up identical, and a "reset to default" loses the token instead of restoring its baseline value.

testANullValueSentinelRemovesTheBaselineEntry locks in the deletion, so the test and the docblock also disagree.

Which is intended — fix is_removal() to exclude is_reset() (keep baseline), or update the doc/comment to say $value: null removes? Either way the other needs to change too.

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
d4mation: That's not what I meant 😅 

Blocks should maintain its own copy of the class. Relying on Harbor's feels weird.

includes/resources/Design_Tokens/Resolver/Effective_Document.php
```diff
@@ -0,0 +1,111 @@
...
+			}
+
+			// Two groups: descend. A baseline leaf is never a group, so when the override puts a group
+			// where the baseline holds a leaf, the override replaces it (falls through below) rather than
+			// merging the group's children alongside the leaf's $type/$value and corrupting the node.
+			if ( is_array( $value )
+				&& isset( $base[ $key ] )
+				&& is_array( $base[ $key ] )
+				&& ! array_key_exists( Sentinels::get_value_key(), $base[ $key ] )
+			) {
```
d4mation: If we're going to split this across lines, we should put the first one on its own line too. 

```suggestion
			if (
				is_array( $value )
				&& isset( $base[ $key ] )
				&& is_array( $base[ $key ] )
				&& ! array_key_exists( Sentinels::get_value_key(), $base[ $key ] )
			) {
```

includes/resources/Design_Tokens/Resolver/Effective_Document.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @param mixed $value
+	 */
+	private function is_removal( $value ): bool {
+		if ( ! is_array( $value ) ) {
+			return false;
+		}
+
+		// Both override-suppression sentinels drop the baseline entry from the effective document; the
+		// spellings live on Sentinels so this never hard-codes "$value"/"$disabled".
+		return Sentinels::is_reset( $value ) || Sentinels::is_disabled( $value );
```
d4mation: This smells of Claude documenting the CR change based on the conversation rather than what makes sense to have in the code. We could shorten this a little.

```suggestion
		// Both override-suppression sentinels drop the baseline entry from the effective document.
		return Sentinels::is_reset( $value ) || Sentinels::is_disabled( $value );
```

includes/resources/Design_Tokens/Schema/Vocabulary/Layers.php
```diff
@@ -0,0 +1,59 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary;
+
+/**
+ * The top-level DTCG document layers, single-sourced so the baseline index, the deep-merge and the
+ * Resolver agree on exactly which layers carry resolvable tokens.
```
d4mation: This also reads like Claude trying to document the conversation based on the CR changes. We shouldn't reference the places this is used because it _could_ change. Clarifying that it is meant as a single source of truth is fine though.

includes/resources/Design_Tokens/Resolver/Effective_Document.php
d4mation: Related comment from @dave-green-uk: https://github.com/stellarwp/kadence-blocks/pull/1013#pullrequestreview-4426720411

Previously it only checked against `is_disabled()` but now it is checking against either `is_disabled()` or `is_reset()`. 

`is_reset()` should be resetting to the baseline, but `is_removal()` _removes_ the token entirely and it should probably only be checking `is_disabled()`.

We need to ensure that the `is_resest()` sentinel is used correctly to reset to baseline and that `is_disabled()` is used correctly to remove it entirely from the effective document.

includes/resources/Design_Tokens/Schema/Vocabulary/Layers.php
pauloiankoski: 9d22a5368bddc9c5db5caff5dd1cdef5827a1c6b

includes/resources/Design_Tokens/Resolver/Css_Renderer.php
pauloiankoski: 7832455b458c5e1a07de51934cb3bad653245c50

includes/resources/Design_Tokens/Resolver/Effective_Document.php
pauloiankoski: This is now fixed by moving the `is_array( $value )` check outside the if statement, allowing it to be used as an early return. f3e68c0dded0dfd64e6eaccb7b209879a86a911d

includes/resources/Design_Tokens/Resolver/Effective_Document.php
pauloiankoski: The `is_removal` method have been entirely removed as it's no longer being used after fixing the `is_reset` / `is_disabled` comment. f3e68c0dded0dfd64e6eaccb7b209879a86a911d

dave-green-uk (approved):
LGTM, nothing more to add!

d4mation (approved):

PR #1014, opened 2026-06-03
feat: add legacy license key validation function

gilbert-hernandez:
Introduces a new function, `kadence_blocks_is_legacy_key_valid`, to validate
  legacy license keys directly against the licensing server. This ensures
  compatibility with older keys while maintaining the original intent of key
  validation. The result is cached to optimize performance during editor and
  REST loads.

  Legacy keys (e.g. non LWSW…) are activated via direct license-key validation
  against the kadencewp.com licensing server and never obtain a V3 authorization
  token, so is_authorized() returns false for them even when the key is good.
  As a result, `kadence_blocks_get_authorized_license_key()` returned an empty
  string for these accounts, which flowed through
  `kadence_blocks_get_current_license_data()` into an empty api_key on the
  pattern/template/AI requests to startertemplatecloud.com — so those servers
  rejected the imports as unlicensed. This adds a legacy-validation fallback
  inside the candidate loop: if the V3 token check fails, we validate the key
  directly via validate_license() (mirroring how kadence-blocks-pro validates
  its own legacy license). Revoked or expired legacy keys still fail validation,
  so we never ship a stale key. The result is cached in a transient keyed by
  slug + key (WEEK_IN_SECONDS) to avoid hitting the licensing server on every
  editor/REST load.

  🎫  [SMTNC-1449](https://linear.app/nexcess/issue/SMTNC-1449/legacy-licenses-cannot-import-premium-design-patterns-or-starter)

https://www.loom.com/share/8b98f1eb131f4529949eb76b8f0d8d86
  ...

  Checklist

  Block specific checklist (where relevant)

### DIFF
--- a/includes/helper-functions.php
+++ b/includes/helper-functions.php
@@ -14,6 +14,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 use function KadenceWP\KadenceBlocks\StellarWP\Uplink\get_authorization_token;
 use function KadenceWP\KadenceBlocks\StellarWP\Uplink\get_license_domain;
 use function KadenceWP\KadenceBlocks\StellarWP\Uplink\get_license_key;
+use function KadenceWP\KadenceBlocks\StellarWP\Uplink\get_resource;
 use function KadenceWP\KadenceBlocks\StellarWP\Uplink\is_authorized;
 use function KadenceWP\KadenceBlocks\StellarWP\Uplink\validate_license;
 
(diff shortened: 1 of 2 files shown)


includes/helper-functions.php
```diff
@@ -231,6 +241,45 @@ function_exists( 'lw_harbor_get_unified_license_key' )
...
+	if ( 'invalid' === $cached ) {
+		return false;
+	}
+
+	$validation = validate_license( $slug, $key );
+	$is_valid   = $validation && method_exists( $validation, 'is_valid' ) && $validation->is_valid();
+
+	set_transient( $transient, $is_valid ? 'valid' : 'invalid', WEEK_IN_SECONDS );
+
+	return $is_valid;
```
jonwaldstein: So Uplink already cache's when license is valid, which means we could be getting false positives. In other words, positive caching for a week risks shipping a revoked key but we can prefer Uplink's persisted `is_valid()` as the source of truth and only short-cache negatives.

Here's a suggestion that would still need to be tested.

```suggestion
	$resource = get_resource( $slug );

	// Authoritative positive signal: trust Uplink's persisted, self-maintained
	// status when it already says the key is valid.
	if ( $resource && $resource->get_license_object()->is_valid() ) {
		return true;
	}

	// No stored "valid" status yet (never validated, invalid, or revoked).
	// Short-cache the negative so we don't re-POST every load, then validate;
	// a successful validate_license() persists the status, so the fast path
	// above answers subsequent requests without another network call.
	$transient = 'kadence_blocks_legacy_key_invalid_' . md5( $slug . '_' . $key );
	if ( 'invalid' === get_transient( $transient ) ) {
		return false;
	}

	$validation = validate_license( $slug, $key );
	$is_valid   = $validation && method_exists( $validation, 'is_valid' ) && $validation->is_valid();

	if ( ! $is_valid ) {
		set_transient( $transient, 'invalid', HOUR_IN_SECONDS );
	}

	return $is_valid;
```

jonwaldstein (changes requested):

jonwaldstein (approved):
@gilbert-hernandez thanks for the update, looks good - ready for QA 🌈

PR #1015, opened 2026-06-04
[DTM] Resolver caching — wp_cache L2 layer for Token_Resolver

pauloiankoski:
## Summary

- Adds a `wp_cache_get` / `wp_cache_set` L2 cache inside `Token_Resolver::resolve()`, between the existing per-request L1 memo and the cold document-walk
- Cache key is `resolved_tokens_{slug}_{version}` in group `kb_design_tokens`; TTL is 0 — invalidation is implicit via version rotation on every store write
- Adds three wpunit tests: cold-path cache write, L2 hit (sentinel), and implicit invalidation on version bump

## Test plan

- [x] `slic run wpunit Resources/Design_Tokens/Resolver/Token_ResolverTest` — 13/13 passing
- [ ] Confirm object-cache hits are observed in a staging environment with a persistent cache (Memcached or Redis)

### DIFF
--- a/includes/resources/Design_Tokens/Resolver/Token_Resolver.php
+++ b/includes/resources/Design_Tokens/Resolver/Token_Resolver.php
@@ -21,6 +21,7 @@ final class Token_Resolver {
 	 * Object-cache group shared by all resolved-token entries.
 	 *
 	 * @since TBD
+	 *
 	 * @var string
 	 */
 	private const CACHE_GROUP = 'kb_design_tokens';
--- a/includes/resources/Design_Tokens/Resolver/Token_Resolver.php
+++ b/includes/resources/Design_Tokens/Resolver/Token_Resolver.php
@@ -76,7 +76,9 @@ final class Token_Resolver {
 		$cached    = wp_cache_get( $cache_key, self::CACHE_GROUP, false, $found );
 
 		if ( $found && $cached instanceof Resolved_Tokens ) {
-			return $this->memo[ $key ] = $cached;
+			$this->memo[ $key ] = $cached;
+
+			return $cached;
 		}
 
 		$raw      = $this->store->get_document( $slug );
--- a/includes/resources/Design_Tokens/Resolver/Token_Resolver.php
+++ b/includes/resources/Design_Tokens/Resolver/Token_Resolver.php
@@ -17,6 +17,12 @@ use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Layers;
  */
 final class Token_Resolver {
 
+	/**
+	 * Object-cache group shared by all resolved-token entries.
+	 *
+	 * @since TBD
+	 * @var string
+	 */
 	private const CACHE_GROUP = 'kb_design_tokens';
 
 	/**
--- a/includes/resources/Design_Tokens/Resolver/Token_Resolver.php
+++ b/includes/resources/Design_Tokens/Resolver/Token_Resolver.php
@@ -17,6 +17,8 @@ use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Layers;
  */
 final class Token_Resolver {
 
+	private const CACHE_GROUP = 'kb_design_tokens';
+
 	/**
 	 * @var Token_Store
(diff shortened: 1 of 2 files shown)


includes/resources/Design_Tokens/Resolver/Token_Resolver.php
```diff
@@ -69,14 +71,27 @@ public function resolve( string $slug = 'default' ): Resolved_Tokens {
...
+
 		$raw      = $this->store->get_document( $slug );
 		$decoded  = $raw === '' ? [] : json_decode( $raw, true );
 		$over     = is_array( $decoded ) ? $decoded : [];
 		$document = $this->effective->build( $over );
 
-		$this->memo[ $key ] = $this->resolve_document( $document );
+		$result = $this->resolve_document( $document );
+
+		wp_cache_set( $cache_key, $result, self::CACHE_GROUP );
```
dave-green-uk: Wondering if we need to set the fourth param to expire the cache entries, otherwise they will build up? Thinking `DAY_IN_SECONDS` or similar?

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
```diff
@@ -69,14 +71,27 @@ public function resolve( string $slug = 'default' ): Resolved_Tokens {
 			return $this->memo[ $key ];
 		}
 
+		// L2: object cache — survives across requests, keyed on the store version.
```
dave-green-uk: Might be worth updating this to state that this is only the cache with a persistent object cache drop-in enabled.

dave-green-uk (approved):
I changed the base branch to `SOFT-3380/token-resolver` to simplify the review.

LGTM. I left two suggestions.

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
d4mation: We could probably go lower since the cache "soft-invalidates" whenever the version changes, but `DAY_IN_SECONDS` should be safe :) That'll ensure they're cleared out as-needed for sure.

Edit: Nah, `DAY_IN_SECONDS` is good. Ideally, we should be hitting this cache relatively often once they've settled on the Tokens they want.

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
```diff
@@ -17,6 +17,8 @@
  */
 final class Token_Resolver {
 
+	private const CACHE_GROUP = 'kb_design_tokens';
```
d4mation, line 27: Please add a variable doc with `@since` and the type :)

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
pauloiankoski: Good call — added `DAY_IN_SECONDS` as the TTL in commit 0e1af359daced8548a02a7084e164a4b3937a345.

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
pauloiankoski: Good point — updated the comment in commit 2b770abbdd408ee500c47237d095fed27099ce70 to mention it requires a persistent object-cache drop-in (Memcached/Redis).

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
pauloiankoski: Added in commit bad41c4.

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
```diff
@@ -17,6 +17,14 @@
  */
 final class Token_Resolver {
 
+	/**
+	 * Object-cache group shared by all resolved-token entries.
+	 *
+	 * @since TBD
+	 * @var string
```
d4mation, line 25:
```suggestion
	 * @since TBD
	 *
	 * @var string
```

d4mation (approved):

PR #1016, opened 2026-06-04
feat: Design Tokens REST controller skeleton (SOFT-3387)

d4mation:
🎫 [SOFT-3387](https://linear.app/nexcess/issue/SOFT-3387/rest-controller-skeleton)

Skeleton for the Design Tokens REST surface — no endpoints yet. The read/write/variant/Design MD tickets extend this base.

- Abstract `Rest\V1\Controller` (extends `WP_REST_Controller`): owns the `kb-design-tokens/v1` namespace and the shared capability gate (`edit_theme_options`, filterable via `kadence_blocks_design_tokens_capability`, request-aware for per-route overrides); delegates the standard `*_permissions_check()` methods to one shared check.
- Versioned wiring: `Rest\Provider` boots version sub-providers; `Rest\V1\Provider` registers concrete controllers on `rest_api_init` (empty seam for now). A future v2 is a parallel `Rest\V2\` namespace plus one line.
- wpunit coverage of the base (namespace, capability default/filter, grant/deny).

### Deviations from the EngDoc (§11)
- **Nonce:** the doc has the UI pass the `kadence-blocks-ajax-verification` nonce as `X-WP-Nonce`. That is an `admin-ajax.php` nonce; WP core verifies `X-WP-Nonce` only against the `wp_rest` action, so reusing it returns `403 rest_cookie_invalid_nonce`. This relies on the standard `wp_rest` nonce (automatic via `apiFetch`) with the capability check as the real gate — matching existing controllers.
- **REST namespace:** the doc specifies `kbp/v1`, which is the legacy posts-controller namespace. Every modern controller uses `kb-<feature>/v1`, so this uses `kb-design-tokens/v1`.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
@@ -8,18 +8,11 @@ use WP_REST_Controller;
 use WP_REST_Request;
 
 /**
- * Base controller for the Design Tokens REST surface.
+ * Abstract base for the v1 Design Tokens REST controllers.
  *
- * A thin, well-typed boundary over the store, registry, and resolver: concrete
- * controllers extend this to expose read/write, variant, and Design MD routes.
- * It supplies the shared namespace and the single capability gate every route
- * is protected by, leaning on WP_REST_Controller for everything else (schema,
- * collection params, the standard *_permissions_check() seams).
- *
- * Authentication uses the standard WordPress REST flow: the editor calls these
- * routes through apiFetch, which attaches the core wp_rest nonce as X-WP-Nonce,
- * and MCP clients authenticate with application passwords. The capability check
- * is the real gate.
+ * Provides the shared `kb-design-tokens/v1` namespace and a capability gate
+ * applied to every route through the standard permission-check methods. Concrete
+ * controllers extend this to register their routes and implement their handlers.
  *
  * @since TBD
  */
--- a/includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
@@ -51,24 +51,24 @@ abstract class Controller extends WP_REST_Controller {
 	/**
 	 * The capability a request must satisfy to access design-token routes.
 	 *
-	 * The request, when available, is passed through to the filter so a callback
-	 * can vary the requirement per route or method (for example, requiring a
-	 * higher capability on writes than reads) without registering a new filter.
+	 * The request is passed through to the filter so a callback can vary the
+	 * requirement per route or method (for example, requiring a higher capability
+	 * on writes than reads) without registering a new filter.
 	 *
 	 * @since TBD
 	 *
-	 * @param WP_REST_Request|null $request The current request, if any.
+	 * @param WP_REST_Request $request The current request.
 	 *
 	 * @return string
 	 */
-	public function get_capability( ?WP_REST_Request $request = null ): string {
+	public function get_capability( WP_REST_Request $request ): string {
 		/**
 		 * Filters the capability required to access the Design Tokens REST routes.
 		 *
 		 * @since TBD
 		 *
-		 * @param string               $capability The required capability.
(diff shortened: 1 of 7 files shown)


includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
```diff
@@ -0,0 +1,161 @@
...
+	 */
+	private const DEFAULT_CAPABILITY = 'edit_theme_options';
+
+	/**
+	 * Sets the shared REST namespace for every Design Tokens route.
+	 *
+	 * @since TBD
+	 */
+	public function __construct() {
+		$this->namespace = self::REST_NAMESPACE;
```
d4mation: `WP_REST_Controller` has `protected $namespace` without a type. This makes phpstan angry and would have [required us to confirm that it isn't empty](https://github.com/stellarwp/harbor/pull/169).

Having it be a `private const` that is assigned to `protected $namepsace` it _knows_ cannot be empty gets around this problem.

includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
```diff
@@ -0,0 +1,161 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 *
+	 * @return true|WP_Error
+	 */
+	public function get_items_permissions_check( $request ) {
+		return $this->permission_check( $request );
+	}
```
d4mation, line 84: For our purposes, all the different permissions checks likely are going to be the same. But extending Controllers can choose to modify this, or it can simply be filtered above as we have the full Request object there.

includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
d4mation: Lol, just kidding. This doesn't solve this issue. I was hoping it would. I'll make another small commit to see if maybe there's a good way around it.

d4mation: Fixed with [`a87c508` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1016/commits/a87c5088a84f7d4a9604ec31b4adad87f6c42113)

includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
```diff
@@ -0,0 +1,158 @@
...
+	 *
+	 * Redeclared from WP_REST_Controller both to set the shared value and to narrow the type to a
+	 * guaranteed non-empty string, so concrete controllers can pass it straight to register_rest_route()
+	 * without re-narrowing it at each call.
+	 *
+	 * @since TBD
+	 *
+	 * @var non-falsy-string
+	 */
+	protected $namespace = 'kb-design-tokens/v1';
```
dave-green-uk, line 44: Observation - in the ticket and doc we say `kbp/v1`, we should probably update to reflect this alternative name?

includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
```diff
@@ -0,0 +1,158 @@
+<?php declare( strict_types=1 );
```
dave-green-uk, line 1: Observation: we mention a verification nonce as being part of the scope, are we dropping that, or just implementing in a future PR?

dave-green-uk (approved):
LGTM! Left two observation comments.

includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
d4mation: I clarified in the PR Description that this deviation is intentional :)

includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
d4mation: I clarified in the PR Description that this deviation is intentional :)

dave-green-uk (approved):
LGTM!

PR #1017, opened 2026-06-04
[DTM] Add CSS-variable backbone (Css_Var projector)

pauloiankoski:
## Summary

- Adds `Css_Var\Projector` — emits `--kb-token--*` declarations and a `--wp--preset--*: var(--kb-token--*)` bridge, scoped to `:root,:root:where(.kb-tokens)`, never `!important`
- Adds `Css_Var\Legacy_Filter_Bridge` — feeds `kadence_blocks_pattern_global_colors` and `kadence_blocks_variable_font_sizes` via `var(--kb-token--…, <literal>)` so existing blocks inherit token values without per-block changes; no-op under the Kadence theme
- Adds `Css_Var\Hooks` — all WordPress hook callbacks in one place; `enqueue_editor` is gated to known block-editor page slugs via `kadence_blocks_load_editor_token_vars` filter
- Adds `Projection\Provider` — thin wiring layer: registers `Projector`, `Legacy_Filter_Bridge` and `Hooks` as singletons then attaches them to `wp_enqueue_scripts:100`, `admin_init:5`, and the two legacy filters at priority 20; all paths gated on `Token_Registry::is_active()`
- Adds `Projection/Contracts/Projector` — marker interface for all projectors; output types vary so no method is enforced, but the interface provides a named type for container registrations and future tooling
- Adds per-request memo + object-cache layer to `Projector::css_for_version()`, keyed on `KADENCE_BLOCKS_VERSION + store_version`
- Guards all `resolve()` call sites in WP hook callbacks against `RuntimeException` (alias cycle / dangling alias) so a corrupt stored document degrades gracefully rather than fatalling the page

## Test plan

- [ ] `slic run wpunit Resources/Design_Tokens/Projection/Css_Var/Wp_Preset_VarTest`
- [ ] `slic run wpunit Resources/Design_Tokens/Projection/Css_Var/ProjectorTest`
- [ ] `slic run wpunit Resources/Design_Tokens/Projection/Css_Var/Legacy_Filter_BridgeTest`
- [ ] `slic run wpunit Resources/Design_Tokens/Projection/Projection_ProviderTest`
- [ ] `slic run snapshot Resources/Design_Tokens/Projection/Css_Var/Projector_SnapshotTest`
- [ ] `composer phpcs` passes with no errors
- [ ] `npx cspell` passes on changed files


dave-green-uk (commented):
A few non-blocking nits from review (the PR looks solid — clean projector/Provider split, correct hook timing vs init.php, and fail-open verified since both resolver exceptions extend RuntimeException). Details inline.

includes/resources/Design_Tokens/Projection/Provider.php
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		/** @var Css_Var_Projector $projector */
+		$projector = $this->container->get( Css_Var_Projector::class );
+
+		return $projector->css_for_version( $resolved, $version );
+	}
+
+	/**
+	 * Whether token projection is active. Fail-closed: any error reading the registry means "off".
```
dave-green-uk: Doc/impl mismatch: this says "Fail-closed: any error reading the registry means off," but `is_active()` doesn't actually catch anything — if `container->get()` or `is_active()` threw, it would propagate (unlike `build_css()`, which *is* properly guarded). Either wrap this in a try/catch to honour the fail-closed contract, or soften the docblock to match. Minor.

includes/resources/Design_Tokens/Projection/Provider.php
```diff
@@ -0,0 +1,158 @@
...
+		$this->container->singleton( Css_Var_Projector::class, Css_Var_Projector::class );
+		$this->container->singleton( Legacy_Filter_Bridge::class, Legacy_Filter_Bridge::class );
+
+		// Front end: append our declarations to the global-variables handle (KB enqueues at 90).
+		add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_front_end' ], 91 );
+
+		// Editor: the editor-styles handle is registered at admin_init priority 1 and wired as a
+		// wp-block-library dependency at priority 2. Appending at priority 5 ensures the handle
+		// exists and keeps that dependency, which is what gives editor-iframe coverage for free.
+		add_action( 'admin_init', [ $this, 'enqueue_editor' ], 5 );
```
dave-green-uk: `enqueue_editor` fires on every `admin_init` (all wp-admin pages), so `build_css()` (resolve + cache read) runs even on screens where the editor-styles handle is never enqueued and the CSS is discarded. It's cached so the cost is low, and it mirrors KB's own `admin_init:1` build — but a block-editor context check would skip the needless work on unrelated admin pages. Optional.

includes/resources/Design_Tokens/Projection/Css_Var/Projector.php
```diff
@@ -0,0 +1,255 @@
...
+	 * Each preset points at its token's variable rather than the literal value, so the two can never
+	 * disagree and a token change updates both in one place.
+	 *
+	 * @since TBD
+	 *
+	 * @param Resolved_Tokens $resolved The resolved token maps.
+	 *
+	 * @return string
+	 */
+	private function preset_block( Resolved_Tokens $resolved ): string {
```
dave-green-uk, line 163: Unlike `token_block()`, `preset_block()` doesn't run `sanitize_value()` over `category`/`slug`/`css_var`. That's fine — those come from developer-declared registry config, not store/user input — but since the sibling method *does* sanitise store-derived values, a one-line comment noting the trust boundary would save the next reader the double-take. Nit.

includes/resources/Design_Tokens/Registry/Token_Registry.php
```diff
@@ -233,4 +233,15 @@ public function missing_from_baseline( Baseline_Document $baseline ): array {
...
 	}
+
+	/**
+	 * Re-enable projection after a previous deactivate() call.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function activate(): void {
```
dave-green-uk, line 244: `activate()`'s only caller is `Projection_ProviderTest` — a production method that currently exists purely for test setup. It's symmetric with `deactivate()` so it's defensible, but worth a second look: could the test exercise the deactivated path from the default state instead, or is runtime reactivation a real need? Nit.

includes/resources/Design_Tokens/Projection/Css_Var/Projector.php
d4mation, line 1: I wonder if we could use `Design_Tokens/Projection/Css_Var/Projector.php` instead?

Then these other related files, which wouldn't relate to other Projections, are not all at the root `Projection` namespace but kept within `Css_Var`

includes/resources/Design_Tokens/Projection/Css_Var_Projector.php
```diff
@@ -0,0 +1,255 @@
...
+			return [ $config['category'], $slug ];
+		}
+
+		return [ '', '' ];
+	}
+
+	/**
+	 * The last dot-segment of a token id, used as the default preset slug.
+	 *
+	 *   semantic.color.button-bg => button-bg
```
d4mation: The extra spaces here are weird. Should it instead read like this?

```suggestion
	 * Example: semantic.color.button-bg => button-bg
```

includes/resources/Design_Tokens/Projection/Provider.php
```diff
@@ -0,0 +1,158 @@
...
+	 * @inheritDoc
+	 *
+	 * @since TBD
+	 */
+	public function register(): void {
+		$this->container->singleton( Css_Var_Projector::class, Css_Var_Projector::class );
+		$this->container->singleton( Legacy_Filter_Bridge::class, Legacy_Filter_Bridge::class );
+
+		// Front end: append our declarations to the global-variables handle (KB enqueues at 90).
+		add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_front_end' ], 91 );
```
d4mation: This Provider is too smart. We should avoid having so much responsibility in a Provider. Ideally, it should be responsible for registering how a Class should be built by the Container, registering hooks, and nothing else.

tests/wpunit/Resources/Design_Tokens/Projection/Css_Var/Css_BuilderTest.php
d4mation, line 1: Now that we're working with Projections, the Snapshot Test suite may start to be more appropriate for these types of tests.

You can still do assertions within them to make sure certain things (like `:root`) are in the proper location, but then we're also asserting that it matches compared to a stored snapshot file which can be easier to track for generated changes like this.

d4mation (commented):
Still going through the rest of the PR, but wanted to get these comments through early. :)

Edit: Done for now. :)

dave-green-uk (changes requested):
Generally LGTM! Just a couple of nits that were added.

includes/resources/Design_Tokens/Projection/Css_Var/Projector.php
d4mation, line 1: Maybe we should have a Projector Contract? It would likely need to be very basic as different Projectors do different things, but that would give us some level of consistency between Projectors.

tests/wpunit/Resources/Design_Tokens/Projection/Legacy_Filter_BridgeTest.php
```diff
@@ -0,0 +1,338 @@
...
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
+use Tests\Support\Classes\Fake_Baseline_Document;
+use Tests\Support\Classes\TestCase;
+
+final class Legacy_Filter_BridgeTest extends TestCase {
+
+	private Token_Registry $registry;
+
+	/**
+	 * The default palette shipped by KB's init.php — keys and values must stay in lockstep.
```
d4mation: Probably better to reference the relative path directly. Especially since when saying "shipped by KB's" it makes it sound like it is a different plugin 😅 

I wish there was a way we could just grab these via the code directly, but it looks like there isn't a good way to do so because of how they're defined :(

```suggestion
	 * The default palette included in `includes/init.php` — keys and values must stay in lockstep.
```

includes/resources/Design_Tokens/Projection/Provider.php
```diff
@@ -0,0 +1,158 @@
...
+
+		// Editor: the editor-styles handle is registered at admin_init priority 1 and wired as a
+		// wp-block-library dependency at priority 2. Appending at priority 5 ensures the handle
+		// exists and keeps that dependency, which is what gives editor-iframe coverage for free.
+		add_action( 'admin_init', [ $this, 'enqueue_editor' ], 5 );
+
+		// Legacy variable families (init.php applies these in both editor and front-end functions).
+		// Merge semantics: the bridge overrides only token-claimed slots; everything else passes through.
+		add_filter( 'kadence_blocks_pattern_global_colors', [ $this, 'filter_global_colors' ] );
+		add_filter( 'kadence_blocks_variable_font_sizes', [ $this, 'filter_font_sizes' ] );
```
d4mation: What if someone else has filtered this at priority `10`? We probably should filter this later to attempt to ensure our code is what runs. A multiple of `10` like `20` is good because then if someone _wants_ to run before or after us, they easily can.

includes/resources/Design_Tokens/Projection/Provider.php
d4mation: I don't think this should be optional. I think we need to have this check. We could make that check be filterable to _force_ it to enqueue the scripts, but we do need to try to guard it.

includes/resources/Design_Tokens/Projection/Provider.php
d4mation: Also, priority `100` would be cleaner. Then if someone needs to enqueue between the default functionality and us they can.

includes/resources/Design_Tokens/Projection/Provider.php
pauloiankoski: Updated the docblock — the fail-closed note belonged on the class docblock, not here. Container resolution of a registered binding cannot throw; the guard is the boolean returned by `Token_Registry::is_active()`. Fix in commit 09b354972.

includes/resources/Design_Tokens/Projection/Provider.php
pauloiankoski: Added the guard with a `kadence_blocks_load_editor_token_vars` filter so callers can force-load on custom editor pages. Fix in commit e35895495.

includes/resources/Design_Tokens/Projection/Css_Var/Projector.php
pauloiankoski: Added a one-line comment noting the trust boundary. Fix in commit ab14f6232.

includes/resources/Design_Tokens/Registry/Token_Registry.php
pauloiankoski: `activate()` is a runtime hook for the admin force re-enable action planned alongside the admin UI. The test just happens to be its first caller. Keeping it.

includes/resources/Design_Tokens/Projection/Css_Var/Projector.php
pauloiankoski: Done in this PR. Mechanical namespace migration — no logic change. Fix in commit 87297484a.

includes/resources/Design_Tokens/Projection/Css_Var_Projector.php
pauloiankoski: Fixed. Commit 330d3068a.

includes/resources/Design_Tokens/Projection/Provider.php
pauloiankoski: Extracted all hook callbacks to `Css_Var\Hooks`; Provider is now three `singleton()` calls plus four hook registrations. Also bumped `wp_enqueue_scripts` to priority 100 and both filters to 20. Fix in commit e35895495.

tests/wpunit/Resources/Design_Tokens/Projection/Css_Var/Css_BuilderTest.php
pauloiankoski: Added in this PR, sequenced after the namespace migration commit so the path and `use` statements are correct from day one. Structural assertions stay in `Css_Var/ProjectorTest` (fast, no snapshot file to regenerate); the snapshot test adds golden-file coverage on top. Fix in commit 0138cb4a0.

includes/resources/Design_Tokens/Projection/Css_Var/Projector.php
pauloiankoski: Added in this PR. The interface is intentionally minimal (marker + rationale docblock) so the shape can be extended when the second projector lands without breaking `final class` constraints. Fix in commit a5724af87.

tests/wpunit/Resources/Design_Tokens/Projection/Legacy_Filter_BridgeTest.php
pauloiankoski: Updated both docblocks to reference the relative path. Fix in commit 3b0dc4096.

includes/resources/Design_Tokens/Projection/Provider.php
pauloiankoski: Bumped both filters to priority 20 in the same commit as the Provider refactor. Fix in commit e35895495.

includes/resources/Design_Tokens/Projection/Provider.php
```diff
@@ -0,0 +1,44 @@
...
+
+	/**
+	 * @inheritDoc
+	 *
+	 * @since TBD
+	 */
+	public function register(): void {
+		$this->container->singleton( Projector::class, Projector::class );
+		$this->container->singleton( Legacy_Filter_Bridge::class, Legacy_Filter_Bridge::class );
+		$this->container->singleton( Hooks::class, Hooks::class );
```
d4mation: We shouldn't need to define the second param. It is only needed if you need to build it in a special way (like this example: https://github.com/stellarwp/learndash-core/blob/f37e608b945e43d6f74a17b51db27e27c5a73444/src/Core/Modules/Reports/Legacy/Settings/Provider.php#L32-L40)

includes/resources/Design_Tokens/Projection/Provider.php
```diff
@@ -0,0 +1,44 @@
...
+
+		/** @var Hooks $hooks */
+		$hooks = $this->container->get( Hooks::class );
+
+		// Front end: append our declarations to the global-variables handle (KB enqueues at 90).
+		add_action( 'wp_enqueue_scripts', [ $hooks, 'enqueue_front_end' ], 100 );
```
d4mation: All hooks should be attached to Container Callbacks: https://github.com/lucatume/di52#the-callback-method

includes/resources/Design_Tokens/Projection/Provider.php
```diff
@@ -0,0 +1,44 @@
...
+	 *
+	 * @since TBD
+	 */
+	public function register(): void {
+		$this->container->singleton( Projector::class, Projector::class );
+		$this->container->singleton( Legacy_Filter_Bridge::class, Legacy_Filter_Bridge::class );
+		$this->container->singleton( Hooks::class, Hooks::class );
+
+		/** @var Hooks $hooks */
+		$hooks = $this->container->get( Hooks::class );
```
d4mation: A generic "Hooks" class feels incorrect. This functionality seems pretty tightly coupled to the CSS Var Projector, right? Should it go there? 

I wonder if the real issue here is that these hooked methods are logically the "Projector" and that the logic in `Projector.php` right now is what they call to function. 

We should rename the `Projector.php` class to something that makes sense and then rename `Hooks.php` to `Projector.php` as the logic in there is what is _actually_ doing the Projecting.

includes/resources/Design_Tokens/Projection/Contracts/Projector.php
```diff
@@ -0,0 +1,17 @@
...
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection\Contracts;
+
+/**
+ * Marker contract for all design-token projectors.
+ *
+ * Projectors consume a resolved token set and produce output in a format specific to their target
+ * (CSS custom properties, theme.json preset array, REST payload, etc.). The output type varies
+ * across projectors, so no method signature is enforced here — each concrete class declares its
+ * own. The interface provides a named type for container registrations and makes projectors
+ * discoverable by tooling or admin UI that enumerates available projectors.
```
d4mation: > The output type varies
> across projectors, so no method signature is enforced here — each concrete class declares its
> own. The interface provides a named type for container registrations and makes projectors
> discoverable by tooling or admin UI that enumerates available projectors.
 
 Yeah, I was hoping that we _could_ find a way to have at least a `run()` method or something, but it may not be possible. If that's the case, this Contract will just be dead code. It helps tie them together conceptually, but so does the Namespace.
 
 As long as we do `Projection/<name>/Projector.php` for each Projector, I think that's enough. We can just delete this Contract. Sorry!

includes/resources/Design_Tokens/Projection/Provider.php
pauloiankoski: I was concerned about that provider becoming too bloated, so I went with the sub-provider pattern instead. This allows each projector to have its own provider, keeping responsibilities isolated. I think this approach will make the codebase more organized and better encapsulated as it grows. df0c70d56598b72e8e01a20bb429c10376864177.

Also renamed many classes in 6fc4c2ddeada658fc34a203192c0b6bf438d2030.

includes/resources/Design_Tokens/Projection/Contracts/Projector.php
pauloiankoski: Removed the contract in 13f8e3b4c9ee72f7a088285b12367cb390097404

includes/resources/Design_Tokens/Projection/Provider.php
pauloiankoski: Implemented on 6fc4c2ddeada658fc34a203192c0b6bf438d2030 but moved to the sub-provider in df0c70d56598b72e8e01a20bb429c10376864177

includes/resources/Design_Tokens/Projection/Provider.php
pauloiankoski: Implemented on https://github.com/stellarwp/kadence-blocks/commit/6fc4c2ddeada658fc34a203192c0b6bf438d2030 but moved to the sub-provider in https://github.com/stellarwp/kadence-blocks/commit/df0c70d56598b72e8e01a20bb429c10376864177

d4mation (approved):
The reorganization looks great! The sub-provider especially should help us keep the Projectors organized and as consistent as we can make them be.

PR #1018, opened 2026-06-04
feat: Design Tokens read endpoints (SOFT-3388)

d4mation:
🎫 [SOFT-3388](https://linear.app/nexcess/issue/SOFT-3388/read-endpoints)

Adds the Design Tokens v1 REST read surface over the store, registry and resolver, building on the SOFT-3387 controller skeleton.

New routes (namespace `kb-design-tokens/v1`):

- `GET /documents` — lists the token-set documents (v1 ships the single `default` set); returns each set's slug, version and raw overrides-only DTCG document.
- `GET /documents/<slug>` — the raw overrides-only DTCG document for a set; 404 for an unknown slug.
- `GET /documents/<slug>/resolved` — the resolved/flattened token map for previews (`by_id` and `by_var` maps of CSS values); 422 when the stored document has an alias cycle or dangling alias.
- `GET /schema` — the published DTCG JSON Schema (the committed `dtcg.schema.json`).

Every route declares its args and item schema so the MCP surface can introspect request/response shapes. Write endpoints are out of scope (SOFT-3389).

### Checklist


includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
```diff
@@ -0,0 +1,397 @@
...
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 *
+	 * @return WP_REST_Response|WP_Error
+	 */
+	public function get_resolved( $request ) {
+		$slug = Cast::to_string( $request->get_param( self::SLUG_PARAM ) );
+
+		try {
+			$resolved = $this->resolver->resolve( $slug );
```
pauloiankoski, line 225: `get_item()` correctly returns a 404 for any slug that isn't `Token_Store::default_slug()`, but `get_resolved()` passes the slug directly to the resolver without performing the same validation:

```php
$resolved = $this->resolver->resolve( $slug ); // no slug guard
```

If the resolver throws anything other than `Alias_Cycle_Exception` or `Dangling_Alias_Exception` for an unknown token set, the exception will propagate uncaught and result in a 500 response.

Consider mirroring the guard used in `get_item()` at the beginning of this method:
```php
if ( $slug !== Token_Store::default_slug() ) {
	return new WP_Error(
		'rest_design_tokens_not_found',
		__( 'Sorry, that design token set does not exist.', 'kadence-blocks' ),
		[
			'status' => WP_Http::NOT_FOUND,
			'slug'   => $slug,
		]
	);
}
```

This keeps the behavior consistent across endpoints and avoids unexpected 500s for invalid slugs.

includes/resources/Design_Tokens/Rest/V1/Schema_Controller.php
```diff
@@ -0,0 +1,134 @@
...
+				echo $this->dtcg_schema->json(); // phpcs:ignore StellarWP.XSS.EscapeOutput.OutputNotEscaped, WordPress.Security.EscapeOutput.OutputNotEscaped -- Raw JSON Schema body; escaping would corrupt it.
+
+				return true;
+			},
+			10,
+			2
+		);
+
+		return $response;
+	}
```
pauloiankoski, line 126: `Dtcg_Schema::json()` returns an empty string when the schema file is missing or unreadable, but the controller currently returns that value with an HTTP 200 response. An empty response body is not valid JSON, so consumers expecting JSON may fail when attempting to parse the response.

Consider validating the result before returning it and surfacing a server error when the schema cannot be loaded:

```php
$json = $this->dtcg_schema->json();

if ( $json === '' ) {
	return new WP_Error(
		'rest_design_tokens_schema_unavailable',
		__( 'The DTCG schema file could not be read.', 'kadence-blocks' ),
		[
			'status' => WP_Http::INTERNAL_SERVER_ERROR,
		]
	);
}
```
This makes the failure explicit and avoids returning a successful response containing invalid JSON.

tests/wpunit/Resources/Design_Tokens/Rest/V1/DocumentsControllerTest.php
```diff
@@ -0,0 +1,256 @@
...
+	 * @return void
+	 */
+	public function testItDeniesAccessToUsersWithoutTheCapability(): void {
+		wp_set_current_user( $this->factory()->user->create( [ 'role' => 'subscriber' ] ) );
+
+		$result = $this->controller->get_items_permissions_check( new WP_REST_Request( WP_REST_Server::READABLE ) );
+
+		$this->assertInstanceOf( WP_Error::class, $result );
+		$this->assertSame( 'rest_forbidden', $result->get_error_code() );
+	}
```
pauloiankoski, line 231: `testItDeniesAccessToUsersWithoutTheCapability()` currently exercises only `get_items_permissions_check()`. The single-document and resolved endpoints are wired through `get_item_permissions_check()`, so a regression in that permission callback would not be caught by the existing test suite.

Consider adding a parallel test that explicitly covers `get_item_permissions_check()` for a user lacking the required capability. That would ensure both permission callbacks remain protected and help catch any future routing or wiring changes.

d4mation: [`cfd1c10` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1018/commits/cfd1c1018e207a6316608f4df9737408295f65f5)

includes/resources/Design_Tokens/Rest/V1/Schema_Controller.php
d4mation: [`cfd1c10` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1018/commits/cfd1c1018e207a6316608f4df9737408295f65f5)

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
d4mation: [`cfd1c10` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1018/commits/cfd1c1018e207a6316608f4df9737408295f65f5)

dave-green-uk (approved):
LGTM, nothing obvious jumped out at me.

Claude pointed out that `Schema_Controller` would return an empty 200 body if consumed server side rather than over HTTP, but I guess that's expected behaviour right now? 🤔

d4mation:
@dave-green-uk Generally I'd say that's not expected usage particularly because we have to change how that endpoint works to ensure we return the raw JSON (because otherwise it would require a potentially lossless decode/encode step to return the exact same data anyway) but I accounted for it with a note that it isn't the preferred way to call it :) 

[`dcfd4c3` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1018/commits/dcfd4c371592566a41b41987ecb13d4b30c89abd)

pauloiankoski (approved):

PR #1019, opened 2026-06-05
[DTM] Add foundation-preset selection plumbing

dave-green-uk:
🎫  SOFT-3397

## Summary

Adds the **foundation-preset selection plumbing** (SOFT-3397, Eng Doc §10) — the beginner on-ramp of the Design Tokens dual-mode strategy. The baseline already ships the preset catalogue (four type scales + three starter palettes); this adds the PHP service layer that **reads** those presets and **applies** a choice by seeding the `primitive` layer as store overrides. Scope is service-only and overrides-only (no REST/UI, no persisted "selected preset" marker — those are separate concerns).

 ## Changes

- **`Foundation_Presets`** — catalogue reader over the baseline `$extensions`: `groups()`, `options()`, `default_for()`, `tokens_for()`.
- **`Preset_Selector`** — `apply($group, $choice)` turns a preset's flat token map into fully-typed DTCG override leaves (each `$type` sourced from the baseline), deep-merges onto the current store overrides, and persists via `Token_Store` (which bumps the version + fires the change action, so every projector/cache reacts).
- **`Extensions` vocabulary** — single-sources the `$extensions` namespace/section/key spellings; `Dtcg_Validator` now consumes it instead of its private literals.
- **Wiring** — new `Provider` (registered in `Design_Tokens_Provider`) + two global helpers in `functions.php` (`kadence_blocks_apply_design_foundation_preset()`, `kadence_blocks_design_foundation_presets()`).
- **Tests** — `Foundation_PresetsTest` + `Preset_SelectorTest` (real-DB wpunit).

Because every preset value aliases a baseline primitive, picking a palette/scale rebrands all aliased semantics automatically — no content rewrite.

## Test plan (QA)

NOTE: I asked Claude to run these commands and the output is [here](https://github.com/stellarwp/kadence-blocks/pull/1019#issuecomment-4631658014).

No editor/UI surface yet — drive it through the global helpers via WP-CLI. Commands use `lando wp eval '…'` (swap for `lando php bin/wp-cli.phar eval '…'` if wp-cli isn't on PATH).

> ⚠️ These mutate the live token store. **Run step 0 first** to snapshot, and **step 6** to reset when done.

**0. Snapshot + view the menu** (read-only)
```bash
# Save current overrides so you can restore the exact prior state later:
lando wp eval '$s = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Database\\Token_Store"); echo "VERSION: ".$s->get_version()."\nDOC: ".$s->get_document()."\n";'

# What can a beginner pick?
lando wp eval '$p = kadence_blocks_design_foundation_presets(); foreach ($p->groups() as $g) { echo $g." (default: ".$p->default_for($g)."): ".implode(", ", array_keys($p->options($g)))."\n"; }'
```
Expected:
```
typeScale (default: majorThird): minorThird, majorThird, perfectFourth, goldenRatio
colorPalette (default: kadence): kadence, sunset, forest
```

**1. Baseline values before any change** (read-only)
```bash
lando wp eval '$r = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Resolver\\Token_Resolver"); $by = $r->resolve()->by_id(); echo "fontSize.3xl=".$by["primitive.fontSize.3xl"]."  brand.primary=".$by["primitive.color.brand.primary"]."  link=".$by["semantic.color.link"]."\n";'
```
Expected (fresh site): `fontSize.3xl=2.441rem  brand.primary=#3182CE  link=#3182CE`

**2. Apply a type scale → font sizes shift**
```bash
lando wp eval 'kadence_blocks_apply_design_foundation_preset("typeScale", "goldenRatio");'
lando wp eval '$r = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Resolver\\Token_Resolver"); echo $r->resolve()->value("primitive.fontSize.3xl")."\n";'
```
Expected: `6.854rem`. Reload the editor / front end → heading sizes jump, **no content edits**.

**3. Apply a palette → confirm the cascade**
```bash
lando wp eval 'kadence_blocks_apply_design_foundation_preset("colorPalette", "sunset");'
lando wp eval '$r = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Resolver\\Token_Resolver"); $by = $r->resolve()->by_id(); echo "brand.primary=".$by["primitive.color.brand.primary"]."  link=".$by["semantic.color.link"]."  button-bg=".$by["semantic.color.button-bg"]."\n";'
```
Expected: all three `#DD6B20`. Only `brand.primary` was written, yet `link`/`button-bg` followed — the alias cascade.

**4. Overrides stay sparse (primitives only)**
```bash
lando wp eval '$s = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Database\\Token_Store"); $d = json_decode($s->get_document(), true); echo "top-level keys: ".implode(", ", array_keys($d))."\n";'
```
Expected: `primitive` only — `semantic` is never stored; it cascades. (Also confirms steps 2 + 3 stacked without clobbering each other.)

**5. Bad input errors cleanly, writes nothing**
```bash
lando wp eval 'try { kadence_blocks_apply_design_foundation_preset("typeScale", "nope"); echo "NO EXCEPTION — BUG\n"; } catch (\Throwable $e) { echo "threw ".get_class($e).": ".$e->getMessage()."\n"; }'
```
Expected: `threw …Unknown_Preset_Exception: Unknown foundation preset "nope" in group "typeScale".` Re-run step 4 to confirm the store is unchanged.

**6. Reset to baseline (revert QA changes)**
```bash
lando wp eval '$s = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Database\\Token_Store"); $s->save_document("");'
```
Re-run step 1 → values return to `2.441rem` / `#3182CE`. (If step 0 showed pre-existing overrides, restore them with `save_document('<that JSON>')` instead.)

**7. Theme-agnostic spot check** — repeat steps 2–3 on a non-Kadence block theme (Kadence theme inactive); the shifts should still apply.

### DIFF
--- a/includes/resources/Design_Tokens/Foundation_Presets/Catalog.php
+++ b/includes/resources/Design_Tokens/Foundation_Presets/Catalog.php
@@ -13,7 +13,7 @@ use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Extensions;
  * hand-authoring tokens: typographic type scales (minor third, major third, …) and starter color
  * palettes. They live in the shipped baseline under
  * `$extensions.<namespace>.foundationPresets.<group>.<slug>` and are read-only here — applying a choice
- * (which seeds the primitive layer as store overrides) is {@see Preset_Selector}.
+ * (which seeds the primitive layer as store overrides) is {@see Selector}.
  *
  * The baseline is the single source of truth; this class never holds its own copy of the catalogue.
  *
@@ -93,7 +93,7 @@ final class Catalog {
 	}
 
 	/**
-	 * A preset's flat token map: primitive dot-path => raw value. This is what {@see Preset_Selector}
+	 * A preset's flat token map: primitive dot-path => raw value. This is what {@see Selector}
 	 * turns into store overrides.
 	 *
 	 * @since TBD
@@ -121,7 +121,7 @@ final class Catalog {
 	/**
 	 * The union of every token dot-path any preset in the group touches — the group's "footprint".
 	 *
-	 * {@see Preset_Selector} uses this to make a selection replace the group wholesale: paths the chosen
+	 * {@see Selector} uses this to make a selection replace the group wholesale: paths the chosen
 	 * preset omits are cleared rather than left behind, so switching presets can never leave a previous
 	 * preset's exclusive paths applied.
 	 *
--- a/includes/resources/Design_Tokens/Foundation_Presets/Catalog.php
+++ b/includes/resources/Design_Tokens/Foundation_Presets/Catalog.php
@@ -19,7 +19,7 @@ use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Extensions;
  *
  * @since TBD
  */
-final class Foundation_Presets {
+final class Catalog {
 
 	/**
 	 * @var Baseline_Document The shipped, read-only baseline the catalogue is read from.
@@ -116,6 +116,33 @@ final class Foundation_Presets {
 		$tokens = $set[ $choice ][ Extensions::get_tokens_key() ] ?? [];
 
 		return is_array( $tokens ) ? $tokens : [];
+	}
+
+	/**
+	 * The union of every token dot-path any preset in the group touches — the group's "footprint".
+	 *
+	 * {@see Preset_Selector} uses this to make a selection replace the group wholesale: paths the chosen
+	 * preset omits are cleared rather than left behind, so switching presets can never leave a previous
+	 * preset's exclusive paths applied.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $group The preset group key.
+	 *
+	 * @throws Unknown_Preset_Exception When the group does not exist.
+	 *
+	 * @return string[] Dot-paths, de-duplicated.
+	 */
+	public function group_paths( string $group ): array {
+		$paths = [];
+
+		foreach ( array_keys( $this->options( $group ) ) as $slug ) {
+			foreach ( array_keys( $this->tokens_for( $group, (string) $slug ) ) as $path ) {
+				$paths[ (string) $path ] = true;
+			}
+		}
+
+		return array_keys( $paths );
 	}
 
 	/**
--- a/includes/resources/Design_Tokens/Foundation_Presets/Preset_Selector.php
+++ b/includes/resources/Design_Tokens/Foundation_Presets/Preset_Selector.php
@@ -6,6 +6,7 @@ use KadenceWP\KadenceBlocks\Design_Tokens\Database\Token_Store;
 use KadenceWP\KadenceBlocks\Design_Tokens\Foundation_Presets\Exception\Unknown_Preset_Exception;
 use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Contracts\Baseline_Document;
 use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Sentinels;
+use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Token_Type;
 use RuntimeException;
 
 /**
@@ -17,6 +18,11 @@ use RuntimeException;
  * the baseline leaf at the same dot-path, because the Resolver's deep-merge replaces a token leaf
  * WHOLESALE, so an override missing its $type would break the per-$type CSS renderer downstream.
  *
+ * Selecting a preset is a **clean swap**: the group's whole footprint (every path any preset in the
+ * group can touch) is cleared first, then the chosen preset's paths are written. So switching presets
+ * never leaves an earlier preset's exclusive paths applied — and a manual override within the footprint
+ * is discarded by a later preset selection. Re-selecting the current preset is a no-op (no write).
+ *
  * Persisting through {@see Token_Store::save_document()} bumps the set version and fires
  * `kadence_blocks_design_tokens_changed`, so every projector and cache reacts automatically.
  *
@@ -25,18 +31,11 @@ use RuntimeException;
 final class Preset_Selector {
 
 	/**
-	 * @var string The DTCG leaf key carrying a token's type.
+	 * @var Catalog The catalogue reader the chosen preset's tokens and the group footprint come from.
 	 *
 	 * @since TBD
 	 */
-	private const TYPE_KEY = '$type';
-
-	/**
-	 * @var Foundation_Presets The catalogue reader the chosen preset's tokens come from.
-	 *
-	 * @since TBD
-	 */
-	private Foundation_Presets $presets;
+	private Catalog $catalog;
 
 	/**
 	 * @var Baseline_Document The baseline each override leaf's $type is sourced from.
@@ -55,12 +54,12 @@ final class Preset_Selector {
 	/**
 	 * @since TBD
 	 *
-	 * @param Foundation_Presets $presets  The catalogue reader.
-	 * @param Baseline_Document  $baseline The shipped baseline document.
-	 * @param Token_Store        $store    The persistence gateway.
+	 * @param Catalog           $catalog  The catalogue reader.
+	 * @param Baseline_Document $baseline The shipped baseline document.
+	 * @param Token_Store       $store    The persistence gateway.
 	 */
-	public function __construct( Foundation_Presets $presets, Baseline_Document $baseline, Token_Store $store ) {
-		$this->presets  = $presets;
+	public function __construct( Catalog $catalog, Baseline_Document $baseline, Token_Store $store ) {
+		$this->catalog  = $catalog;
 		$this->baseline = $baseline;
 		$this->store    = $store;
 	}
@@ -84,17 +83,32 @@ final class Preset_Selector {
 
 		// Resolve first: an unknown choice throws here, before the store is touched, so a bad call writes
 		// nothing.
-		$tokens = $this->presets->tokens_for( $group, $choice );
+		$tokens = $this->catalog->tokens_for( $group, $choice );
 
 		$document = $this->current_overrides( $slug );
+		$original = $document;
+
+		// Clear the group's whole footprint, then write the chosen preset — a clean swap, so a preset the
+		// choice omits can never persist from a previous selection.
+		foreach ( $this->catalog->group_paths( $group ) as $dot_path ) {
+			$this->unset_leaf( $document, $dot_path );
+		}
 
 		foreach ( $tokens as $dot_path => $value ) {
 			$leaf = [
-				self::TYPE_KEY             => $this->baseline_type( (string) $dot_path ),
+				Token_Type::get_type_key() => $this->baseline_type( (string) $dot_path ),
 				Sentinels::get_value_key() => $value,
 			];
 
 			$this->set_leaf( $document, (string) $dot_path, $leaf );
+		}
+
+		// No-op guard: re-selecting the current preset (or any selection that nets no change) leaves the
+		// document equal to what is stored, so skip the write rather than bump the version, fire the
+		// changed action and churn history for nothing. Loose equality ignores key order, which clearing
+		// then re-writing the same paths can shuffle.
+		if ( $document == $original ) { // phpcs:ignore Universal.Operators.StrictComparisons.LooseEqual
+			return;
 		}
 
 		$json = wp_json_encode( $document );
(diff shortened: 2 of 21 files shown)


dave-green-uk:
# Foundation Presets — QA Run Log (SOFT-3397)

**Result: ✅ all steps passed.** Walked through the PR #1019 test plan against the live dev token store and captured real output.

| Step | Check | Expected | Actual | |
|---|---|---|---|---|
| 0 | Snapshot + catalogue | menu lists 4 scales + 3 palettes | matched | ✅ |
| 1 | Baseline resolved values | `2.441rem` / `#3182CE` | matched | ✅ |
| 2 | Apply type scale | `fontSize.3xl` → `6.854rem` | `6.854rem` | ✅ |
| 3 | Apply palette + cascade | `brand`/`link`/`button-bg` all `#DD6B20` | all `#DD6B20` | ✅ |
| 4 | Overrides stay sparse | top-level key `primitive` only | `primitive` | ✅ |
| 5 | Bad input | throws `Unknown_Preset_Exception`, no write | threw, store unchanged | ✅ |
| 6 | Reset to baseline | values back to `2.441rem` / `#3182CE` | matched | ✅ |

**Environment notes**
- Run via `lando php bin/wp-cli.phar eval '…'` (this box has no `lando wp` binary; the commands below show the canonical `lando wp eval` form, which is equivalent).
- Unrelated WP "block namespace prefix" admin notices were filtered from the captured output.
- The store was **empty before the run** (no pre-existing overrides), so step 6 restores the true as-found state. (After reset, `get_document()` is `""` — an empty-doc row that renders identically to baseline.)

---

## Step 0 — Snapshot + view the menu (read-only)

```bash
lando wp eval '$s = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Database\\Token_Store"); echo "VERSION: ".$s->get_version()."\nDOC: ".$s->get_document()."\n";'
lando wp eval '$p = kadence_blocks_design_foundation_presets(); foreach ($p->groups() as $g) { echo $g." (default: ".$p->default_for($g)."): ".implode(", ", array_keys($p->options($g)))."\n"; }'
```
**Output:**
```
VERSION: 
DOC: 
typeScale (default: majorThird): minorThird, majorThird, perfectFourth, goldenRatio
colorPalette (default: kadence): kadence, sunset, forest
```
Store empty; catalogue lists all shipped presets. ✅

## Step 1 — Baseline resolved values (read-only)

```bash
lando wp eval '$r = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Resolver\\Token_Resolver"); $by = $r->resolve()->by_id(); echo "fontSize.3xl=".$by["primitive.fontSize.3xl"]."  brand.primary=".$by["primitive.color.brand.primary"]."  link=".$by["semantic.color.link"]."\n";'
```
**Output:**
```
fontSize.3xl=2.441rem  brand.primary=#3182CE  link=#3182CE
```
Renders entirely from baseline (majorThird scale, Kadence palette). ✅

## Step 2 — Apply a type scale → font sizes shift

```bash
lando wp eval 'kadence_blocks_apply_design_foundation_preset("typeScale", "goldenRatio");'
lando wp eval '$r = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Resolver\\Token_Resolver"); echo $r->resolve()->value("primitive.fontSize.3xl")."\n";'
```
**Output:**
```
applied
6.854rem
```
`fontSize.3xl` jumps from `2.441rem` → `6.854rem`, no content edits. ✅

## Step 3 — Apply a palette → confirm the cascade

```bash
lando wp eval 'kadence_blocks_apply_design_foundation_preset("colorPalette", "sunset");'
lando wp eval '$r = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Resolver\\Token_Resolver"); $by = $r->resolve()->by_id(); echo "brand.primary=".$by["primitive.color.brand.primary"]."  link=".$by["semantic.color.link"]."  button-bg=".$by["semantic.color.button-bg"]."\n";'
```
**Output:**
```
applied
brand.primary=#DD6B20  link=#DD6B20  button-bg=#DD6B20
```
Only `brand.primary` was written, yet `link` and `button-bg` followed — the alias cascade works. ✅

## Step 4 — Overrides stay sparse (primitives only)

```bash
lando wp eval '$s = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Database\\Token_Store"); $d = json_decode($s->get_document(), true); echo "top-level keys: ".implode(", ", array_keys($d))."\n";'
```
**Output:**
```
top-level keys: primitive
```
Both the scale (step 2) and palette (step 3) coexist under `primitive`; `semantic` is never stored — it cascades. Confirms stacking doesn't clobber. ✅

## Step 5 — Bad input errors cleanly, writes nothing

```bash
lando wp eval 'try { kadence_blocks_apply_design_foundation_preset("typeScale", "nope"); echo "NO EXCEPTION - BUG\n"; } catch (\Throwable $e) { echo "threw ".get_class($e).": ".$e->getMessage()."\n"; }'
```
**Output:**
```
threw KadenceWP\KadenceBlocks\Design_Tokens\Foundation_Presets\Exception\Unknown_Preset_Exception: Unknown foundation preset "nope" in group "typeScale".
```
Throws before touching the store (step 4's `primitive`-only state is unchanged). ✅

## Step 6 — Reset to baseline (revert QA changes)

```bash
lando wp eval '$s = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Database\\Token_Store"); $s->save_document("");'
lando wp eval '$r = kadence_blocks()->get("KadenceWP\\KadenceBlocks\\Design_Tokens\\Resolver\\Token_Resolver"); $by = $r->resolve()->by_id(); echo "fontSize.3xl=".$by["primitive.fontSize.3xl"]."  brand.primary=".$by["primitive.color.brand.primary"]."  link=".$by["semantic.color.link"]."\n";'
```
**Output:**
```
reset
fontSize.3xl=2.441rem  brand.primary=#3182CE  link=#3182CE
```
Values snap back to baseline — clearing overrides reverts the whole site. ✅

includes/resources/Design_Tokens/Foundation_Presets/Catalog.php
d4mation, line 1: We shouldn't have a class be an exact match to the namespace if we can avoid it.

includes/resources/Design_Tokens/functions.php
```diff
@@ -51,3 +53,40 @@ function kadence_blocks_register_design_variant_set( array $set ): void {
 		$registry->register_variant_set( $set );
 	}
 }
+
+if ( ! function_exists( 'kadence_blocks_design_foundation_presets' ) ) {
+	/**
+	 * The foundation-preset catalogue reader — the beginner on-ramp's "what can I pick?" surface. Thin
+	 * accessor over the Foundation_Presets service for callers (a future admin UI / MCP) that need to
```
d4mation: Referencing a "future feature" seems weird. We are planning on doing it. We should drop that from the description so we don't end up neglecting to fix it later.

includes/resources/Design_Tokens/Foundation_Presets/Preset_Selector.php
```diff
@@ -0,0 +1,202 @@
...
+ * @since TBD
+ */
+final class Preset_Selector {
+
+	/**
+	 * @var string The DTCG leaf key carrying a token's type.
+	 *
+	 * @since TBD
+	 */
+	private const TYPE_KEY = '$type';
```
d4mation: We have `$type` hardcoded in a handful of places now. I think we should add something to the Vocabulary namespace so that we have a single source of truth for it.

includes/resources/Design_Tokens/Foundation_Presets/Selector.php
```diff
@@ -0,0 +1,202 @@
...
+	 * @param string      $group  The preset group key, e.g. "typeScale" or "colorPalette".
+	 * @param string      $choice The preset slug within the group, e.g. "goldenRatio".
+	 * @param string|null $slug   The token set to write to. Defaults to the single v1 "default" set.
+	 *
+	 * @throws Unknown_Preset_Exception|RuntimeException When the group/choice does not exist, a preset
+	 *         token targets a dot-path the baseline has no $type for, or the overrides cannot be encoded.
+	 *
+	 * @return void
+	 */
+	public function apply( string $group, string $choice, ?string $slug = null ): void {
```
d4mation, line 81: If the same preset is chosen as the already selected one, this runs through all the logic again with no actual change. We should guard against that.

includes/resources/Design_Tokens/Foundation_Presets/Selector.php
```diff
@@ -0,0 +1,202 @@
...
+		$document = $this->current_overrides( $slug );
+
+		foreach ( $tokens as $dot_path => $value ) {
+			$leaf = [
+				self::TYPE_KEY             => $this->baseline_type( (string) $dot_path ),
+				Sentinels::get_value_key() => $value,
+			];
+
+			$this->set_leaf( $document, (string) $dot_path, $leaf );
+		}
```
d4mation, line 104: If we ever have a case where Preset A and Preset B have entirely exclusive Tokens at their own Paths, then switching from A -> B -> back to A could result in Tokens from B persisting after the swap. 

We may need to account for this, but that layering effect could be considered a feature 🤔

d4mation (commented):
One other thing I noticed that isn't explicitly part of this PR:

All other variant values are Aliases in the Baseline, but `transparent` is raw for Ghost Buttons. We may want to address that. But it doesn't need to be here.

includes/resources/Design_Tokens/Foundation_Presets/Catalog.php
dave-green-uk: I called this out to Claude myself, but it suggested the naming convention wasn't unique in the repo. I didn't really push back 😅

dave-green-uk: Done in cd76fb18 — renamed the class to `Catalog` (`Foundation_Presets\Catalog`). Namespace unchanged; updated the Provider binding, `Preset_Selector`, the helper, and the test (now `CatalogTest`).

includes/resources/Design_Tokens/functions.php
dave-green-uk: Reworded in cd76fb18 — dropped the "future" framing. The docblock now describes it in present tense: "the admin UI and MCP surface read it to render the picker."

includes/resources/Design_Tokens/Foundation_Presets/Preset_Selector.php
dave-green-uk: Agreed — added `Token_Type::get_type_key()` as the single source and swept the literal out of `Preset_Selector`, `Dtcg_Validator` (×2) and `Token_Resolver` (cd76fb18). The only `'$type'` left in the module is the vocab constant itself.

includes/resources/Design_Tokens/Foundation_Presets/Selector.php
dave-green-uk: Added a no-op guard in cd76fb18: if a selection nets no change versus the stored document, `apply()` returns before writing — no version bump, no `changed` action, no history entry. Covered by a new "re-selecting the current preset is a no-op" test.

includes/resources/Design_Tokens/Foundation_Presets/Selector.php
dave-green-uk: Good catch. Went with a clean swap rather than layering (cd76fb18): `apply()` now clears the group's whole footprint — `Catalog::group_paths()`, the union of every preset's paths in the group — before writing the chosen preset, so a preset's exclusive paths can't persist across a swap.

One trade-off worth flagging: this also means selecting a preset discards any *manual* override within that group's footprint, i.e. a preset selection is authoritative for its group. New test `testSelectingAPresetClearsAnotherPresetsExclusivePaths` covers it via a disjoint-path fake group (the shipped presets all share paths, so none trigger the leftover case today).

dave-green-uk:
Re: the `transparent` literal in the Ghost variant — agreed it's out of scope here; I've noted it for the variant data-model work (SOFT-3393), where the Ghost button's bindings get formalized.

Everything else is addressed in cd76fb18, kept as a **separate follow-up commit** (not squashed into the original) so the review changes are reviewable on their own.

includes/resources/Design_Tokens/Foundation_Presets/Catalog.php
d4mation: I think you may have forgotten to push the changes, because I cannot see that commit or any other change you've referenced😅

includes/resources/Design_Tokens/Foundation_Presets/Selector.php
d4mation: I think that makes sense personally. If I were using Preset A, customized some Tokens, and then switched to Preset B, I would expect it to do a full swap to Preset B and then I'd need to customize again. 

It could be annoying, but trying to do a partial merge could be confusing in other ways as the user.

includes/resources/Design_Tokens/Foundation_Presets/Selector.php
d4mation, line 1: Minor thing, but maybe it'd make sense to rename this to just `Selector` since we know we're already dealing with Presets.

d4mation (approved):

PR #1020, opened 2026-06-05
feat: Design Tokens write endpoints + validation (SOFT-3389)

d4mation:
🎫 https://linear.app/nexcess/issue/SOFT-3389/write-endpoints-validation

Adds the write surface to the Design Tokens `Documents_Controller`, alongside the existing read endpoints. Branches off the Read Endpoints branch since the writes live in that same controller.

## Endpoints

- **Bulk** on `/documents/{slug}`: `POST`/`PATCH` deep-merge, `PUT` replace, `DELETE` reset to baseline. `POST /documents` creates-or-merges the default set.
- **Single token** on `/documents/{slug}/tokens/{path}`: `POST`/`PUT` set, `DELETE` remove. The leaf is the bare request body; `$type` is inferred from the existing token, so the common case is a one-field body.

Verb semantics follow WordPress: `POST` is a partial update (merge), `PUT` is the full replace.

## Validation

Every write runs one pipeline before anything is stored:

1. DTCG grammar validation.
2. Dry-run resolver pass — rejects alias cycles and dangling aliases with HTTP 422.
3. Save through `Token_Store::save_document()`, which bumps the version and fires the change action.

A single-token write that targets a token group (or nests under a leaf) is rejected with a clear 422, instead of a confusing downstream dangling-alias error. The single-token `DELETE` runs this same pipeline, so removing a token that another override still aliases is rejected with 422 rather than persisting a dangling alias.

## Why `Document\Mutator`

A write needs three structural edits to the stored overrides document, and none had an existing home:

- merge a partial document into the stored overrides
- set a token leaf by dot-path
- remove a leaf by dot-path and prune emptied groups

The one nearby class, `Effective_Document`, can't be reused — for a write it does the opposite of what is needed:

- merges against the **baseline**, not the stored overrides
- **applies** the `$value:null` / `$disabled` sentinels instead of storing them
- strips `$extensions`

So writes get a small, pure helper (`Design_Tokens\Document\Mutator`) that keeps sentinels verbatim and preserves `$extensions`.

## Why `Effective_Document::build()` takes `$apply_disabled`

`Mutator` owns the structural edits, but a single-token write still reads the **effective** document (baseline merged with overrides) for two read-only checks:

- inferring a leaf's `$type` from the token already at that path, so the request body can omit it
- guarding that the path addresses a leaf, not a token group, and is not nested under a leaf

This is the flip side of why `Mutator` exists. `Mutator` is needed because `Effective_Document` is the wrong tool to *produce* what gets stored; but `Effective_Document` is still the right — and only — place to *read* the merged baseline-plus-overrides view those two checks need. The catch is that its rendering view **applies** the `$disabled` sentinel: it drops disabled tokens entirely, which is correct for what gets rendered but wrong for authoring. A token disabled by an override is still a real token you can re-enable, so at write time it must stay visible to read its `$type` and to tell a disabled group apart from a disabled leaf.

So `build()` takes an `$apply_disabled` flag rather than the write layer growing its own merge (which would reintroduce exactly the duplication `Mutator` was kept small to avoid):

- `true` (default): the Resolver's rendering view — disabled tokens are dropped.
- `false`: the write layer's inspection view — disabled tokens stay, keeping the baseline `$type` and group shape readable.

The split stays clean: `Mutator` decides what gets stored; `Effective_Document` remains the single place that merges baseline with overrides, now serving both rendering and author-time inspection of that merge.

### Checklist


includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
```diff
@@ -245,6 +408,232 @@ public function get_resolved( $request ) {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param mixed           $value   The submitted path.
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @param string          $key     The parameter name.
+	 *
+	 * @return true|WP_Error
+	 */
+	public function validate_token_path( $value, $request, $key ) {
```
dave-green-uk, line 620: Do we need to handle empty segments here?

dave-green-uk (approved):
LGTM. I left one non-blocking comment

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
d4mation: Fixed in [`f4edd13` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1020/commits/f4edd13a8c797ef81b44f0b4d921505cbf3d52ea) :)

pauloiankoski (approved):

PR #1021, opened 2026-06-08
feat: Variant data model and registration

dave-green-uk:
🎫  SOFT-3393

## Summary

This PR formalises the variant data model and its registration for the Design Tokens module (SOFT-3393).

A block's variants — and its default "preset" look — live in the DTCG document under $extensions.com.kadence.designTokens.variants.<block>. Each block declares a $default (the variant shown when nothing is selected) alongside a set of named variants, where every variant carries a human-readable label and a tokens map binding block properties to token aliases or literal values.

This PR adds the code-side structure behind that document shape: a way to declare which blocks accept variants and how each of their properties reaches output, plus a resolver that turns a variant's bindings into CSS-ready values. It's the foundation that the downstream tickets build on — native block styles (SOFT-3394), Kadence kbVariant (SOFT-3395), the block-preset $default (SOFT-3396), and the read/write endpoints (SOFT-3390). Block rendering, editor UI, and endpoints are intentionally out of scope here.

One nice outcome: the provisional baseline shape shipped back in SOFT-3377 needed no migration, because the model was built to match what was already there.

## Overview of changes

- Variant_Set — the registration that a block accepts variants, together with its per-property bindings. It holds structure only; the variant names, default, and values stay in the document, so there's a single source of truth and nothing to drift.
- Binding — describes how one block property reaches output. It can reference an existing token (reusing that token's projections), declare inline targets (kadence_slot, wp_preset, block_attr, css_var), or combine both.
- Token_Registry — gains register_variant_set(), for_block(), and effective_projections(), which merges a referenced token's projections with a binding's inline targets.
- Variant_Resolver — flattens a variant's tokens into a property => value map. Aliases resolve through the existing cached token resolver; literal values pass straight through. Also resolves a block's $default (its preset).
- Unknown_Variant_Exception — raised for an unknown block or variant, or a missing default.
- declarations.php — registers the Button variant set as the representative case for v1. The remaining blocks will be registered alongside their own per-block tickets.
- Tests — unit coverage for the model and resolver, plus a baseline guard that checks every shipped block: its $default must name a real variant, and each variant must have a label and a non-empty tokens map. A bad edit to baseline.json now fails CI rather than slipping through.

## Test plan for QA

This is data-model and registration plumbing with no user-facing surface yet, so verification is mostly automated:

- [ ] slic run wpunit - Resources/Design_Tokens/Registry passes
- [ ] slic run wpunit - Resources/Design_Tokens/Resolver/Variant_ResolverTest.php passes
- [ ] slic run wpunit - Resources/Design_Tokens/Registry/Baseline/Baseline_VariantsTest.php passes
- [ ] PHPStan and PHPCS come back clean
- [ ] Sanity check: on a site with no design-token authoring, blocks render exactly as before. This PR adds no projection or render path, so there should be no visible change in the front end or the editor.


.cspell.json
d4mation, line 1: There's some weird indentation introduced in this file. We should fix it.

tests/wpunit/Resources/Design_Tokens/Registry/Baseline/Baseline_VariantsTest.php
```diff
@@ -0,0 +1,121 @@
...
+ * `$default` names a variant that actually exists, and every named variant carries a label and a
+ * non-empty tokens map. Runtime writes are validated at the endpoint boundary (SOFT-3390); this guards
+ * the trusted, shipped document so a bad hand-edit to baseline.json fails CI instead of surfacing later.
+ */
+final class Baseline_VariantsTest extends TestCase {
+
+	/**
+	 * Absolute path to the shipped baseline document.
+	 */
+	private const BASELINE_PATH = KADENCE_BLOCKS_PATH . 'includes/resources/Design_Tokens/Registry/Baseline/baseline.json';
```
d4mation, line 18: Not a blocker, but I wonder if we should just add a method to `TestCase` to retrieve the global Baseline doc 🤔

includes/resources/Design_Tokens/Resolver/Variant_Resolver.php
```diff
@@ -0,0 +1,316 @@
...
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Resolver;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Contracts\Baseline_Document;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Exception\Unknown_Variant_Exception;
+use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Alias;
+
+/**
+ * Flattens a block variant's token bindings to CSS-ready values — the variant counterpart of the
+ * Token_Resolver, and the seam SOFT-3394/3395/3396/3390 build on.
```
d4mation: We should avoid ticket references in production code.

includes/resources/Design_Tokens/Resolver/Variant_Resolver.php
```diff
@@ -0,0 +1,316 @@
...
+ * (`{semantic.color.button-bg}`) is looked up in the Token_Resolver's already-resolved id map (so it
+ * reuses the existing cycle-safe, cached flattening); a literal passes straight through. The output is
+ * `property => value`; the projection *target* for each property comes from the Variant_Set's
+ * {@see \KadenceWP\KadenceBlocks\Design_Tokens\Registry\Binding} — value and target are kept separate
+ * so each downstream projector maps them its own way.
+ *
+ * v1 reads variant definitions from the shipped baseline. Merging store overrides into the variant
+ * section (so a user can re-skin a variant) is a follow-up that arrives with the write endpoints; the
+ * core Resolver's Effective_Document deliberately strips `$extensions`, so variants are resolved here
+ * rather than through that deep-merge.
```
d4mation: We're going to do this, so we don't need to say it'll come in a follow-up. Updating this could be missed when we make the change.

includes/resources/Design_Tokens/Resolver/Variant_Resolver.php
```diff
@@ -0,0 +1,316 @@
...
+	 * @return array<string, mixed>
+	 */
+	private function variant_tokens( string $block, string $variant ): array {
+		$block_variants = $this->block_variants( $block );
+
+		if ( ! isset( $block_variants[ $variant ] ) || ! is_array( $block_variants[ $variant ] ) ) {
+			throw Unknown_Variant_Exception::for_variant( $block, $variant );
+		}
+
+		$tokens = $block_variants[ $variant ][ self::TOKENS_KEY ] ?? [];
```
d4mation, line 251: What if a Variant exists but the Tokens Key is missing? I wonder if we should return an error in this case instead of an empty array as that just silently fails.

`effective_projections()` intentionally silently fails for a similar case so maybe this is fine. But we should document it clearly if it is expected.

tests/wpunit/Resources/Design_Tokens/Resolver/Variant_ResolverTest.php
```diff
@@ -0,0 +1,104 @@
...
+		$properties = $this->resolver->value_properties( self::BUTTON );
+
+		sort( $properties );
+		$this->assertSame(
+			[ 'button-bg', 'button-border', 'button-radius', 'button-text' ],
+			$properties
+		);
+	}
+
+	public function testTheShippedButtonSetIsConsistent(): void {
```
d4mation, line 79: Maybe we should also assert that the token ref binding resolves with `effective_projections()` to be extra safe. The silent failures that can happen during resolution could result in these tests not catching that case if we have a typo when retrieving via alias.

d4mation (commented):
Code overall looks good :) Just a few minor things.

.cspell.json
dave-green-uk: Fixed — normalised the whole `words` list to 8-space indentation and dropped the stray tab-indented entries.

tests/wpunit/Resources/Design_Tokens/Registry/Baseline/Baseline_VariantsTest.php
dave-green-uk: Added a `baseline_document()` helper to `TestCase`. Kept this test reading the file directly though — that's deliberate, to bypass the version-keyed object cache (otherwise a corrupted baseline gets served stale and the guard wouldn't catch it).

includes/resources/Design_Tokens/Resolver/Variant_Resolver.php
dave-green-uk: Removed — reworded to plain terms here, and cleared the same refs from the other new files in this PR (Binding, declarations).

includes/resources/Design_Tokens/Resolver/Variant_Resolver.php
dave-green-uk: Good call — dropped the follow-up framing; the docblock now just describes current behaviour.

includes/resources/Design_Tokens/Resolver/Variant_Resolver.php
dave-green-uk: Documented as intentional: a variant may legitimately set no values, so a missing/non-array `tokens` resolves to an empty map — only an undefined block/variant errors. Mirrors `effective_projections()`'s fail-soft.

tests/wpunit/Resources/Design_Tokens/Resolver/Variant_ResolverTest.php
dave-green-uk: Added an assertion that the `button-bg` token-ref resolves to non-empty projections via `effective_projections()`, so a typo'd alias can't slip through silently.

d4mation (approved):

PR #1022, opened 2026-06-08
feat: Block preset $default integration

dave-green-uk:
🎫  SOFT-3396

## Summary

This PR wires a block's preset — its default, on-brand look — into Kadence Blocks' rendering, building on the variant data model from SOFT-3393.

Every block can have a $default variant: the look it should show out of the box when no one has customised it. This PR makes that default flow into the block's attributes automatically, so a freshly dropped block is "locked" to its preset — yet a user editing an individual block still overrides it as normal.

It does this by reusing KB's existing per-block defaults machinery rather than inventing a parallel one. A new filter sits in the middle of how a block builds its defaults, letting the design-tokens module slot the preset in as a layer between the block's block.json defaults and the per-instance attributes:

```
block.json defaults   →  the generic out-of-box default (overlaid by)
token preset          →  the block's $default look[this PR] (overlaid by)
instance attributes   →  what the user set on this block (alwayswins)
```

This is the reusable mechanism. It's intentionally not hooked up to any real block yet — that happens per-block in SOFT-3406, which adds the bindings that map a token to an actual block attribute. So there's no visible change in the editor or front end from this PR alone.

## Overview of changes

- Block_Preset_Projector (new) — resolves a block's $default variant and maps each value onto the block attribute its binding names. No-ops safely when a block has no preset, when the document defines no default, or when token projection is switched off.
- Projectors\Provider (new) — registers the projector and hooks it into the render path.
- Core: class-kadence-blocks-abstract-block.php — adds one filter (kadence_blocks_block_default_attributes) inside get_attributes_with_defaults(). This is the only core change, and it's a no-op unless something hooks it.
- Binding::block_attr() — small accessor for the block attribute a binding targets.
- Tests — unit coverage for the projector (mapping, override precedence, the no-op/fail-closed paths) and the block_attr() accessor.
- Also: fixed a test-isolation bug in Token_ResolverTest (a cold-cache test now uses a fresh resolver so it can't be affected by another test warming the shared one).

## Reviewer Notes

- One open decision: the preset currently writes a resolved value (e.g. a hex colour) into the attribute. An alternative is to write a CSS-variable reference so presets stay live in the browser, consistent with how variants render — I've held off pending the CSS-variable backbone (SOFT-3382) and flagged it here for a call.

##Test plan (QA)

This is rendering plumbing with no user-facing surface yet, so verification is automated:

- [ ] slic run wpunit Resources/Design_Tokens/Projectors passes
- [ ] slic run wpunit Resources/Design_Tokens/Resolver/Token_ResolverTest.php passes
- [ ] Full Resources/Design_Tokens suite passes (one pre-existing dev-autoload error in the schema conformance test is environmental and green in CI)
- [ ] PHPStan + PHPCS clean
- [ ] Sanity check: on a normal site, blocks render exactly as before — this PR adds no projection or render output on its own, so there should be zero visible change in the editor or on the front end.


includes/resources/Design_Tokens/Projectors/Provider.php
```diff
@@ -0,0 +1,41 @@
...
+				}
+
+				/** @var Block_Preset_Projector $projector */
+				$projector = $this->container->get( Block_Preset_Projector::class );
+
+				return $projector->add_preset_defaults( $defaults, $block );
+			},
+			10,
+			2
+		);
```
d4mation: Why not use a Container Callback and do the early returns within it?

I wonder if we should wait until #1017 is merged so we have its structure to build on. @pauloiankoski made some architectural changes there which could help here.

Alternatively, we could point this PR to that one. I figure it should be merged soon regardless.

tests/wpunit/Resources/Design_Tokens/Projectors/Block_Preset_ProjectorTest.php
```diff
@@ -0,0 +1,129 @@
...
+	public function testThePresetWinsOverBlockJsonDefaultsButLeavesOthersIntact(): void {
+		$defaults = $this->projector( $this->button_set() )->add_preset_defaults(
+			[
+				'padding'    => '10px',          // untouched by the preset.
+				'background' => 'rebeccapurple', // overridden by the preset.
+			],
+			self::BUTTON
+		);
+
+		$this->assertSame( '10px', $defaults['padding'] );
```
d4mation: I don't think this technically tests anything? Wouldn't it write `10px` regardless of whether it was overwriting the value or not?

includes/blocks/class-kadence-blocks-abstract-block.php
```diff
@@ -414,7 +414,22 @@ public function get_attributes_with_defaults( $cache_key, $attributes, $cache =
...
+		/**
+		 * Filters a block's default attributes before they are merged with the instance's attributes.
+		 *
+		 * Lets a module contribute an extra layer of defaults (e.g. the Design Tokens block preset) that
+		 * sits above the block.json defaults yet below the instance attributes, so the merge below still
+		 * lets a per-instance value win.
+		 *
+		 * @since TBD
+		 *
+		 * @param array  $default_attributes The block's registration defaults, keyed by attribute name.
```
d4mation: We should document the array shape here.

includes/blocks/class-kadence-blocks-abstract-block.php
```diff
@@ -414,7 +414,22 @@ public function get_attributes_with_defaults( $cache_key, $attributes, $cache =
...
+		 * @param string $block_name         The full block name, e.g. "kadence/advancedbtn".
+		 */
+		$default_attributes = apply_filters( 'kadence_blocks_block_default_attributes', $default_attributes, 'kadence/' . $this->block_name );
```
d4mation: We should include `@return` to make it really clear. Technically the first param should always be the return value, but it can be helpful. This isn't a blocker as I don't think we have an official style related to this, but this is how we always handled it at LearnDash.

includes/blocks/class-kadence-blocks-abstract-block.php
```diff
@@ -414,7 +414,22 @@ public function get_attributes_with_defaults( $cache_key, $attributes, $cache =
...
+
+		$merged_attributes = $this->merge_attributes_with_defaults( $attributes, $default_attributes );
```
d4mation, line 441: What happens a third party uses the filter and `$default_attributes` isn't an array? We should protect against this.

includes/resources/Design_Tokens/Projectors/Provider.php
dave-green-uk: #1017 is merged now, so I folded this into the new `Projection/` architecture: the projector moved to `Projection/Block_Preset/{Projector,Provider}`, registered through `Projection\Provider`, and wired with `$this->container->callback( Projector::class, 'add_preset_defaults' )` to match the `Css_Var` sibling. Dropped the parallel `Projectors/` namespace. The non-array protection now lives at the filter's hook site (see the `get_attributes_with_defaults` thread).

tests/wpunit/Resources/Design_Tokens/Projectors/Block_Preset_ProjectorTest.php
dave-green-uk: Good framing catch — the override is actually proven by the `background` assertion just below (`rebeccapurple` -> `#3182CE`), but you're right the `padding` line alone doesn't show it. Made it explicit: the test now asserts overwrite (`background`), an additive binding (`color`), and the untouched survivor (`padding`), with a comment noting the `padding` assertion guards against a regression where the projector replaces instead of overlays.

includes/blocks/class-kadence-blocks-abstract-block.php
dave-green-uk: Done — documented the shape as `array<string, mixed>` (attribute name => default value).

includes/blocks/class-kadence-blocks-abstract-block.php
dave-green-uk: Done — added `@return array<string, mixed>` to the filter docblock.

includes/blocks/class-kadence-blocks-abstract-block.php
dave-green-uk: Done — the hook site now captures the filtered value separately and keeps KB's own defaults if a callback returns a non-array, so a misbehaving third-party filter can't corrupt the merge.

d4mation (approved):

tests/wpunit/Resources/Design_Tokens/Projectors/Block_Preset_ProjectorTest.php
d4mation: Ah, that makes sense. Thanks for adding the clarifying comments!

PR #1023, opened 2026-06-08
Fix breakout margin for nested row layout columns (SMTNC-798)

gilbert-hernandez:
## Summary

Fixes the breakout left/right behavior so that nested row layouts break out correctly. Previously, the CSS selector matched `.wp-block-kadence-column` descendants at any depth, which caused further-nested row layout columns to receive the breakout margin unintentionally.

## Changes

In `src/blocks/rowlayout/edit.js` (`KadenceRowLayout`), for the `breakoutLeft` case:

- Scoped the column selector to direct children using the child combinator (`>`) instead of a descendant selector, so only the immediate columns of the row are targeted — not columns inside deeper nested rows.
- Corrected the `max-width: 1289px` margin calculation to use `paddingLeftBreakout` instead of `paddingRightBreakout`, matching the left-side breakout intent.

## Testing

- Verify left/right breakout rows render with the correct negative margin.
- Verify nested row layouts no longer inherit the parent's breakout margin.


WayneRocha (approved):
Approved with additional commit [a464942](https://github.com/stellarwp/kadence-blocks/pull/1023/commits/a464942f7e538aef23a22f76a2c01baa8d3d3cf5) for cspell to [failing CI](https://github.com/stellarwp/kadence-blocks/actions/runs/27166168026/job/80193792874?pr=1023) to pass.

PR #1024, opened 2026-06-08
feat: variant read/write REST endpoints

d4mation:

Adds a `Variants_Controller` (`kb-design-tokens/v1/variants`) for reading and writing block variants per block: list registered variant blocks, per-block create/merge/replace/reset, per-variant delete, and `$default` get/set. Reads merge the shipped baseline with stored overrides through a new `Effective_Variants` reader, so reads reflect writes; only blocks with a registered variant set are addressable. Writes reuse the existing DTCG validation and resolver dry-run pipeline. Resolved/preview values stay out of scope here, owned by the projector tickets `SOFT-3394` / `SOFT-3395`. Also points `Variant_Resolver`'s DTCG keys at the shared `Extensions` vocabulary.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Registry/Token_Registry.php
+++ b/includes/resources/Design_Tokens/Registry/Token_Registry.php
@@ -151,6 +151,18 @@ final class Token_Registry {
 	}
 
 	/**
+	 * The block names that have a registered variant set, in registration order — what the REST variants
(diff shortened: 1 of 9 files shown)


d4mation:
Closing in favor of #1031 #1032 #1033 and #1034

The code is the same, just split into multiple PRs.

PR #1025, opened 2026-06-09
[DTM][1/3] Extract shared Wp_Preset_Target value object

pauloiankoski:
## Summary

Phase 1 of 3 for the Theme_Json projector (SOFT-3383). A behavior-preserving refactor that extracts the `(category, slug)` derivation out of `Css_Var\Css_Builder` into a shared `Projection\Wp_Preset_Target` value object.

This is the foundation for the theme.json projector: the theme.json preset `slug` **must** match the `--wp--preset--<category>--<slug>` segment the `Css_Var` bridge emits, or the editor swatch and the preset variable point at different names. Sharing one derivation guarantees the two projectors never drift.

- Adds `Projection\Wp_Preset_Target` — normalizes a token's `wp_preset` projection (bare-string category or `['category'=>…, 'slug'=>…]`) into a `(category, slug)` pair, or `null` to signal skip
- Refactors `Css_Var\Css_Builder::preset_block()` to consume it; removes the now-duplicated `preset_target()` / `slug_from_id()` private methods and the local `WP_PRESET` constant

No output change — the existing `Css_Var` builder + snapshot tests pass unchanged.

## Test plan

- [x] `slic run wpunit Resources/Design_Tokens/Projection` (45 tests green, incl. 8 new `Wp_Preset_Target` cases)
- [x] `composer phpcs` clean
- [x] `composer phpstan` clean
- [x] `npx cspell` clean on changed files

## Stacked PRs

1/3 — this PR → `SOFT-3383/theme-json-projector`
2/3 — `Theme_Json_Builder` → this branch
3/3 — `Theme_Json\Projector` + wiring → 2/3

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Wp_Preset_Target.php
+++ b/includes/resources/Design_Tokens/Projection/Wp_Preset_Target.php
@@ -31,7 +31,7 @@ final class Wp_Preset_Target {
 	 *
 	 * @var string
 	 */
-	public const PROJECTION = 'wp_preset';
+	private const PROJECTION = 'wp_preset';
 
 	/**
 	 * The preset category, e.g. "color", "font-family", "spacing", "shadow".
@@ -58,6 +58,17 @@ final class Wp_Preset_Target {
 	private function __construct( string $category, string $slug ) {
 		$this->category = $category;
 		$this->slug     = $slug;
+	}
+
+	/**
+	 * Get the projection key.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public static function get_projection_key(): string {
+		return self::PROJECTION;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


includes/resources/Design_Tokens/Projection/Wp_Preset_Target.php
```diff
@@ -0,0 +1,111 @@
...
+final class Wp_Preset_Target {
+
+	/**
+	 * The projection key a token declares to opt into preset projection.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const PROJECTION = 'wp_preset';
```
d4mation: We cannot use `public const` as it cannot be deprecated in the future if necessary.

Please change this to a `protected`/`private` constant and add a `public static` accessor instead. Something like `get_projection_key()` could work.

pauloiankoski: Good call. 0caa7f44ece928e777420b26241c96affcb8f332

d4mation (approved):

PR #1027, opened 2026-06-09
[DTM][3/3] Add Theme_Json projector and wire into theme.json filters

pauloiankoski:
## Summary

Phase 3 of 3 for the Theme_Json projector (SOFT-3383). Adds the WordPress-facing projector that injects token presets into the `theme.json` pipeline, plus its provider, and registers it in `Projection\Provider`.

- Adds `Theme_Json\Projector::inject()`, bound to **both** `wp_theme_json_data_default` and `wp_theme_json_data_theme`, mutating via `$theme_json->update_with([...])` — the same mechanism KB uses in `class-kadence-blocks-settings.php`
- **Bails fast** when projection is inactive (`Token_Registry::is_active()`) or no token declares a `wp_preset` projection
- Catches `RuntimeException` from the resolver (alias cycle / dangling alias from a raw DB write) so a corrupt stored document degrades gracefully rather than fatalling the request — mirrors `Css_Var\Projector`
- **Merges** token presets onto whatever the incoming theme.json already declares, so KB *adds to* — never replaces — the active theme's palette and KB's own `kadence_blocks_colors` injection; existing entries win on a slug collision
- Adds `Theme_Json\Provider` and registers it in `Projection\Provider`

### Note for reviewers — filter priority

Registered at priority **1000** so it runs *after* KB's existing `wp_theme_json_data_theme` injection (priority 999), letting `merge_existing()` see and preserve KB's `kadence_blocks_colors` palette. Worth a Site Editor smoke test to confirm ordering empirically before merge.

## Test plan

- [x] `slic run wpunit Resources/Design_Tokens/Projection` (64 tests green, incl. 6 new `Theme_Json\ProjectorTest` cases: happy path, inactive registry, early-bail, merge-preserves-existing, slug-collision, filter wiring)
- [x] `composer phpcs` / `composer phpstan` / `npx cspell` clean
- [ ] Manual: confirm tokens appear as presets in the Site Editor and that changing a token value updates them live

## Stacked PRs

Targets 2/3 (`SOFT-3383/phase-2-theme-json-builder`). Review/merge 1/3 → 2/3 → this in order.

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Theme_Json/Projector.php
+++ b/includes/resources/Design_Tokens/Projection/Theme_Json/Projector.php
@@ -1,0 +1,244 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection\Theme_Json;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Database\Token_Store;
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Token_Registry;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
+use RuntimeException;
+use WP_Theme_JSON_Data;
+
+/**
+ * Projects the resolved token set into the WordPress theme.json pipeline so tokens surface as native
+ * presets in the Site Editor (color palette, font families, spacing sizes, shadow presets).
+ *
+ * Hooks both wp_theme_json_data_default and wp_theme_json_data_theme and mutates via
+ * $theme_json->update_with([...]) — the same mechanism KB uses in class-kadence-blocks-settings.php.
+ * Every preset value is a CSS variable (var(--kb-token--*)), so changing a token updates all presets
+ * without rewriting theme.json. Gated on Token_Registry::is_active() and on at least one token
+ * declaring a wp_preset projection, so an inactive or preset-free registry is a no-op.
+ *
+ * Token preset lists are merged onto whatever the incoming theme.json already declares, so KB adds to
+ * — never replaces — the active theme's palette and KB's own kadence_blocks_colors injection.
+ *
+ * @since TBD
+ */
+final class Projector {
+
+	/**
+	 * @var Token_Registry
+	 */
+	private Token_Registry $registry;
+
+	/**
+	 * @var Token_Resolver
+	 */
+	private Token_Resolver $resolver;
+
+	/**
+	 * @var Token_Store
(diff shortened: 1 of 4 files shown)


includes/resources/Design_Tokens/Projection/Theme_Json/Provider.php
```diff
@@ -0,0 +1,31 @@
...
+	public function register(): void {
+		$this->container->singleton( Theme_Json_Builder::class );
+		$this->container->singleton( Projector::class );
+
+		// Both layers: default (so user/theme can still override) and theme (so presets read as
+		// theme-origin in the Site Editor). KB's existing palette injection runs at priority 999 on
+		// wp_theme_json_data_theme; we run at 1000 so our merge_existing() sees and preserves it.
+		$callback = $this->container->callback( Projector::class, 'inject' );
+		add_filter( 'wp_theme_json_data_default', $callback, 1000 );
+		add_filter( 'wp_theme_json_data_theme', $callback, 1000 );
```
d4mation: We should add a small buffer so someone else could choose to inject _just_ before us if possible. `1010` would be better for that purpose.

includes/resources/Design_Tokens/Projection/Theme_Json/Projector.php
```diff
@@ -0,0 +1,244 @@
...
+
+	/**
+	 * Read the value at a path inside a nested array, or null when any segment is missing.
+	 *
+	 * @since TBD
+	 *
+	 * @param array<string, mixed> $tree The tree to read from.
+	 * @param string[]             $path The path segments.
+	 *
+	 * @return mixed
```
d4mation: `mixed|null` may be nicer here to help directly hint that `null` is explicitly expected. But technically `mixed` _does_ cover it.

pauloiankoski: Resolved in https://github.com/stellarwp/kadence-blocks/pull/1027/commits/a23a4fdf3b090e77ece0c6fdcd3cb5ab3f49437c

includes/resources/Design_Tokens/Projection/Theme_Json/Provider.php
pauloiankoski: Resolved in https://github.com/stellarwp/kadence-blocks/pull/1027/commits/a23a4fdf3b090e77ece0c6fdcd3cb5ab3f49437c

d4mation (approved):

PR #1028, opened 2026-06-09
Kadence Concierge architecture plan

pramodjodhani:
🎫  #[Jira Ticket]

...

### Open Questions:

- [ ] We might we need to create a Result collection depending on how UI used


plans/concierge/Kadence-Concierge-Architecture-Brief.md
```diff
@@ -0,0 +1,360 @@
...
+├── Provider.php                        # Top-level; registers 3 sub-providers
+├── Contracts/
+│   ├── Check.php                       # Core interface all checks implement
+│   ├── Fixable.php                     # Optional: auto-repair capability
+│   └── Serializable.php                # REST response contract
+├── Check_Context.php                   # Value object: post_id, scope, params
+├── Check_Result.php                    # Immutable VO: ::pass() ::warn() ::fail()
+├── Check_Status.php                    # Enum: PASS | WARN | FAIL
+├── Check_Registry.php                  # Holds all registered checks
+├── Check_Runner.php                    # Executes checks, returns Report
```
d4mation, line 53: Maybe these should be in a `Concierge\Checks` namespace. Then you would have `Checks\Context`, `Checks\Result`, etc.

plans/concierge/Kadence-Concierge-Architecture-Brief.md
```diff
@@ -0,0 +1,360 @@
...
+
+## Directory Structure
+
+```
+includes/resources/Concierge/
+├── Provider.php                        # Top-level; registers 3 sub-providers
+├── Contracts/
+│   ├── Check.php                       # Core interface all checks implement
+│   ├── Fixable.php                     # Optional: auto-repair capability
+│   └── Serializable.php                # REST response contract
```
d4mation, line 48: Having the Contracts be "closer" to their use would be a bit better. So if you move the `Check_` classes to `Concierge\Checks`, you'd have `Concierge\Checks\Contracts\Check.php` (and the other Contracts may move there as well if they're related strictly to Checks).

plans/concierge/Kadence-Concierge-Architecture-Brief.md
```diff
@@ -0,0 +1,360 @@
...
+│   ├── Oversized_Images_Check.php
+│   ├── Missing_Alt_Text_Check.php
+│   └── Missing_Meta_Title_Check.php
+├── Site_Hygiene/
+│   ├── Provider.php
+│   ├── Missing_Favicon_Check.php
+│   ├── Default_Tagline_Check.php
+│   ├── Placeholder_Content_Check.php
+│   ├── Inactive_Plugins_Check.php
+│   └── Inactive_Themes_Check.php
```
d4mation, line 70: The namespaces both have their own Check classes. Maybe `Concierge\Pre_Publish\Provider.php` which then registers each check under `Concierge\Pre_Publish\Checks`. This way if other functionality is added to this namespace later, it is kept separate from the Check logic.

But you then may want to revisit my `Concierge\Checks` suggestion above. Maybe there's a better way to organize those?

plans/concierge/Kadence-Concierge-Architecture-Brief.md
```diff
@@ -0,0 +1,360 @@
...
+    $this->container->bind( Missing_Meta_Title_Check::class );
+
+    add_action( 'init', function() {
+        $registry = $this->container->make( Check_Registry::class );
+        $registry->add( $this->container->make( Broken_Links_Check::class ) );
+        $registry->add( $this->container->make( Unlinked_CTAs_Check::class ) );
+        $registry->add( $this->container->make( Oversized_Images_Check::class ) );
+        $registry->add( $this->container->make( Missing_Alt_Text_Check::class ) );
+        $registry->add( $this->container->make( Missing_Meta_Title_Check::class ) );
+    } );
```
d4mation, line 228: Do we want to just build each of these classes on `init`? Is the idea that they have a constructor which sets up their actions/filters?

It would be better to instead hook each class's methods using container callbacks. Then the objects are only built once they're needed instead of immediately at `init`. https://github.com/lucatume/di52#the-callback-method

plans/concierge/Kadence-Concierge-Architecture-Brief.md
```diff
@@ -0,0 +1,360 @@
...
+│   ├── Fixable.php                     # Optional: auto-repair capability
+│   └── Serializable.php                # REST response contract
+├── Check_Context.php                   # Value object: post_id, scope, params
+├── Check_Result.php                    # Immutable VO: ::pass() ::warn() ::fail()
+├── Check_Status.php                    # Enum: PASS | WARN | FAIL
+├── Check_Registry.php                  # Holds all registered checks
+├── Check_Runner.php                    # Executes checks, returns Report
+├── Report.php                          # Aggregated results collection
+├── Exceptions/
+│   └── Unknown_Check_Exception.php
```
d4mation, line 56: This looks Check-specific as well and likely would go under `Concierge\Checks\Exceptions`

plans/concierge/Kadence-Concierge-Architecture-Brief.md
pramodjodhani: Right. We dont need to make these classes at init. I will remove this.

dave-green-uk:
Closing as this is now handled over in stellarwp/kadence-concierge

PR #1030, opened 2026-06-09
feat: Added Kadence kbVariant selectable variants

dave-green-uk:
🎫  [SOFT-3395](https://linear.app/nexcess/issue/SOFT-3395/kadence-kbvariant-variants)

## Summary

Kadence blocks can have multiple "looks" defined as design-token variants (e.g. a Button's primary / secondary / ghost). Today the token system can resolve those variants but nothing lets a block actually use one. This PR adds the mechanism that lets a block instance switch to a named variant.

The approach reuses the cascade rather than rewriting any block's render path: an editor filter tags a block with a kbVariant--<name> class, and a new projector emits scoped CSS that re-points the --global-paletteN custom properties those blocks already render with. So selecting a variant re-skins the block on-brand, and an unselected block keeps its default look. Because every variant value is an alias into the same token graph, changing a brand color updates the default and every variant together.

This delivers the generic mechanism + a tested pilot. It does not wire any shipped block yet — that's per-block work (SOFT-3406, which this unblocks).

### PHP — the projector ("Projector B")
- Projection/Variant/Css_Builder — builds, per (block, variant), a scoped rule (.wp-block-<block>.kb-variant--<name> {--global-paletteN: var(--kb-token--variant--…) }) plus the global variant-var definitions. Per-(block, variant) scoping means two blocks sharing a variant name never collide. Cached on the store version, like the existing token CSS.
- Projection/Variant/Projector + Provider — enqueues that CSS on the front end and in the editor, gated on the registry being active, failing soft so a bad token set never breaks a page.
- Token_Registry::variant_blocks() — small accessor so the projector can enumerate variant-bearing blocks.
- Shared helpers Projection/Scope and Projection/Editor_Page, now used by both the Css_Var and Variant projectors (removed duplicated scope constant + editor-page check).

### JS — the editor plumbing (src/early-filters.js)
- Adds a kbVariant attribute to any block that opts in via a kbVariant block-support flag (mirrors the existing kbMetadata pattern).
- Outputs the kb-variant--<name> class on the saved markup and in the editor canvas, with the slug sanitized to match the projector's selector.

### Automated
- [ ] slic run wpunit Resources/Design_Tokens/Projection → green
  (51 tests)
- [ ] slic run snapshot Resources/Design_Tokens/Projection → green
- [ ] PHPCS + PHPStan clean on changed files; lint-js + cspell clean

### Manual — no-regression smoke (primary risk, since the mechanism is dormant until a block opts in)
- [ ] With Design Tokens active, load the editor and front end of a page with several Kadence blocks → blocks render exactly as before; palette colors and token CSS-vars unchanged.
- [ ] Deactivate the token set (fail-closed) → blocks fall back to normal KB behavior; no errors.

### Manual — mechanism check (temporarily opt a block in to exercise the path)
- [ ] Temporarily add kbVariant to a block's supports (e.g. Button) and register a variant set with a kadence_slot binding, then:
- [ ] Confirm a kbVariant attribute exists on the block and persists across save/reload.
- [ ] Set it to a variant name → the block wrapper gets kb-variant--<name> in the editor and the saved/front-end markup.
- [ ] Confirm the page CSS contains .wp-block-….kb-variant--<name>{ --global-paletteN: var(--kb-token--variant--…) } and the block visually re-skins to the variant's colors.
- [ ] Make a per-instance color edit on that block → the instance edit still wins over the variant.
- [ ] Revert the temporary opt-in before merge.


includes/resources/Design_Tokens/Projection/Scope.php
```diff
@@ -0,0 +1,25 @@
...
+final class Scope {
+
+	/**
+	 * The root scope selector shared by the projection CSS builders.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const ROOT = ':root,:root:where(.kb-tokens)';
```
d4mation: We cannot use `public const`. Please swap this for `protected`  or `private` with a `public static` accessor.

includes/resources/Design_Tokens/Projection/Css_Var/Css_Builder.php
```diff
@@ -29,14 +30,13 @@
...
-	 * The root scope for both declaration blocks. :where() contributes zero specificity, so neither
-	 * selector weighs more than a bare :root. Never !important.
+	 * The root scope for both declaration blocks, defined once in {@see Scope::ROOT}.
 	 *
 	 * @since TBD
 	 *
 	 * @var string
 	 */
-	public const SCOPE = ':root,:root:where(.kb-tokens)';
+	public const SCOPE = Scope::ROOT;
```
d4mation: It looks like this was missed in a previous review. We should avoid using `public const` here as well.

It may mean that places where we're calling `self::SCOPE` in this class will need to be replaced with calling the public static method directly.

src/early-filters.js
```diff
@@ -68,6 +68,93 @@ function convertArrayTitleToString(arr) {
...
+ * Add the kbVariant attribute to any block that opts in via the `kbVariant` block support.
+ *
+ * The attribute holds the slug of the selected design-token variant (e.g. "ghost"); empty means the
+ * block keeps its $default look (the block preset). The scoped CSS that re-skins the block for a
+ * selected variant is emitted server-side by the Design Tokens variant projector.
+ *
+ * @param {Object} settings The block settings.
+ * @return {Object} The block settings with the kbVariant attribute added.
+ */
+export function blockVariantAttribute(settings) {
```
d4mation, line 84: It would be worthwhile to include `@since TBD` in JS functions as well for easy tracking in the future.

```js
* @param ...
*
* @since TBD
*
* @return ...
```

includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
```diff
@@ -0,0 +1,270 @@
...
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Projection\Scope;
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Binding;
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Css_Var;
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Token_Registry;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Variant_Resolver;
+use RuntimeException;
+
+/**
+ * Builds the scoped CSS for selectable block variants — "Projector B" for Kadence blocks.
```
d4mation: "Projector B" is specific to the EngDoc and doesn't make sense in the context of the codebase as a whole.

includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
```diff
@@ -0,0 +1,270 @@
...
+	 * token's slot and an inline binding declares its own; both reach the same --global-paletteN.
+	 *
+	 * @since TBD
+	 *
+	 * @param Binding $binding The variant binding.
+	 *
+	 * @return string|null The slot ("palette3"), or null.
+	 */
+	private function palette_slot( Binding $binding ): ?string {
+		$slot = $this->registry->effective_projections( $binding )[ Binding::KADENCE_SLOT ] ?? null;
```
d4mation: It looks like `Binding::KADENCE_SLOT` must be a `public const`? I guess that was missed too. We should swap that for a non-public const.

includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
```diff
@@ -0,0 +1,270 @@
...
+	 * @since TBD
+	 *
+	 * @param string $block    The block name.
+	 * @param string $variant  The variant slug.
+	 * @param string $property The block property.
+	 *
+	 * @return string
+	 */
+	private function variant_var( string $block, string $variant, string $property ): string {
+		return Css_Var::PREFIX . self::VARIANT_SEGMENT
```
d4mation: `Css_Var::prefix` shouldn't be a `public const`

src/early-filters.js
```diff
@@ -68,6 +68,93 @@ function convertArrayTitleToString(arr) {
...
+
+	return props;
+}
+addFilter('blocks.getSaveContent.extraProps', 'kadence/kb-variant-save-class', blockVariantSaveClass);
+
+/**
+ * Mirror the kb-variant--<name> class onto the block in the editor canvas, so a selected variant previews
+ * live with the same scoped overrides the front end uses.
+ */
+const withBlockVariantClass = createHigherOrderComponent((BlockListBlock) => {
```
d4mation, line 148: Adding a `@since TBD` here can be a good idea as well.

includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
```diff
@@ -0,0 +1,270 @@
...
+	 *
+	 * @param string $value The raw CSS value.
+	 *
+	 * @return string
+	 */
+	private function sanitize_value( string $value ): string {
+		$value = preg_replace( '/[\x00-\x1F\x7F]/', '', $value ) ?? '';
+
+		return str_replace( [ '{', '}', ';', '<', '>' ], '', $value );
+	}
```
d4mation: This method is identical to the one in `Css_Var\Css_Builder`. Maybe we should create a Trait or an Abstract Class that these share?

includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
```diff
@@ -0,0 +1,270 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string $segment The raw segment.
+	 *
+	 * @return string
+	 */
+	private function ident( string $segment ): string {
+		return (string) preg_replace( '/[^A-Za-z0-9_-]+/', '-', $segment );
+	}
```
d4mation, line 255: `ident` is easy to misread (or mistype) as `indent`. We should make this clearer at a glance as to its purpose by renaming it to `sanitize_identifier`.

includes/resources/Design_Tokens/Projection/Variant/Provider.php
```diff
@@ -0,0 +1,31 @@
...
+	 * @inheritDoc
+	 *
+	 * @since TBD
+	 */
+	public function register(): void {
+		$this->container->singleton( Css_Builder::class );
+		$this->container->singleton( Projector::class );
+
+		// Front end: append after the token vars (KB enqueues the handle at 90; the Css_Var projector at 100).
+		add_action( 'wp_enqueue_scripts', $this->container->callback( Projector::class, 'enqueue_front_end' ), 100 );
```
d4mation: Do we want this to load after `Css_Var/Css_Builder`? We may want to hook this a bit later, such as `110` then.

includes/resources/Design_Tokens/Projection/Variant/Provider.php
```diff
@@ -0,0 +1,31 @@
...
+
+		// Editor: append at admin_init priority 5, after the editor-styles handle is registered.
+		add_action( 'admin_init', $this->container->callback( Projector::class, 'enqueue_editor' ), 5 );
```
d4mation: Same possible concern here in relation to `Css_Var/Css_Builder` priority.

includes/resources/Design_Tokens/Projection/Editor_Page.php
```diff
@@ -0,0 +1,29 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public static function is_block_editor(): bool {
+		global $pagenow;
+
+		return in_array( $pagenow, [ 'post.php', 'post-new.php', 'site-editor.php', 'widgets.php' ], true );
+	}
```
d4mation: This technically would return `true` if Classic Editor was enabled for `post.php` and `post-new.php`. We should improve this check.

includes/resources/Design_Tokens/Projection/Editor_Page.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection;
+
+/**
+ * Detects whether the current admin request is a block-editor page, shared by the projectors that append
+ * their CSS to the editor style handles.
+ *
+ * @since TBD
+ */
+final class Editor_Page {
```
d4mation: Maybe we should move this to the Utilities namespace and rename the class to `Location`?

Then we could do things like `if ( Location::is_block_editor() )` which reads nicely.

src/early-filters.js
```diff
@@ -68,6 +68,93 @@ function convertArrayTitleToString(arr) {
...
+addFilter('blocks.registerBlockType', 'kadence/kb-variant-attribute', blockVariantAttribute);
+
+/**
+ * The class a selected variant outputs, or an empty string when none is selected.
+ *
+ * @param {string} kbVariant The selected variant slug.
+ * @return {string} The `kb-variant--<slug>` class, or an empty string.
+ */
+function kbVariantClassName(kbVariant) {
+	// Mirror the projector's PHP ident() sanitizer, so the output class always matches the scoped selector
```
d4mation: We'll want to update the method reference here to match the new name. Add it could be worth including the class name too in case we create a similar method name in the future elsewhere.

includes/resources/Design_Tokens/Projection/Scope.php
dave-green-uk: Done — `ROOT` is private now with a public static `Scope::root()` accessor.

src/early-filters.js
dave-green-uk: Added `@since TBD` to the variant JSDoc blocks.

includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
dave-green-uk: Done — `PREFIX` is private with a public static `Css_Var::get_prefix()`.

includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
dave-green-uk: Renamed to `sanitize_identifier()` (and updated the JS comment that referenced it).

includes/resources/Design_Tokens/Projection/Variant/Provider.php
dave-green-uk: Bumped the front-end enqueue to 110 so it lands after the Css_Var projector (100).

includes/resources/Design_Tokens/Projection/Editor_Page.php
dave-green-uk: Improved — post.php/post-new.php now defer to `use_block_editor_for_post(_type)()`, so a classic-editor post type correctly returns false. Added `LocationTest` coverage.

includes/resources/Design_Tokens/Projection/Editor_Page.php
dave-green-uk: Done — moved to `Utils\Location`; reads as `Location::is_block_editor()` now.

includes/resources/Design_Tokens/Projection/Css_Var/Css_Builder.php
dave-green-uk: Done — dropped the `SCOPE` const entirely (it only aliased `Scope::ROOT`); both builders call `Scope::root()` directly.

includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
dave-green-uk: Reworded — dropped the "Projector B" phrasing here and in the Projector docblock.

includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
dave-green-uk: Done — `KADENCE_SLOT` is private with a `get_kadence_slot_key()` accessor. Swept the other four key-constants to `private const` too while here.

src/early-filters.js
dave-green-uk: Added.

includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
dave-green-uk: Extracted into a shared `Sanitizes_Css_Value` trait used by both builders.

includes/resources/Design_Tokens/Projection/Variant/Provider.php
dave-green-uk: Bumped the editor `admin_init` enqueue to 6, after the Css_Var projector at 5.

src/early-filters.js
dave-green-uk: Updated the comment to reference `Css_Builder::sanitize_identifier()`.

includes/resources/Design_Tokens/Projection/Variant/Provider.php
d4mation: `6` doesn't leave any wiggle room for other code to run between these implementations. Would `10` work?

includes/resources/Design_Tokens/Projection/Sanitizes_Css_Value.php
```diff
@@ -0,0 +1,31 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection;
```
d4mation: A `Traits` namespace here would be nicer so it wasn't at the same root level as the other classes.

d4mation (approved):

includes/resources/Design_Tokens/Utils/Location.php
d4mation, line 1: Looks like we have some silly cspell issues. Could you add them to the global ignore quickly, please? Then this PR is 100% good.

```
includes/resources/Design_Tokens/Utils/Location.php:36:64 - Unknown word (unslash)
includes/resources/Design_Tokens/Utils/Location.php:43:40 - Unknown word (absint)
includes/resources/Design_Tokens/Utils/Location.php:43:51 - Unknown word (unslash)
```

d4mation (changes requested):

PR #1031, opened 2026-06-09
Variant endpoints (1/4): vocabulary, registry, and resolver groundwork

d4mation:

First of four stacked PRs that split the variant read/write endpoints into reviewable pieces. Based on `feature/design-tokens-module`.

Groundwork shared by the variants reader and controller:

- Points `Variant_Resolver`'s DTCG keys at the shared `Extensions` vocabulary instead of hardcoded constants.
- Adds the `Extensions` accessor for the variants section.
- Adds `Token_Registry::variant_blocks()` to list the blocks that have a registered variant set.

No endpoint or behavior change ships here; this is the vocabulary and registry surface the later PRs build on.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Resolver/Variant_Resolver.php
+++ b/includes/resources/Design_Tokens/Resolver/Variant_Resolver.php
@@ -279,11 +279,7 @@ final class Variant_Resolver {
 	private function variants_section(): array {
 		$node = $this->baseline->document();
 
-		$path = [
-			Extensions::get_extensions_key(),
-			Extensions::get_namespace(),
-			Extensions::get_section_variants(),
-		];
+		$path = Extensions::get_variants_path();
 
 		foreach ( $path as $key ) {
 			if ( ! is_array( $node ) || ! isset( $node[ $key ] ) ) {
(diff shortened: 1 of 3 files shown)


pauloiankoski (approved):

includes/resources/Design_Tokens/Resolver/Variant_Resolver.php
```diff
@@ -306,7 +279,13 @@ private function block_variants( string $block ): array {
 	private function variants_section(): array {
 		$node = $this->baseline->document();
 
-		foreach ( self::VARIANTS_PATH as $key ) {
+		$path = [
+			Extensions::get_extensions_key(),
+			Extensions::get_namespace(),
+			Extensions::get_section_variants(),
+		];
```
pauloiankoski: In PR #1032, we are rebuilding the same array again. Maybe we could extract it into `Extensions::get_variants_path()` so it can be reused.

d4mation: [`c6a5698` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1031/commits/c6a5698d94e091bfceb17a4bfb8aaea55b5cd0fb)

pauloiankoski (approved):

PR #1033, opened 2026-06-09
Variant endpoints (3/4): variants REST read endpoints

d4mation:

Third of four stacked PRs. Based on `feature/SOFT-3390/variants-2-effective-reader`.

Adds `Variants_Controller` with the read surface of the variants resource (`kb-design-tokens/v1/variants`):

- List the blocks that have a registered variant set, each with its default and named variant slugs.
- Read a single block's baseline-merged variant set.
- Read a block's `$default`.

Reads merge the shipped baseline with stored overrides through `Effective_Variants`. Only blocks with a registered variant set are addressable; others are a 404. The block name carries a slash, so it is routed as two path segments and reassembled. The controller is registered on the v1 REST provider. Write endpoints land in the fourth PR.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
@@ -209,11 +209,12 @@ final class Variants_Controller extends Controller {
 	 * @return WP_REST_Response
 	 */
 	public function get_items( $request ) {
-		$slug   = Token_Store::default_slug();
-		$blocks = [];
+		$slug    = Token_Store::default_slug();
+		$section = $this->variants->section( $slug );
+		$blocks  = [];
 
 		foreach ( $this->registry->variant_blocks() as $block ) {
-			$node = $this->variants->block( $block, $slug ) ?? [];
+			$node = isset( $section[ $block ] ) && is_array( $section[ $block ] ) ? $section[ $block ] : [];
 
 			$blocks[] = [
 				'block'   => $block,
--- a/includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
@@ -1,0 +1,502 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Rest\V1;
+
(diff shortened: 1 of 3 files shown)


includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
```diff
@@ -0,0 +1,502 @@
...
+
+			$blocks[] = [
+				'block'   => $block,
+				'default' => $this->default_of( $node ),
+				'names'   => $this->variant_names( $node ),
+			];
+		}
+
+		return new WP_REST_Response( [ 'blocks' => $blocks ], WP_Http::OK );
+	}
```
pauloiankoski, line 227: We are calling `$this->variants->block($block, $slug)` inside the loop. Each call to `block()` ends up calling `section()`, which in turn calls `store->get_document()`. That means we're doing one database query per block.

With only a single registered block in v1, the impact is negligible, but this pattern could silently degrade performance as more blocks are added.

One possible improvement would be to hoist a single `$this->variants->section($slug)` call before the loop and then access `$section[$block] ?? null` directly inside the loop.

tests/wpunit/Resources/Design_Tokens/Rest/V1/VariantsControllerTest.php
pauloiankoski, line 1: Just like we did in `DocumentsControllerTest`, I think it's worth adding tests for `get_items_permissions_check()` and `get_item_permissions_check()` with both authenticated and unauthenticated users.

includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
d4mation: [`85ed39d` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1033/commits/85ed39d8b5dbe8cfd9f6470cbd9a529cfeca8088)

tests/wpunit/Resources/Design_Tokens/Rest/V1/VariantsControllerTest.php
d4mation: [`85ed39d` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1033/commits/85ed39d8b5dbe8cfd9f6470cbd9a529cfeca8088)

pauloiankoski (approved):

PR #1034, opened 2026-06-09
Variant endpoints (4/4): variants REST write endpoints

d4mation:

Fourth of four stacked PRs. Based on `feature/SOFT-3390/variants-3-read-endpoints`.

Adds the write surface to `Variants_Controller`:

- POST creates or merges a single variant, leaving siblings and `$default` intact.
- PUT replaces a block's whole variant set, dropping any variant absent from the body.
- DELETE on a block resets it to baseline; DELETE on a variant removes that one variant.
- A `default` sub-route gets and sets the block's default.

Writes assemble a partial overrides document for the addressed variants node, deep-merge it onto the stored set, then run the shared pipeline: DTCG grammar validation (HTTP 422), a dry-run Resolver pass that rejects alias cycles and dangling aliases (HTTP 422), then a single `Token_Store::save_document()` that bumps the version. Empty variant slugs are rejected, and a `$default` that does not name a present variant is rejected before commit. Resolved/preview values stay out of scope here, owned by the projector tickets `SOFT-3394` / `SOFT-3395`.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
@@ -760,7 +760,21 @@ final class Variants_Controller extends Controller {
 			);
 		}
 
-		return $this->persist( (string) wp_json_encode( $candidate ), $block, $status );
+		$encoded = wp_json_encode( $candidate );
+
+		// Guard the encode: a false return cast to "" would clear the whole set on persist instead of storing it.
+		if ( $encoded === false ) {
+			return new WP_Error(
+				'rest_design_tokens_save_failed',
+				__( 'The design token set could not be encoded.', 'kadence-blocks' ),
+				[
+					'status' => WP_Http::INTERNAL_SERVER_ERROR,
+					'block'  => $block,
+				]
+			);
+		}
+
+		return $this->persist( $encoded, $block, $status );
 	}
 
 	/**
@@ -982,7 +996,7 @@ final class Variants_Controller extends Controller {
 	 * @return array<string, mixed>
 	 */
 	private function unset_block( array $document, string $block ): array {
-		return $this->unset_path( $document, array_merge( $this->variants_path(), [ $block ] ) );
+		return $this->mutator->remove_by_keys( $document, array_merge( $this->variants_path(), [ $block ] ) );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
```diff
@@ -354,6 +708,88 @@ public function get_collection_params(): array {
...
+				'rest_design_tokens_unresolvable',
+				$e->getMessage(),
+				[
+					'status' => WP_Http::UNPROCESSABLE_ENTITY,
+					'block'  => $block,
+				]
+			);
+		}
+
+		return $this->persist( (string) wp_json_encode( $candidate ), $block, $status );
```
pauloiankoski: `(string) wp_json_encode( $candidate )` converts a `false` encoding failure into an empty string (`""`), which `persist()` then saves to the store. This can silently clear all overrides instead of returning an error.

The same pattern exists in `Documents_Controller:817`, so this is already established in the codebase. However, since this PR introduces its own write path, it seems like a good opportunity to fix the issue consistently in both places.

includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
```diff
@@ -400,6 +923,136 @@ private function prepare_item( string $block ): array {
...
+		}
+
+		$node[ $head ] = $this->unset_path( $node[ $head ], $keys );
+
+		if ( $node[ $head ] === [] ) {
+			unset( $node[ $head ] );
+		}
+
+		return $node;
+	}
```
pauloiankoski: This method seems to be a duplicate of `Mutator::remove_segments()`. Maybe we could introduce a public `Mutator::remove_by_keys( array $document, array $keys ): array` method by renaming the current `remove_segments()` implementation and making it public.

Then `Variants_Controller::unset_path()` could delegate to it, and `remove_segments()` could become a thin wrapper around `remove_by_keys()`. That would avoid duplicating the same logic in multiple places.

includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
```diff
@@ -400,6 +923,136 @@ private function prepare_item( string $block ): array {
...
+	 *
+	 * @return string[]
+	 */
+	private function variants_path(): array {
+		return [
+			Extensions::get_extensions_key(),
+			Extensions::get_namespace(),
+			Extensions::get_section_variants(),
+		];
+	}
```
pauloiankoski, line 1026: See https://github.com/stellarwp/kadence-blocks/pull/1031#pullrequestreview-4462379665

includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
```diff
@@ -459,6 +1112,25 @@ private function named_variants( array $node ): array {
...
+		$raw = $this->store->get_document( Token_Store::default_slug() );
+
+		if ( $raw === '' ) {
+			return [];
+		}
+
+		$decoded = json_decode( $raw, true );
+
+		return is_array( $decoded ) ? $decoded : [];
+	}
```
pauloiankoski, line 1099: This looks similar to `Effective_Variants::stored_document()`. Maybe we could add a public `Effective_Variants::raw( string $slug = 'default' ): array` method (or `decoded_document()`, if that name fits better) that returns the decoded overrides.

Then we could remove the private implementation from `Variants_Controller` and have both callers reuse the same logic. As a follow-up, `Effective_Variants::section()` could call `$this->raw( $slug )` instead of maintaining its own private path to the decoded document.

d4mation: [`61a937e` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1034/commits/61a937e4f4c76bbb8b0af0150d19ad893249aded)

includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
d4mation: [`61a937e` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1034/commits/61a937e4f4c76bbb8b0af0150d19ad893249aded)

includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
d4mation: [`61a937e` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1034/commits/61a937e4f4c76bbb8b0af0150d19ad893249aded)

includes/resources/Design_Tokens/Rest/V1/Variants_Controller.php
d4mation: [`61a937e` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1034/commits/61a937e4f4c76bbb8b0af0150d19ad893249aded)

pauloiankoski (approved):

PR #1036, opened 2026-06-09
[DTM][2/3] SOFT-3384: pure Palette_Builder

pauloiankoski:
**Stacked PR 2 of 3** — base: `SOFT-3384/phase-1-palette-slot`. Review only this phase's diff.

Adds the pure `Palette_Builder`: turns the registry + resolved tokens into a `slug => {color,name}` map, then merges that onto a decoded existing option. No WordPress calls — the projector (Phase 3) owns all I/O. Two deliberately different merge rules:

- `kadence_blocks_colors` (KB's own) — overwrite color + name of a claimed slug, append a claimed slug not present, preserve every other entry and the `override` flag.
- `kadence_global_palette` (the theme's) — overwrite **only** color of a claimed slug; never append, never touch names.

Includes unit tests for every merge rule plus a committed snapshot of the merged `kadence_blocks_colors` payload.

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Builder.php
+++ b/includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Builder.php
@@ -10,11 +10,17 @@ use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Resolved_Tokens;
 /**
  * Pure builder for the two palette options the Kadence_Option projector writes.
  *
- * Produces a slot => resolved-hex map from every token that declares a palette kadence_slot, then
- * MERGES that map onto a decoded existing option payload — overwriting only the color of token-claimed
- * slots, keyed by slug, and preserving every other entry, key and flag (e.g. kadence_blocks_colors'
- * "override"). No WordPress calls, no globals, no I/O: the projector owns get_option/update_option and
- * hands decoded arrays in and takes decoded arrays out.
+ * Produces a slug => { color, name } map from every token that claims a palette kadence_slot and
+ * resolves to a value, then merges that map onto a decoded existing option. No WordPress calls, no
+ * globals, no I/O: the projector owns get_option/update_option and hands decoded arrays in and out, so
+ * every merge rule here is unit-testable without the database.
+ *
+ * The two options have deliberately different merge rules:
+ *
+ *   - kadence_blocks_colors (KB's own) — overwrite color AND name of a claimed slug, append a claimed
+ *     slug not yet present, and preserve every other entry plus the "override" flag.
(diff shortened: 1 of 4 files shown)


includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Builder.php
```diff
@@ -0,0 +1,221 @@
...
+	 * { color, name, slug } — exactly what KB's consumers read.
+	 *
+	 * @since TBD
+	 *
+	 * @param array<string, mixed>                              $existing Decoded option, or [] when absent.
+	 * @param array<string, array{color: string, name: string}> $entries  slug => {color,name} from entries().
+	 *
+	 * @return array<string, mixed> The merged payload (re-encode and store).
+	 */
+	public function merge_kb_colors( array $existing, array $entries ): array {
```
d4mation: Since this is a public method, I think we should avoid using the `kb` shorthand in it. 

`merge_kadence_blocks_colors`

or 

`merge_plugin_colors` (This is more vague, but `merge_theme_colors` also doesn't say _what_ theme so it matches)

would be better I think.

includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Builder.php
```diff
@@ -0,0 +1,221 @@
...
+	 *
+	 * @param array<string, mixed>                              $existing Decoded option (caller guarantees it exists).
+	 * @param array<string, array{color: string, name: string}> $entries  slug => {color,name} from entries().
+	 *
+	 * @return array<string, mixed> The merged payload (re-encode and store).
+	 */
+	public function merge_theme_palette( array $existing, array $entries ): array {
+		if ( ! isset( $existing['palette'] ) || ! is_array( $existing['palette'] ) ) {
+			return $existing; // Unexpected shape — leave the theme's option exactly as found.
+		}
```
d4mation: Couldn't we re-use `$this->pallete_list()` here?

If it returns an empty array, then we know that either `$existing` is malformed or empty. So we could just return `$existing` in that case.

includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Builder.php
pauloiankoski: Renamed in 8df0da8b96798afe09a5701b8a7a6682d4cc2fd1

includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Builder.php
pauloiankoski: Refactored in 8df0da8b96798afe09a5701b8a7a6682d4cc2fd1

d4mation (approved):

PR #1037, opened 2026-06-09
[DTM][3/3] SOFT-3384: Kadence_Option projector + provider wiring

pauloiankoski:
**Stacked PR 3 of 3** — base: `SOFT-3384/phase-2-palette-builder`. Review only this phase's diff.

The write-time projector and its provider — the only layer that touches WordPress options and hooks.

- Syncs `kadence_blocks_colors` **always**; syncs the Kadence theme's `kadence_global_palette` **only when it already exists** (sentinel probe) and **never creates it**.
- `reconcile()` runs once per request on a late `init` pass and on `kadence_blocks_design_tokens_changed`, so a token write syncs immediately and a later theme switch is caught next request.
- A `{store-version}:{theme-present}` marker short-circuits the no-change case before any resolve; gated on `Token_Registry::is_active()` (fail-closed) with a fail-open catch around resolution.
- Both options read every request are written **autoloaded** (the marker and `kadence_blocks_colors`); the theme option is updated with `null` autoload (it always already exists).

Full `ProjectorTest` coverage incl. fail-closed, fail-open, idempotency, theme-switch, and the autoloaded-marker assertion.

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
+++ b/includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
@@ -1,0 +1,234 @@
+<?php declare( strict_types=1 );
+// cspell:ignore palette .
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection\Kadence_Option;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Database\Token_Store;
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Token_Registry;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
+use RuntimeException;
+
+/**
+ * Syncs resolved palette-token values into two stored options so pre-existing code paths reflect tokens.
+ *
+ *   - kadence_blocks_colors    — KB's own palette option. Synced ALWAYS (any active theme), so KB's
+ *                                editor palette UI and its theme.json / editor-settings injection track
+ *                                tokens with no change to those code paths.
+ *   - kadence_global_palette   — the Kadence theme's option. Synced ONLY when it already exists, NEVER
+ *                                created, so non-Kadence-theme sites are left untouched.
+ *
+ * reconcile() is wired to a once-per-request boot pass AND to kadence_blocks_design_tokens_changed, so a
+ * token write syncs immediately and a later theme switch (which newly exposes kadence_global_palette) is
+ * caught on the next request. A version+theme marker short-circuits the common no-change case; update_option
+ * is itself a no-op when the encoded value is unchanged, the correctness backstop. Gated on
+ * Token_Registry::is_active() (fail-closed) and a fail-open catch around resolution.
+ *
+ * @since TBD
+ */
+final class Projector {
+
+	/**
+	 * KB's own palette option key.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const KB_COLORS_OPTION = 'kadence_blocks_colors';
+
+	/**
+	 * The Kadence theme's palette option key.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const THEME_PALETTE_OPTION = 'kadence_global_palette';
+
+	/**
+	 * Marker option storing the last-synced "{store-version}:{theme-present}" signature, so a request
+	 * where nothing changed skips resolution entirely.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const SYNC_MARKER_OPTION = 'kadence_blocks_design_tokens_palette_sync';
+
+	/**
+	 * Sentinel for the existence probe. A value a real option is overwhelmingly unlikely to equal, so
+	 * get_option($key, SENTINEL) === SENTINEL reliably means "absent".
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const ABSENT = "\0kb-design-tokens-absent\0";
+
+	private Token_Registry $registry;
+	private Token_Resolver $resolver;
+	private Token_Store $store;
+	private Palette_Builder $builder;
+
+	/**
+	 * Guards against re-running the boot pass more than once per request (the action may also fire).
+	 *
+	 * @var bool
+	 */
+	private bool $reconciled_this_request = false;
+
+	public function __construct(
+		Token_Registry $registry,
+		Token_Resolver $resolver,
+		Token_Store $store,
+		Palette_Builder $builder
+	) {
+		$this->registry = $registry;
+		$this->resolver = $resolver;
+		$this->store    = $store;
+		$this->builder  = $builder;
+	}
+
+	/**
+	 * Once-per-request reconcile. Bound to the boot pass (low priority on init) and may be re-entered by
+	 * on_tokens_changed(); the per-request guard keeps it to one pass unless a write forces a re-sync.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function reconcile(): void {
+		if ( $this->reconciled_this_request ) {
+			return;
+		}
+		$this->reconciled_this_request = true;
+
+		$this->sync();
+	}
+
+	/**
+	 * Force a re-sync after a token write, regardless of the per-request guard, so the new values land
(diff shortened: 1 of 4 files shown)


includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
```diff
@@ -0,0 +1,246 @@
...
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const ABSENT = "\0kb-design-tokens-absent\0";
+
+	private Token_Registry $registry;
+	private Token_Resolver $resolver;
+	private Token_Store $store;
+	private Palette_Builder $builder;
```
d4mation, line 92: These need variable docs.

includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
```diff
@@ -0,0 +1,246 @@
...
+
+	/**
+	 * Sentinel for the existence probe. A value a real option is overwhelmingly unlikely to equal, so
+	 * get_option($key, SENTINEL) === SENTINEL reliably means "absent".
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const ABSENT = "\0kb-design-tokens-absent\0";
```
d4mation: Does this need the `\0` before and after? Took me a couple reads through the code to understand what you're doing with this, but I don't know if the `\0` is necessary?

includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
```diff
@@ -0,0 +1,246 @@
...
+	 *
+	 * @param array<string, array{color: string, name: string}> $entries
+	 *
+	 * @return void
+	 */
+	private function sync_theme_palette( array $entries ): void {
+		$raw = get_option( self::THEME_PALETTE_OPTION, self::ABSENT );
+		if ( $raw === self::ABSENT ) {
+			return; // raced away between probe and here; never create.
+		}
```
d4mation: Couldn't we use `$this->theme_palette_exists()` here? It would mean an extra read, but it is weird that we aren't using it. I think the performance hit is worth it to remove duplicated logic.

includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
```diff
@@ -0,0 +1,246 @@
...
+	 * @return array<string, mixed>
+	 */
+	private function decode( string $raw ): array {
+		if ( $raw === '' ) {
+			return [];
+		}
+		$decoded = json_decode( $raw, true );
+
+		return is_array( $decoded ) ? $decoded : [];
+	}
```
d4mation, line 266: Do we really need this helper? It looks like you're using a ternary when calling it every time regardless. Maybe we could just protect against the type that is passed into `json_decode()` a bit better and avoid this helper.

Alternatively, we could put that ternary guard directly into this method:

```php
private function decode( $raw ): array {
    if ( ! is_string( $raw ) || $raw === '' ) {
        return [];
    }
    $decoded = json_decode( $raw, true );
    return is_array( $decoded ) ? $decoded : [];
}
```

includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
```diff
@@ -0,0 +1,246 @@
...
+	 *
+	 * @return void
+	 */
+	private function sync(): void {
+		if ( ! $this->registry->is_active() || ! $this->builder->has_palette_tokens() ) {
+			return;
+		}
+
+		$theme_present = $this->theme_palette_exists();
+		$signature     = $this->store->get_version() . ':' . ( $theme_present ? '1' : '0' );
```
d4mation: This signature would technically go stale if the shipped Baseline updates on plugin update, wouldn't it? That means this sync wouldn't occur again unless the user were to update their colors _after_ the plugin update.

Something like `KADENCE_BLOCKS_VERSION . ':' $this->store->get_version() . ':' ( $theme_present ? '1' : '0' )` would avoid that by also syncing on plugin update.

tests/wpunit/Resources/Design_Tokens/Projection/Kadence_Option/ProjectorTest.php
d4mation, line 1: Few test cases that are missing:

1. The empty-entries branch (`$entries === []` -> advance marker, write nothing) isn't covered by a test. If the marker write inside the `if` were to move then it wouldn't get flagged with a failing test right now.
2. `testSecondReconcileWithUnchangedSignaturePerformsNoWrite` proves the value is unchanged, but `update_option()`'s no-op-on-unchanged makes it pass even if the short-circuit were deleted. It doesn't prove resolution was skipped right now.
3. The resolver-exception test never pre-seeds the options, so it proves "nothing created," not the actual fail-open guarantee "existing values preserved."

includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
```diff
@@ -0,0 +1,246 @@
...
+	private function sync_kb_colors( array $entries ): void {
+		$raw     = get_option( self::KB_COLORS_OPTION, '' );
+		$decoded = $this->decode( is_string( $raw ) ? $raw : '' );
+		$merged  = $this->builder->merge_kb_colors( $decoded, $entries );
+
+		// Autoloaded: this option is read on every front-end request — load_color_palette() on
+		// after_setup_theme and load_color_palette_theme_json() on wp_theme_json_data_theme both
+		// get_option() it. On an existing option update_option ignores this flag (preserving KB's own
+		// setting, normally autoloaded); it only takes effect when WE create the option first.
+		update_option( self::KB_COLORS_OPTION, (string) wp_json_encode( $merged ), true );
```
d4mation: This is not 100% accurate, but the underlying code is good.

Suggested updated here:

```php
// Pass true to ensure this option is autoloaded: it is read on every front-end request
// (load_color_palette() on after_setup_theme, load_color_palette_theme_json() on
// wp_theme_json_data_theme). update_option only writes when the value changes, and re-asserts
// autoload='on' each time it does -- which is the state we want.
```

Suggested update for `sync_theme_palette`:

```
// Pass null so update_option preserves the theme's existing autoload setting rather than
// forcing it. (It only writes at all when the value changes.)
```

includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
d4mation: I figure it is to further ensure this value couldn't exist in the DB, but would it _ever_ actually exist? I don't think it would.

d4mation:
@pauloiankoski Can you clarify which commits are the new ones? When you force push, the timestamps get messed up and it becomes harder to track for reviews 😅 

Edit: I resolved the ones that appear to have been addressed. The others are still unaddressed from what I can see.

image

includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
pauloiankoski: Would it make sense to simplify `theme_palette_exists()` to:

```php
return get_option( self::THEME_PALETTE_OPTION ) !== false;
```

`kadence_global_palette` is always persisted as a JSON-encoded string, so when the option exists its value will never be `false`. Conversely, `get_option()` returns `false` when the option row does not exist (never `''` or `null`). Given those guarantees, this check accurately expresses the intent while being simpler and more idiomatic.

pauloiankoski: Refactored in 84054ef510396d3e3fd309d11187dc79de4d6e67

includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
pauloiankoski: Refactored in 21e35f37a84500a5605ce2eb276a633ae7f4a76d

pauloiankoski:
> @pauloiankoski Can you clarify which commits are the new ones? When you force push, the timestamps get messed up and it becomes harder to track for reviews 😅

@d4mation I'll avoid rebasing the stacked branches and will merge one branch into the other instead. That should preserve the existing commit timestamps and avoid rewriting history.

d4mation (approved):

PR #1038, opened 2026-06-09
feat: add native block variants via register_block_style (SOFT-3394)

dave-green-uk:
🎫 [SOFT-3394](https://linear.app/nexcess/issue/SOFT-3394/native-block-variants-register-block-style)

## Summary

Native (non-Kadence) blocks — `core/button` and the like — can now carry **token-driven variants**, using WordPress's own block-style system rather than a custom attribute. Pick a style in the editor's block-styles switcher and the block re-skins on-brand; an unstyled block keeps its baseline (`$default`) look. Because every variant value is an alias into the same token graph, changing a brand colour updates the default and every variant together.

This is the **native sibling of SOFT-3395** (Kadence kbVariant variants): same idea, but native blocks get the selection UI, the attribute, and the `is-style-kb-<name>` class for free from WordPress — so this slice is **PHP-only, no JS/build**.

Sole owner of native-block variants (the per-block tickets 3401–3406 are all `kadence/*`), so it ships a real `core/button` pilot rather than only a synthetic one.

## Changes

**New `Projection/Block_Style/` module**
- `Registrar` — `register_block_style($block, 'kb-<variant>')` on `init`; WordPress shows the switcher and adds `is-style-kb-<variant>`. Labels come from the document (`Variant_Resolver::label()`), falling back to a title-cased slug.
- `Css_Builder` — emits, per (block, variant), a scoped rule that retargets the `--wp--preset--<cat>--<slug>` vars the block consumes: `.wp-block-button.is-style-kb-ghost{ --wp--preset--color--button-bg: var(--kb-token--variant--core-button--ghost--button-bg) }`, plus the global variant-var definitions.
- `Default_Styles` — builds the theme.json `styles.blocks.<block>` baseline that points each block's properties at their token-backed presets, so an unstyled block is on-brand and consumes the preset the variants then retarget.
- `Projector` — appends the CSS to KB's front-end/editor handles and injects the `$default` into `wp_theme_json_data_*`.
- `Style` — single source for the `register_block_style` name ↔ CSS selector so they can't drift.
- `Provider` — wires it; registered in `Projection/Provider`.

**Registry + data**
- `Token_Registry::variant_blocks()` and `Variant_Resolver::label()`.
- `core/button` variant set in `declarations.php` (reuses the existing `semantic.color.button-*` tokens, which already declare `wp_preset`) and `baseline.json` `variants.core/button` (primary/secondary/ghost). No new tokens — the baseline guard stays satisfied.

## Test plan

**Automated**
- [ ] `slic run wpunit Resources/Design_Tokens` → green (456 tests)
- [ ] `slic run snapshot Resources/Design_Tokens` → green
- [ ] PHPCS + PHPStan + cspell clean on changed files

**Manual (the bit unit tests can't prove — needs a browser)**
- [ ] Drop a **core/button**; it renders on-brand by default (consumes the token preset).
- [ ] In the block-styles switcher, **Primary / Secondary / Ghost** appear; picking one re-skins the button (e.g. Ghost = transparent bg). This relies on standard WP/CSS behaviour — core/button's colour applies to `.wp-block-button__link`, the `is-style-kb-*` class lands on `.wp-block-button`, and the wrapper's custom-property override **inherits** down to the link — so it's worth eyeballing.
- [ ] A per-instance colour edit still wins over the selected style.
- [ ] Front end matches the editor.
- [ ] Deactivate the token set (fail-closed) → core/button falls back to its native styling, no errors.

## Notes for reviewers
- **Independent of the SOFT-3395 PR** — different blocks (`core/*` vs `kadence/*`), target var (`--wp--preset--*` vs `--global-paletteN`), and class scheme (`is-style-kb-*` vs `kb-variant--*`).
- `Token_Registry::variant_blocks()` is the same one-line helper 3395 adds — trivial overlap whenever the second merges. The shared `Scope`/`Editor_Page` helpers from 3395 aren't on this base, so `Block_Style` keeps a local root-scope const + its own editor-page check (consolidation candidate once both land).
- `$default` uses `styles.blocks.<block>` (read of §10's "styles.variations baseline" — it's the *default* look, applied when unstyled, matching §4a "locked to the baseline preset until overridden").
- `Default_Styles::PROPERTY_PATHS` is a small property→theme.json-path map; fine for the pilot, grows per native block.


includes/resources/Design_Tokens/Projection/Block_Editor_Request.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection;
+
+/**
+ * Shared block-editor request detection for the projectors that append editor inline styles at
+ * admin_init priority 5, before get_current_screen() is available.
+ *
+ * @since TBD
+ */
+trait Block_Editor_Request {
```
d4mation: It looks like this was part of https://github.com/stellarwp/kadence-blocks/pull/1030 and has been renamed and refactored in that PR. Maybe it makes more sense to merge that one and sync this branch with the changes before reviewing it?

dave-green-uk: Good call. I merged #1030 in and synced this branch with it: dropped the duplicate `Block_Editor_Request` trait in favour of #1030's `Location::is_block_editor()`, and switched the native-block `Css_Builder` to its `Css_Var::get_prefix()` accessor (the const is private now). Both projectors here use `Location::is_block_editor()`. Projection/Utils/snapshot suites are green.

includes/resources/Design_Tokens/Registry/Baseline/baseline.json
```diff
@@ -450,6 +450,30 @@
...
+						"label": "Secondary",
+						"tokens": {
+							"button-bg": "{primitive.color.brand.secondary}",
+							"button-text": "{primitive.color.neutral.0}"
+						}
+					},
+					"ghost": {
+						"label": "Ghost",
+						"tokens": {
+							"button-bg": "transparent",
```
d4mation: We discussed at one point that we should have this be an Alias. Maybe now is a good time to do it? https://github.com/stellarwp/kadence-blocks/pull/1019#pullrequestreview-4436988514

includes/resources/Design_Tokens/Projection/Block_Style/Css_Builder.php
```diff
@@ -0,0 +1,293 @@
...
+	/**
+	 * Scope for the global variant-var definitions. Matches the token backbone's scope (see
+	 * Css_Var\Css_Builder::SCOPE) so the variant vars live everywhere KB prints variables; :where() adds
+	 * no specificity.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const ROOT_SCOPE = ':root,:root:where(.kb-tokens)';
```
d4mation: This should probably use `Scope::root()` instead, right?

includes/resources/Design_Tokens/Projection/Block_Style/Css_Builder.php
```diff
@@ -0,0 +1,293 @@
...
+	 *
+	 * @param string $value The raw CSS value.
+	 *
+	 * @return string
+	 */
+	private function sanitize_value( string $value ): string {
+		$value = preg_replace( '/[\x00-\x1F\x7F]/', '', $value ) ?? '';
+
+		return str_replace( [ '{', '}', ';', '<', '>' ], '', $value );
+	}
```
d4mation: We have a `Sanitizes_Css_Value` Trait for this now.

includes/resources/Design_Tokens/Projection/Block_Style/Style.php
```diff
@@ -0,0 +1,90 @@
...
+	public const NAME_PREFIX = 'kb-';
+
+	/**
+	 * The selector class WordPress outputs for a kb- style name (is-style- + the name).
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const CLASS_PREFIX = 'is-style-kb-';
```
d4mation: We've got some `public const`s here. We need to switch these to `protected`/`private`. If we need to reference them directly in a Test for whatever reason, Reflection can be used to grab them.

includes/resources/Design_Tokens/Projection/Block_Style/Provider.php
```diff
@@ -0,0 +1,40 @@
...
+
+		// Scoped variant CSS, appended after the token vars (Css_Var enqueues at 100).
+		add_action( 'wp_enqueue_scripts', $this->container->callback( Projector::class, 'enqueue_front_end' ), 100 );
+		add_action( 'admin_init', $this->container->callback( Projector::class, 'enqueue_editor' ), 5 );
+
+		// $default baseline into theme.json. Same hooks and priority as the Theme_Json projector, so the
+		// styles merge alongside its preset settings.
+		$inject = $this->container->callback( Projector::class, 'inject' );
+		add_filter( 'wp_theme_json_data_default', $inject, 1010 );
+		add_filter( 'wp_theme_json_data_theme', $inject, 1010 );
```
d4mation, line 40: We should consider the priority order of these hooks compared to the Variant Provider. It runs deliberately after other things and this probably needs to as well? If it does, we need to increment by `10` so that there's room for 3rd parties to inject between different things.

includes/resources/Design_Tokens/Projection/Block_Style/Css_Builder.php
```diff
@@ -0,0 +1,293 @@
...
+use KadenceWP\KadenceBlocks\Design_Tokens\Projection\Css_Var\Wp_Preset_Var;
+use KadenceWP\KadenceBlocks\Design_Tokens\Projection\Wp_Preset_Target;
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Binding;
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Css_Var;
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Token_Registry;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Variant_Resolver;
+use RuntimeException;
+
+/**
+ * Builds the scoped CSS for native block-style variants — "Projector B" for non-Kadence blocks.
```
d4mation: "Projector B" only makes sense in the context of the EngDoc and should be removed.

includes/resources/Design_Tokens/Projection/Block_Style/Style.php
```diff
@@ -0,0 +1,90 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string $segment The raw segment.
+	 *
+	 * @return string
+	 */
+	public static function ident( string $segment ): string {
+		return (string) preg_replace( '/[^A-Za-z0-9_-]+/', '-', $segment );
+	}
```
d4mation: We should switch this to the same thing as suggested for https://github.com/stellarwp/kadence-blocks/pull/1030#discussion_r3388823755

Also, it looks like it may be an identical method? If so, we should move it to a Trait. Maybe there should be a generic "CSS Transforms"-type trait to pull in this and the Sanitization function.

includes/resources/Design_Tokens/Projection/Block_Editor_Request.php
dave-green-uk: Agreed — #1030 is now merged and this branch has been synced, so this file is gone. Block-editor page detection moved to `Location::is_block_editor()`. Resolved.

includes/resources/Design_Tokens/Registry/Baseline/baseline.json
dave-green-uk: Done in 217ed6d4 — added a `primitive.color.transparent` token and aliased `core/button` ghost's `button-bg` to `{primitive.color.transparent}`. Left `kadence/advancedbtn`'s literal as-is since it came in via #1030 and is out of scope for this PR.

includes/resources/Design_Tokens/Projection/Block_Style/Css_Builder.php
dave-green-uk: Done in 217ed6d4 — dropped the local `ROOT_SCOPE` const and switched to `Scope::root()`.

includes/resources/Design_Tokens/Projection/Block_Style/Css_Builder.php
dave-green-uk: Done in 217ed6d4 — removed the local sanitizer and mixed in the `Sanitizes_Css_Value` trait.

includes/resources/Design_Tokens/Projection/Block_Style/Style.php
dave-green-uk: Done in 217ed6d4 — `NAME_PREFIX` and `CLASS_PREFIX` are now `private`.

includes/resources/Design_Tokens/Projection/Block_Style/Provider.php
dave-green-uk: Done in 217ed6d4 — the enqueue hooks now run after the base token vars and the kbVariant overrides, stepped by 10 (front end 120, editor 20). Left the `theme.json` filters at 1010, since that priority is deliberately aligned with the Theme_Json projector so the styles merge alongside its preset settings.

includes/resources/Design_Tokens/Projection/Block_Style/Css_Builder.php
dave-green-uk: Done in 217ed6d4 — removed the "Projector B" phrasing from the docblock.

includes/resources/Design_Tokens/Projection/Block_Style/Style.php
dave-green-uk: Done in 217ed6d4 — extracted into a shared `Sanitizes_Css_Identifier` trait (sibling to `Sanitizes_Css_Value`). `Style` and both `Css_Builder`s use it, so the duplicate is gone. Made it a `static` trait method so the static `Style` utility can pull it in too.

d4mation (approved):

PR #1040, opened 2026-06-09
[DTM][1/3] SOFT-3385: variant-set structure accessors for the admin UI feed

pauloiankoski:
**Stack:** phase 1 of 3 for [SOFT-3385](https://linear.app/nexcess/issue/SOFT-3385) (admin UI schema feed). Targets the base branch; phase 2 and 3 stack on top.

Adds the two pure, structure-only accessors the feed builder needs to enumerate and serialize registered variant sets — mirroring the existing `Token_Registry::to_ui_schema()` for tokens:

- `Token_Registry::variant_sets()` — lists registered sets keyed by block (the registry had only the single-block `for_block()`).
- `Variant_Set::to_ui_array()` — serializes per-property bindings (token ref + inline projections). Structure only; variant names/default/values remain document data read through `Variant_Resolver`.

No behavior change to existing consumers. Unit tests added to `Token_RegistryTest` and `Variant_SetTest`.

### DIFF
--- a/includes/resources/Design_Tokens/Registry/Token_Registry.php
+++ b/includes/resources/Design_Tokens/Registry/Token_Registry.php
@@ -151,6 +151,18 @@ final class Token_Registry {
 	}
 
 	/**
+	 * All registered variant sets, keyed by block name in registration order. The admin UI feed
+	 * (SOFT-3385) iterates this to render per-block variant editors; mirrors all() for tokens.
+	 *
+	 * @since TBD
+	 *
+	 * @return array<string, Variant_Set>
+	 */
+	public function variant_sets(): array {
+		return $this->variant_sets;
+	}
+
+	/**
 	 * The effective projection targets for a binding: a token reference contributes the referenced
 	 * token's projections (so a variant reuses the variable the base property already feeds), and the
 	 * binding's inline targets are merged on top, supplementing or overriding them (e.g. adding a
--- a/includes/resources/Design_Tokens/Registry/Token_Registry.php
+++ b/includes/resources/Design_Tokens/Registry/Token_Registry.php
@@ -152,7 +152,7 @@ final class Token_Registry {
 
 	/**
 	 * All registered variant sets, keyed by block name in registration order. The admin UI feed
-	 * (SOFT-3385) iterates this to render per-block variant editors; mirrors all() for tokens.
+	 * iterates this to render per-block variant editors; mirrors all() for tokens.
 	 *
 	 * @since TBD
 	 *
@@ -202,7 +202,7 @@ final class Token_Registry {
 	}
 
 	/**
-	 * The schema the admin React app consumes (SOFT-3385). A token registered in PHP appears in the
+	 * The schema the admin React app consumes. A token registered in PHP appears in the
 	 * UI with no JS change. Values are NOT included here — the resolver supplies current values
 	 * separately; this is structure only.
 	 *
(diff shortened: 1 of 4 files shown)


tests/wpunit/Resources/Design_Tokens/Registry/Token_RegistryTest.php
```diff
@@ -1,5 +1,5 @@
 <?php declare( strict_types=1 );
-// cspell:ignore advancedbtn .
+// cspell:ignore advancedbtn advancedheading .
```
d4mation: I had added `advancedbtn` to this local ignore quickly before, but I think we may be better off instead adding all block names to the global ignore list. They aren't going to ever change I'm sure.

includes/resources/Design_Tokens/Registry/Token_Registry.php
```diff
@@ -150,6 +150,18 @@ public function for_block( string $block ): ?Variant_Set {
 		return $this->variant_sets[ $block ] ?? null;
 	}
 
+	/**
+	 * All registered variant sets, keyed by block name in registration order. The admin UI feed
+	 * (SOFT-3385) iterates this to render per-block variant editors; mirrors all() for tokens.
```
d4mation: We should avoid ticket references in production code.

includes/resources/Design_Tokens/Registry/Variant_Set.php
```diff
@@ -110,6 +110,29 @@ public function consistency( array $value_properties ): array {
 		];
 	}
 
+	/**
+	 * Structure-only serialization for the admin UI feed (SOFT-3385): the bound properties and, per
+	 * property, the token reference and inline projection targets. Values, variant names and the default
+	 * are NOT here — those are document data read through the Variant_Resolver. Mirrors
+	 * {@see Token_Registry::to_ui_schema()}.
```
d4mation: We should avoid ticket references in production code.

includes/resources/Design_Tokens/Registry/Variant_Set.php
```diff
@@ -110,6 +110,29 @@ public function consistency( array $value_properties ): array {
...
+	 *
+	 * @since TBD
+	 *
+	 * @return array{bindings: array<string, array{token: string|null, projections: array<string, mixed>}>}
+	 */
+	public function to_ui_array(): array {
```
d4mation: Why `to_ui_array()` here but `to_ui_schema()` in `Token_Registry`? If the purpose is the same they should probably have the same method name.

tests/wpunit/Resources/Design_Tokens/Registry/Token_RegistryTest.php
pauloiankoski: Removed the local ignore in 816f9d80246103b38ba352051e1c12954532e8ef. `advancedbtn` and `advancedheading` are already in `.cspell.json`.

includes/resources/Design_Tokens/Registry/Token_Registry.php
pauloiankoski: Removed the ticket reference in 32c2cb25e548d54220c90054bea90e194db4958f.

includes/resources/Design_Tokens/Registry/Variant_Set.php
pauloiankoski: Removed the ticket reference in 4c692a4314e221519519402a6b95fc8356fdf95c.

includes/resources/Design_Tokens/Registry/Variant_Set.php
pauloiankoski: Renamed to `to_ui_schema()` for consistency with `Token_Registry` in c4f2a411d4360afcbad4b3fd5bd550cc6372f9dd.

d4mation (approved):

PR #1041, opened 2026-06-09
[DTM][2/3] SOFT-3385: pure Feed_Builder + Variant_Feed

pauloiankoski:
**Stack:** phase 2 of 3 for [SOFT-3385](https://linear.app/nexcess/issue/SOFT-3385). Targets phase 1.

The data-assembly layer for the admin UI schema feed, split pure-vs-impure like the SOFT-3383 `Json_Builder`/`Projector` split:

- `Admin\Feed_Builder` — **pure** (no WP calls): shapes the final `window.kadenceDesignTokens` payload from the registry's `to_ui_schema()` plus the resolved values, variants, REST descriptor and version handed in. Applies the fail-closed gate (inactive registry ⇒ empty `active:false` payload) and the `resolved` flag.
- `Admin\Variant_Feed` — joins the registry's variant structure with `Variant_Resolver`'s names/default/resolved values, failing soft per block/variant.

Tests: `Feed_BuilderTest`, `Variant_FeedTest`, and a `Feed_Builder_SnapshotTest` that freezes the payload contract SOFT-3407 will consume (and guards "register a token ⇒ it reaches the feed").


includes/resources/Design_Tokens/Admin/Feed_Builder.php
```diff
@@ -0,0 +1,66 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Admin;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Token_Registry;
+
+/**
+ * Pure assembler for the admin UI schema feed — the `window.kadenceDesignTokens` payload the dashboard
+ * React app reads (SOFT-3385).
```
d4mation: We should avoid ticket references in production code.

includes/resources/Design_Tokens/Admin/Feed_Builder.php
```diff
@@ -0,0 +1,66 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Admin;
```
d4mation: Maybe `Feed_Builder` and `Variant_Feed` should be in an `Admin\Feed` or similar namespace? Then we could have `Feed\Builder` and `Feed\Variants`.

includes/resources/Design_Tokens/Admin/Feed_Builder.php
pauloiankoski: Removed reference in d4e5e0d38df953a4deb9b3c7e2dbcefa5045a603

includes/resources/Design_Tokens/Admin/Feed_Builder.php
pauloiankoski: Refactored in 42562acfbf64a3d26364cd7c869c644946b113a6

d4mation (approved):

PR #1042, opened 2026-06-09
[DTM][3/3] SOFT-3385: localize the schema feed onto the admin dashboard

pauloiankoski:
**Stack:** phase 3 of 3 for [SOFT-3385](https://linear.app/nexcess/issue/SOFT-3385). Targets phase 2.

The WordPress-touching layer that lands the feed on the dashboard bundle, so a token registered in PHP appears in the UI with no JS change:

- `Admin\Localizer` — on `admin_head` (after the dashboard's `admin_print_styles-{page}` enqueue, before the footer print), guarded on `wp_script_is('admin-kadence-home','enqueued')` so it runs only where the bundle loads. Resolves current values fail-open (alias-cycle/dangling-alias ⇒ `resolved:false`, structure intact), gathers REST root + `wp_rest` nonce, and attaches `window.kadenceDesignTokens` via `wp_add_inline_script` + `wp_json_encode` (preserves booleans/types — `wp_localize_script` would stringify them).
- `Admin\Provider` registered in `Design_Tokens_Provider`.
- Plus a follow-up commit exposing the v1 REST namespace through `Controller::namespace_v1()` (static method, not a public const) so the Localizer references the exact namespace the routes register under with no drift.

Localized global is `window.kadenceDesignTokens`; `version` is the store's string hash. `LocalizerTest` covers the happy path, off-page no-op, fail-closed, fail-open, scalar type fidelity, and provider wiring.


includes/resources/Design_Tokens/Admin/Localizer.php
```diff
@@ -0,0 +1,162 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Admin;
```
d4mation: Same comment as https://github.com/stellarwp/kadence-blocks/pull/1041#discussion_r3389747252

includes/resources/Design_Tokens/Admin/Localizer.php
```diff
@@ -0,0 +1,162 @@
...
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Database\Token_Store;
+use KadenceWP\KadenceBlocks\Design_Tokens\Rest\V1\Contracts\Controller;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Exception\Alias_Cycle_Exception;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Exception\Dangling_Alias_Exception;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
+
+/**
+ * Attaches the design-token schema feed to KB's admin dashboard bundle (SOFT-3385).
```
d4mation: We should avoid ticket references in production code.

includes/resources/Design_Tokens/Admin/Localizer.php
```diff
@@ -0,0 +1,162 @@
...
+		try {
+			$values   = $this->resolver->resolve( $slug )->by_id();
+			$variants = $this->variant_feed->all( $slug );
+			$resolved = true;
+		} catch ( Alias_Cycle_Exception | Dangling_Alias_Exception $e ) {
+			$resolved = false; // Corrupt stored document. Fail open: ship structure only.
+		}
+
+		$feed = $this->builder->build( $values, $resolved, $variants, $this->rest(), $version );
+		$json = wp_json_encode( $feed );
```
d4mation: While unlikely, it could be possible that a Token with a XSS exploit was created, right? We should look at escaping this JSON before echoing it into `window`.

As far as I can tell, `wp_add_inline_script()` does no sanitization. 

`esc_js()` can cause some issues with JSON in my experience, but we probably could do 

```php
addslashes(
	Cast::to_string(
		wp_json_encode(
			$feed,
			JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_HEX_QUOT
		)
	)
)
```

includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
```diff
@@ -30,6 +30,15 @@ abstract class Controller extends WP_REST_Controller {
...
 	private const DEFAULT_CAPABILITY = 'edit_theme_options';
 
+	/**
+	 * The shared v1 namespace value. Read by non-controller callers through {@see self::namespace_v1()}.
+	 *
+	 * @since TBD
+	 *
+	 * @var non-falsy-string
+	 */
+	private const NAMESPACE_V1 = 'kb-design-tokens/v1';
```
d4mation: Since the Controller itself is v1-specific I don't think we need to specify that it is the v1 namespace in the const and method name?

includes/resources/Design_Tokens/Admin/Localizer.php
pauloiankoski: Done in d5ac3d4f2 — moved `Localizer` and its test into `Admin\Feed` and updated `Provider` bindings. Related import updates for `Feed\Builder` / `Feed\Variants` are in 319f17471.

includes/resources/Design_Tokens/Admin/Localizer.php
pauloiankoski: Removed in 88ec90770.

includes/resources/Design_Tokens/Admin/Localizer.php
pauloiankoski: Addressed in 6154d2190 — `wp_json_encode()` now uses the `JSON_HEX_*` flags and the output is passed through `addslashes()` before `wp_add_inline_script()`.

includes/resources/Design_Tokens/Rest/V1/Contracts/Controller.php
pauloiankoski: a66418d5e — renamed to `NAMESPACE` / `namespace()` and updated the `Localizer` caller.

d4mation (approved):

PR #1044, opened 2026-06-10
feat(design-tokens): per-block adapter framework (SOFT-3400)

d4mation:
🎫 [SOFT-3400 — Per-block adapter framework](https://linear.app/nexcess/issue/SOFT-3400/per-block-adapter-framework)

## Summary

Adds the per-block adapter framework for the Design Tokens module. An adapter is a named transform keyed to a single Kadence Blocks block type, registered in the Token Registry and applied to that block's attributes while KB assembles them. It exists for the minority of blocks whose attributes are shaped so they cannot be expressed as a single clean token/variable reference: a `Binding` maps one property to one token, but an adapter is free-form and may read several tokens to build one CSS rule (for example `kadence/image` storing `borderRadius` as a four-element array of raw per-corner numbers that render to literal lengths with no `var()` indirection).

The default path is unchanged — no adapter, the CSS variable just works. Adapters apply to Kadence Blocks blocks only; they run through KB's own block-attributes seam, so a WP Core or third-party block is never touched.

## How it works

- `Adapter_Interface` (contract) declares `get_block()` and `apply( array $attributes ): array`.
- `Abstract_Adapter` supplies `get_block()` from a `BLOCK` const, so a concrete adapter only declares the const and the transform.
- `Token_Registry` stores adapters keyed by block (`register_adapter()` / `adapter_for_block()` / `adapters()`), alongside tokens and variant sets.
- The adapter `Projector` is wired to `kadence_blocks_block_default_attributes` (priority 20, after the block-preset overlay) and looks the adapter up by block at render time — fail-soft (a throwing adapter never fatals a render) and fail-closed aware (a no-op when projection is deactivated).
- `kadence_blocks_register_design_token_adapter()` mirrors the module's other global helpers for third-party registration.

## For the other SOFT-3366 tickets

Each per-block wiring ticket that needs an adapter only has to define a class that extends `Abstract_Adapter` with its `BLOCK` constant and an `apply()` method, then add that class to the `ADAPTERS` list on `Projection\Adapter\Provider`. Registration into the Token Registry and invocation on KB's render seam are handled by the framework, so the adapter is auto-wired in place — no hook wiring or filter plumbing per ticket.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Adapter/Projector.php
+++ b/includes/resources/Design_Tokens/Projection/Adapter/Projector.php
@@ -1,0 +1,67 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection\Adapter;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Token_Registry;
+use Throwable;
+
+/**
+ * Applies a block's registered per-block adapter to its attributes while KB assembles them.
+ *
+ * The Provider wires apply() to KB's `kadence_blocks_block_default_attributes` filter, which passes the
+ * block name, so the adapter is looked up per render. It is a no-op — returning the attributes untouched
+ * — when projection is fail-closed, when no adapter is registered for the block, or when the adapter
+ * throws, so it is safe to run for every block on every render.
+ *
+ * @since TBD
+ */
+final class Projector {
+
+	/**
+	 * @var Token_Registry The registry the block's adapter is read from.
+	 *
+	 * @since TBD
+	 */
+	private Token_Registry $registry;
+
+	/**
+	 * @since TBD
+	 *
+	 * @param Token_Registry $registry The token registry.
+	 */
+	public function __construct( Token_Registry $registry ) {
+		$this->registry = $registry;
+	}
+
+	/**
+	 * Run the block's adapter over its attributes, returning the transformed attributes.
+	 *
+	 * @since TBD
+	 *
+	 * @param array<string, mixed> $attributes The block's attributes.
+	 * @param string               $block      The Kadence Blocks block name, e.g. "kadence/advancedheading".
+	 *
+	 * @return array<string, mixed> The transformed attributes; unchanged when the block has no adapter.
+	 */
+	public function apply( array $attributes, string $block ): array {
+		// Respect the fail-closed guard: when projection is disabled (e.g. a declared token is missing
+		// from the baseline) leave the attributes as KB stored them rather than half-transforming them.
+		if ( ! $this->registry->is_active() ) {
+			return $attributes;
+		}
+
+		$adapter = $this->registry->adapter_for_block( $block );
+
+		if ( $adapter === null ) {
+			return $attributes;
+		}
+
+		try {
+			return $adapter->apply( $attributes );
+		} catch ( Throwable $e ) {
+			// This runs in the render path: a faulty adapter must never fatal a page, so fail soft to the
+			// untransformed attributes, matching the block-preset projector.
+			return $attributes;
+		}
+	}
+}
--- a/includes/resources/Design_Tokens/Projection/Adapter/Provider.php
+++ b/includes/resources/Design_Tokens/Projection/Adapter/Provider.php
@@ -1,0 +1,70 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection\Adapter;
+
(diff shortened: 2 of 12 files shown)


includes/resources/Design_Tokens/Registry/Token_Registry.php
```diff
@@ -64,6 +68,20 @@ public function register_variant_set( array $set ): void {
...
+	 * At most one adapter per block; a later registration for the same block replaces the earlier one.
+	 *
+	 * @since TBD
+	 *
+	 * @param Adapter_Interface $adapter The adapter to register.
+	 *
+	 * @return void
+	 */
+	public function register_adapter( Adapter_Interface $adapter ): void {
+		$this->adapters[ $adapter->get_block() ] = $adapter;
```
pauloiankoski: `register_adapter()` stores adapters using whatever value `get_block()` returns, without any validation. Since `Abstract_Adapter::BLOCK` defaults to `''`, a concrete adapter that forgets to override `BLOCK` will be silently registered under an empty key and will never be matched. As a result, its `apply()` method is never called, with no feedback to the developer.

A simple safeguard would be:

```php
if ( $adapter->get_block() === '' ) {
    throw new \InvalidArgumentException(
        'Adapter class ' . get_class( $adapter ) . ' must override BLOCK.'
    );
}
```

This is worth adding before shipping any concrete adapters, as it turns a silent configuration mistake into an explicit error.

includes/resources/Design_Tokens/Projection/Adapter/Provider.php
```diff
@@ -0,0 +1,70 @@
...
+	 * @inheritDoc
+	 *
+	 * @since TBD
+	 */
+	public function register(): void {
+		$this->container->singleton( Projector::class );
+
+		// Register the adapters on init, the same hook the token declarations use, so the registry is fully
+		// populated before render. The filter below then looks each adapter up by block.
+		add_action( 'init', [ $this, 'register_adapters' ], 0 );
```
pauloiankoski: `register_adapters()` runs on `init` at priority `0`, the same priority used by `Registry\Provider::register_declarations()`. At the moment, declarations are registered first because `Registry\Provider` is booted before `Projection\Provider` in `Design_Tokens_Provider::PROVIDERS`, but that dependency is implicit.

If a future adapter constructor depends on the `Token_Registry` being populated—for example, to validate that a token exists—simply reordering the `PROVIDERS` array could introduce a subtle failure.

Consider moving `register_adapters()` to priority `1`. This would make the dependency on declarations explicit and align its sequencing with `guard_baseline()`, which is already scheduled after declarations for the same reason.

includes/resources/Design_Tokens/Projection/Adapter/Projector.php
```diff
@@ -0,0 +1,67 @@
...
+		if ( $adapter === null ) {
+			return $attributes;
+		}
+
+		try {
+			return $adapter->apply( $attributes );
+		} catch ( Throwable $e ) {
+			// This runs in the render path: a faulty adapter must never fatal a page, so fail soft to the
+			// untransformed attributes, matching the block-preset projector.
+			return $attributes;
```
pauloiankoski, line 64: This catches `Throwable`, but the sibling projectors (`Block_Preset\Projector` and `Css_Var\Projector`) only catch `RuntimeException`.

That means a `TypeError` or other `Error` raised during those render paths would propagate uncaught and potentially result in a fatal error, while the adapter projector would handle it gracefully.

Consider widening their catch blocks to `Throwable` so error handling is consistent across all projector implementations.

d4mation: [`fc90e13` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1044/commits/fc90e132c6ff73032188536edd8351c9b9d94206)

includes/resources/Design_Tokens/Projection/Adapter/Provider.php
d4mation: [`fc90e13` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1044/commits/fc90e132c6ff73032188536edd8351c9b9d94206)

includes/resources/Design_Tokens/Registry/Token_Registry.php
d4mation: [`fc90e13` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1044/commits/fc90e132c6ff73032188536edd8351c9b9d94206)

pauloiankoski (approved):

PR #1046, opened 2026-06-11
Chore: create release-prep commands with pup 2.0

jonwaldstein:
## Description
A few things going on here:
- bumps [stellarwp/pup to 2.0.0](https://github.com/stellarwp/pup/releases/tag/2.0.0) to use new commands like `pup replace-version` and `pup replace-tbd`  
- adds [stellarwp/changelogger](https://github.com/stellarwp/changelogger)
- creates a release-prep script that can be called from `composer run release:prep`
- created composer changelogger commands `composer run changelog:add` & `composer run changelog:write`

### Caveats
- I ran into an issue adding `changelogger` to package because of the `stellar-zip.yml` workflow uses github as the node `registry_url` which `changelogger` is not published on (its on npm registry). Might want to circle back and find a solution for this so we can use the package directly via node
- For consistency I added a custom strategy for changelogger

### DEMO

Running `composer run release:prep 1.2.3` to simulate prepping for a release

https://github.com/user-attachments/assets/bbba2217-3816-4278-9282-d410759dede4


composer.json
```diff
@@ -106,9 +106,12 @@
 			"rm -f ./bin/strauss.phar"
 		],
 		"pup": [
-			"test -f ./bin/pup.phar || curl -o bin/pup.phar -L -C - https://github.com/stellarwp/pup/releases/download/1.2.2/pup.phar",
+			"test -f ./bin/pup.phar || curl -o bin/pup.phar -L -C - https://github.com/stellarwp/pup/releases/download/2.0.0/pup.phar",
 			"@php bin/pup.phar"
 		],
+		"release:prep": "dev_scripts/release-prep.sh",
+		"changelog:add": "dev_scripts/changelogger.sh add --auto-filename",
```
d4mation: You may also want to hardcode in `--significance patch` (or similar).

Changelogger is built around trying to infer the correct version to assign when you don't use `--overwrite-version` to specify one by bumping it to the next Major/Minor/Patch based on the changelog entries that exist. I don't think that is a feature we would be using normally though.

If you specify it like this, it is one less prompt for developers to have to click through when creating a changelog entry by pre-selecting the significance for them.

```suggestion
		"changelog:add": "dev_scripts/changelogger.sh add --significance patch --auto-filename",
```

dev_scripts/changelogger.sh
```diff
@@ -0,0 +1,26 @@
...
+#
+# Usage:
+#   dev_scripts/changelogger.sh <changelogger args...>
+#   composer run changelog:add
+#
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+CHANGELOGGER_VERSION="0.10.0"
```
d4mation, line 19: This hardcodes that specific version. Would `^0.10` be better? You'd need to surround it in quotes because `^` is a control character in some shells, but it should work.

`npx -y @stellarwp/changelogger@"^0.10"`

dev_scripts/lib/changelog-strategy.js
```diff
@@ -0,0 +1,113 @@
...
+ * Kadence Blocks readme.txt changelog format:
+ *
+ *   = 3.7.5 =
+ *   Release Date: 1st June 2026
+ *   * Tweak: Pattern and Starter Template libraries now resolve...
+ *   * Fix: Resolved a rendering regression...
+ *
+ * Referenced from package.json -> changelogger.files[].strategy as a path.
+ * The changelogger loads any strategy whose name ends in .js/.ts (resolved
+ * from the current working directory) and validates that it exports
```
d4mation, line 12: _Technically_ this isn't fully accurate. Changelogger only accepts `.ts` for a Writing Strategy for ones _internal_ to Changelogger. You can use TS to create one, but you have to compile it to CommonJS for it to actually be used. I hate it, but haven't been able to come up with a good way to get around it. 

https://github.com/stellarwp/changelogger/#writing-strategies

>IMPORTANT Custom strategy files must be compiled JavaScript (`.js`) files. TypeScript (`.ts`) files are not supported at runtime and must be compiled to JavaScript first. This applies both when using the CLI and programmatically because strategy files are loaded dynamically using Node's `import()`, which requires JavaScript files. If you write your custom writing strategy in TypeScript, compile it to CommonJS JavaScript first. Use the below example and then update your configuration to use the compiled `.js` file.
>
> `tsc path/to/your/custom-writing.ts --outDir path/to/your/ --module CommonJS --target ES2020 --esModuleInterop false --allowSyntheticDefaultImports false --declaration false --sourceMap false --strict --skipLibCheck`

package.json
```diff
@@ -83,5 +83,25 @@
...
+			"type",
+			"content"
+		],
+		"types": {
+			"add": "Add",
+			"update": "Update",
+			"tweak": "Tweak",
+			"fix": "Fix",
+			"security": "Security"
+		},
```
d4mation, line 102: Do we want to introduce custom Type Keys like this? 

It may be more "universal" to use existing keys with custom labels.

Unfortunately, we cannot do that for `update` as it exists in middle ground between `feature` and `tweak`, but we could have `feature` be given the `Add` label.

```suggestion
		"types": {
			"feature": "Add",
			"update": "Update",
			"tweak": "Tweak",
			"fix": "Fix",
			"security": "Security"
		},
```

gilbert-hernandez:
Minor: worth documenting that the changelog step is run-once, not re-runnable

The once-per-release path works correctly — verified against `@stellarwp/changelogger@0.10.0`, output matches the readme.txt format exactly. One sharp edge worth a note in `write-changelog.sh`'s header, though:

Changelogger deletes the consumed `changelog/*.yaml` files after a successful `write`, and re-running `write --overwrite-version <same version>` replaces the *entire* existing version block with only the still-pending entries (it doesn't merge with what's already in `readme.txt`). So running `release:prep <version>` a second time silently drops the entries compiled on the first run (recoverable via git, but not obvious).

Not a blocker — just suggest a line in the script comment like "compile the changelog once per release; handle late entries by editing readme.txt directly rather than re-running."

d4mation:
> Changelogger deletes the consumed `changelog/*.yaml` files after a successful `write`, and re-running `write --overwrite-version <same version>` replaces the _entire_ existing version block with only the still-pending entries (it doesn't merge with what's already in `readme.txt`). So running `release:prep <version>` a second time silently drops the entries compiled on the first run (recoverable via git, but not obvious).

Yes, this is expected, and unfortunately a confusing aspect of `--overwrite-version`. While it _can_ be used to replace an existing version in the changelog file(s), you do need the original entries first.  At that point, the best way to re-run Release Prep is to revert the Release Prep PR itself and then re-run it. Thankfully, GitHub provides a nice link on a merged PR to create a new PR to revert it :) We've had to do it in LearnDash a handful of times when a ticket didn't get merged properly during a maintenance cycle.

composer.json
jonwaldstein: What if its not a _patch_ though?

d4mation: It only matters if you want to trust Changelogger to choose the version number. If you want to trust it, then devs will have to choose the right significance. 

But since the Release Prep uses `--overwrite-version` to set the correct version, then we aren't trusting it.

`--overwrite-version` forces the version number regardless of the set `significance`, so it is effectively ignored. So it doesn't matter what it is.

dev_scripts/changelogger.sh
jonwaldstein: Yeah its a little weird bc we don't have a lock file and can't pull it from npm for reason in the PR description.  I think we actually to keep the pin here so the version doesn't automatically drift away from our already tested workflow.

dev_scripts/lib/changelog-strategy.js
jonwaldstein: ah good to know, that's annoying haha. I think js is fine for now 😄

package.json
jonwaldstein: I'm new to this library, does the key matter? I didn't see a specific default set of anything really but could have been looking in the wrong place

composer.json
jonwaldstein: Ohh, I thought the significance corresponded to the prefix type thingy - but if i'm understanding correctly, its for determining the next version. Your suggestion makes sense now 😄

jonwaldstein: Resolved [80b0592](https://github.com/stellarwp/kadence-blocks/pull/1046/commits/80b05929921063342adf98313632c95142a288d9)

package.json
jonwaldstein: I played around with it and see the key is what you _choose from_. So yeah I think your suggestion makes sense

jonwaldstein: Updated to use feature key for "Add" label [21e4571](https://github.com/stellarwp/kadence-blocks/pull/1046/commits/21e45713886010af332e3b094748110da023d666)

d4mation: The key is also what get stored in the `.yml` file itself :)

d4mation (approved):

PR #1052, opened 2026-06-12
Fix: prevent advanced text block crashing with optional chaining

jonwaldstein:
## Description 
Resolves: [SMTNC-1487](https://linear.app/nexcess/issue/SMTNC-1487/advanced-text-blocks-saved-before-update-break-after-upgrading)

This adds optional chaining to the block's `colorClass` property to prevent a type error.

### Root cause

The existing guard checks color but not colorClass. In block.json, colorClass is {"type": "string"} with no default — it's only ever set when the user picked a named palette color. Any block whose text color is a custom hex (like "color":"#474747") has color set but colorClass undefined, so colorClass.includes(...) throws TypeError: Cannot read properties of undefined (reading 'includes').

### DEMO

https://www.loom.com/share/4b5bcbedc559475b826acc5425e105d2

...

### Checklist

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -57,6 +57,7 @@
         "lottiefiles",
         "lucatume",
         "lwsw",
+		"markformat",
         "mailerlite",
         "mbstring",
         "memize",
--- a/changelog/fix-smtnc-1487.yaml
+++ b/changelog/fix-smtnc-1487.yaml
@@ -1,5 +1,5 @@
 significance: patch
 type: fix
-entry: Resolve an issue with the Advanced Text Block after updating from earlier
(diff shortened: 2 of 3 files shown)


gilbert-hernandez (changes requested):
@jonwaldstein 
LGTM. With the new changelogger, I believe the yaml file should be committed here. Is that correct?

jonwaldstein:
@gilbert-hernandez yup i'll do that, I also noticed the rebase is not working properly 🤔

jonwaldstein:
@gilbert-hernandez all set 🪇

gilbert-hernandez (approved):

PR #1053, opened 2026-06-12
[DTM] SOFT-3407 [1/5] Add Style Book admin screen scaffold

pauloiankoski:
## Summary

- Adds three PHP classes (`Screen`, `Menu`, `Asset_Loader`) that register the Style Book as a submenu under the Kadence admin parent and enqueue its bundle on that screen only.
- Wires the new classes into the Design Tokens admin `Provider` (singletons + `admin_menu` / `network_admin_menu` hooks at priority 11 so the parent menu exists first).
- Adds a `src/style-book.js` webpack entry and a minimal `StyleBookApp` React component that mounts into the PHP-rendered root `<div>`.
- Adds a `src/style-book/style-book.scss` with baseline layout styles for the screen wrapper.
- Registers the new entry as `admin-kadence-style-book` in `webpack.config.js`.

## Test plan

- [ ] Visit **Kadence → Style Book** in a site admin — screen loads with the React scaffold heading and intro text.
- [ ] Verify the submenu appears at position 3 (after Home and Settings).
- [ ] Confirm no JS or PHP errors in the browser console / WP debug log.
- [ ] On a multisite with network-authorize enabled, confirm the submenu also appears under **Network Admin → Kadence → Style Book**.


includes/resources/Design_Tokens/Admin/Style_Book/Screen.php
```diff
@@ -0,0 +1,50 @@
...
+	/**
+	 * The DOM id the React bundle mounts into.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public static function get_root_id(): string {
+		return self::ROOT_ID;
+	}
```
d4mation: This looks like it isn't used in any of the PRs. I'd suggest removing it.

includes/resources/Design_Tokens/Admin/Style_Book/Asset_Loader.php
```diff
@@ -0,0 +1,107 @@
...
+			[ 'wp-components' ],
+			$meta['version'] ?? KADENCE_BLOCKS_VERSION,
+			'all'
+		);
+
+		wp_localize_script(
+			self::SCRIPT_HANDLE,
+			'kadenceStyleBookParams',
+			[
+				'pVersion' => KADENCE_BLOCKS_VERSION,
```
d4mation: `pVersion`? 🤔 I'm guessing `pluginVersion`? We should probably just write it out.

src/style-book/app/StyleBookApp.js
```diff
@@ -0,0 +1,20 @@
...
+import { __ } from '@wordpress/i18n';
+
+/**
+ * The root Style Book admin screen.
+ *
+ * @return {JSX.Element} The Style Book scaffold.
+ */
+export function StyleBookApp() {
+	return (
+		<div className="kadence-style-book">
```
d4mation: All the parent elements are `kadence-blocks`, so I figure we should continue that here.

includes/resources/Design_Tokens/Admin/Provider.php
```diff
@@ -25,6 +28,13 @@ public function register(): void {
 		$this->container->singleton( Builder::class );
 		$this->container->singleton( Variants::class );
 		$this->container->singleton( Localizer::class );
+		$this->container->singleton( Screen::class );
+		$this->container->singleton( Asset_Loader::class );
+		$this->container->singleton( Menu::class );
+
+		// Register after Kadence_Blocks_Settings::add_menu() so the kadence-blocks parent exists.
+		add_action( 'admin_menu', $this->container->callback( Menu::class, 'register_site_menu' ), 11 );
+		add_action( 'network_admin_menu', $this->container->callback( Menu::class, 'register_network_menu' ), 11 );
```
d4mation: Increments of `10` for the hook instead of just being 1 after the default `10` would be better. 

```suggestion
		add_action( 'admin_menu', $this->container->callback( Menu::class, 'register_site_menu' ), 20 );
		add_action( 'network_admin_menu', $this->container->callback( Menu::class, 'register_network_menu' ), 20 );
```

d4mation:
Note: I've written the above reviews for my own reference tomorrow. :)

includes/resources/Design_Tokens/Admin/Style_Book/Screen.php
pauloiankoski: 186baabad66323bc1422c664edf208dbbb2e8735

Just created this because of the private const but yeah, not being used.

includes/resources/Design_Tokens/Admin/Style_Book/Asset_Loader.php
pauloiankoski: a9297efa9b25c440f26a2635658250ba40c84679

includes/resources/Design_Tokens/Admin/Provider.php
pauloiankoski: f08c185e46e65c6ae2582f84598acdfba9a54b71

src/style-book/app/StyleBookApp.js
pauloiankoski: Didn't get this. Are we going to standardize all apps to share the same class name?

d4mation: I figure we should have it be something like `kadence-blocks-style-book`, right? Then child elements could be `kadence-blocks-style-book__header`, etc. 

It feels verbose, but it avoids potential conflicts with Kadence Theme.

pauloiankoski: Oh, it does make sense. I was afraid the idea was to make it `kadence-blocks` only. I'll apply that. It may be reflected on https://github.com/stellarwp/kadence-blocks/pull/1056 too then.

src/style-book/style-book.scss
d4mation, line 1: This is deleted in PR 3, so I won't say how it would be better to format it differently here 😅 I probably ignored it initially because later PRs were were it got out-of-hand.

d4mation (approved):

PR #1055, opened 2026-06-12
[DTM] SOFT-3407 [2/5] Add Style Book data layer (feed, REST client, hooks)

pauloiankoski:
## Summary

- Extends `Localizer` to attach the full design token feed to the Style Book admin bundle as `kadenceStyleBookParams.feed`, making token data available to the React app without an extra network request on load.
- Adds `api/paths.js` (REST endpoint constants) and `api/client.js` (a thin `apiFetch` wrapper for reading and writing tokens).
- Adds `helpers/tokens.js` (group/filter/transform utilities) and `constants.js` (shared app-level constants).
- Adds `hooks/use-design-tokens-feed.js` — reads the localized feed into React state.
- Adds `hooks/use-token-editor.js` — manages inline edit state, dirty tracking, and debounced REST saves with optimistic updates.

## Test plan

- [ ] `kadenceStyleBookParams.feed` is present in the page source when viewing the Style Book screen.


src/style-book/api/client.js
```diff
@@ -0,0 +1,52 @@
...
+ * Internal dependencies
+ */
+import { resolvedPath, tokenPath } from './paths';
+
+/**
+ * Configure apiFetch middleware from the localized REST descriptor.
+ *
+ * @param {{ root: string, nonce: string }} rest REST descriptor from the feed.
+ * @return {void}
+ */
```
d4mation, line 18: It would be a good idea to use `@since TBD` for JS functions as well.

src/style-book/constants.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Default token set slug exposed by the REST API.
+ */
```
d4mation, line 5: Similarly would be good to use `@since TBD` for these consts

src/style-book/api/client.js
pauloiankoski: 9024e1256e855d912b3ef3332392c3b59002c896

src/style-book/constants.js
pauloiankoski: 9024e1256e855d912b3ef3332392c3b59002c896

d4mation (approved):

PR #1056, opened 2026-06-12
[DTM] SOFT-3407 [3/5] Add Style Book token editor UI

pauloiankoski:
## Summary

- Adds atom components: `TokenSwatch` (color preview chip), `TokenTypeBadge` (type label), and `SaveStatus` (saving / saved / error indicator).
- Adds molecule components: `TokenField` (dispatcher that renders the right field type), `ColorTokenField` (WordPress color picker integration), and `GenericTokenField` (plain text input for non-color tokens).
- Adds `TokenGroup` organism — renders a labeled group of token fields.
- Adds `TokenList` template — iterates all token groups and composes them with `TokenGroup`.
- Adds `TokensPage` — the top-level token editor page, wired to `use-token-editor` and `use-design-tokens-feed`.
- Ships each component with a co-located stylesheet (`token-swatch.scss`, `save-status.scss`, `token-field.scss`, `token-group.scss`, `token-list.scss`) imported directly from the component file rather than aggregated in a central entry.

## Test plan

- [ ] The Tokens page renders all tokens grouped by type.
- [ ] Editing a color token opens a color picker; editing a non-color token shows a text input.
- [ ] Saving a token shows the saving → saved indicator; a failed save shows the error state.
- [ ] After a page reload the persisted value is shown.


src/style-book/style-book.scss
d4mation, line 1: This file is getting too large. We should split it into multiple files, ideally by Element where possible. "Token" could easily end up having a lot of styles associated with it and it would be helpful to have it separated out.

pauloiankoski: 8764344f9ca05867f160e6e5adb425a5056a6e92

src/style-book/style-book.scss
```diff
@@ -1,24 +1,22 @@
...
-		min-height: 24rem;
-	}
+@use 'styles/scope' as *;
+@use 'styles/shell' as shell;
+@use 'styles/token' as token;
+@use 'components/pages/tokens-page' as tokens-page;
+@use 'components/templates/token-list' as token-list;
+@use 'components/atoms/token-swatch' as token-swatch;
+@use 'components/atoms/token-type-badge' as token-type-badge;
+@use 'components/atoms/save-status' as save-status;
```
d4mation: Do you intend to use these mixins more than once? If not, they could just be plain `*.scss` files and then included that way and it would work the same way.

pauloiankoski: Refactored to have every component importing its own `*.scss` file.

d4mation (approved):

PR #1059, opened 2026-06-15
feat(design-tokens): deliver media radius to kadence/image

d4mation:
🎫 [SOFT-3502](https://linear.app/nexcess/issue/SOFT-3502)

Delivers the site-wide media-radius design token to `kadence/image` so a fresh image picks up the brand border-radius, while any radius the editor sets (including `0`) still wins.

https://www.loom.com/share/eacdacba10b046c5b6ae0d39e0d5c590

### What this does

Image border-radius is the one dimension that cannot ride an ownable CSS variable: it is empty by default, KB renders it as a literal in both the editor and the front end, and the block has no `var()` indirection to point at the token. This adds a `Block_Default_Css` projector that emits a single low-specificity rule on KB's existing token-var style handles:

```css
.wp-block-kadence-image img { border-radius: var(--kb-token--semantic--radius--media, 1rem); }
```

Because the rule is class-scoped and carries no `!important`, the block's own per-instance radius (a higher-specificity selector) always overrides it when set, so the numeric control keeps working exactly as it does today.

### Scope

This is intentionally narrow — only `kadence/image` border-radius. Other dimensions are delivered elsewhere because they do not fit the mechanism: button radius ships a non-empty `3px` default, and icon size renders as inline SVG width/height and is never empty. Spacing/gap delivery rides the global-var slot mechanism and lands separately.

### Changes

- `Projection/Block_Default_Css/` — the `Css_Builder` (pure CSS generation), `Projector` (front-end + editor enqueue), and `Provider` (container + hook wiring at `wp_enqueue_scripts` 120 / `admin_init` 20).
- `Registry/Binding` — `css_prop` / `css_selector` inline binding targets, with accessors.
- `Registry/declarations.php` — the `semantic.radius.media` token plus the `kadence/image` variant binding.
- wpunit coverage for the builder, the projector wiring, and the new binding targets.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
+++ b/includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
@@ -1,0 +1,191 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection\Block_Default_Css;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Projection\Traits\Sanitizes_Css_Value;
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Token_Registry;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Variant_Resolver;
+use RuntimeException;
+
+/**
+ * Builds the low-specificity, block-scoped CSS that gives a block's dimension property a token-driven
+ * default, e.g. `.wp-block-kadence-image img { border-radius: var(--kb-token--semantic--radius--media, 1rem) }`.
+ *
+ * WHY THIS EXISTS — the use case, in full:
+ *
+ * Most token families reach a block through a variable the block already consumes: colors via the
+ * palette bridge, font sizes via the font-size bridge, spacing/gap by redefining the `--global-kb-*`
+ * variable a stored preset slug already points at (see the Css_Var slot projectors). Radius and icon
+ * size have none of that — Kadence Blocks renders them as raw literals (`border-radius: 12px`), exposes
+ * no ownable `--global-kb-*` variable for them, and their editor controls are plain numeric sliders.
+ *
+ * Two tempting alternatives both fail:
+ *   1. Store a token slug in the attribute and resolve it in the renderer. This hijacks the numeric
+ *      control (the slider can't represent a slug) and has to be mirrored in every block's editor JS.
+ *   2. Redefine a `--global-kb-radius-*` variable. There is none, and nothing references it, so it would
+ *      be a throwaway indirection.
+ *
+ * So instead the token is delivered as a **CSS default, not an attribute value**: one low-specificity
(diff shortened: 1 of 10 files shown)


includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
d4mation, line 1: I want to find a way to at least have an Interface for these builder classes as we have one for each Projector at this point. This is something I'm going to look into separately.

tests/wpunit/Resources/Design_Tokens/Projection/Block_Default_Css/Css_BuilderTest.php
```diff
@@ -0,0 +1,143 @@
...
+	 */
+	public function testItInsertsTheDescendantCombinatorForABareCssSelector(): void {
+		$var = Css_Var::from_id( 'semantic.radius.media' );
+
+		// A bare `img` selector must not need a load-bearing leading space: the builder inserts the
+		// descendant combinator, so it produces the same rule as the explicit ` img` form.
+		$css = $this->builder( $this->image_registry() )->css();
+
+		$this->assertStringContainsString( '.wp-block-kadence-image img{border-radius:var(' . $var . ',0);}', $css );
+		$this->assertStringNotContainsString( '.wp-block-kadence-imageimg', $css );
```
d4mation:
```suggestion
		$this->assertStringNotContainsString( '.wp-block-kadence-imageimg', $css ); // cspell:disable-line -- Checking for invalid selector.
```

includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
```diff
@@ -0,0 +1,219 @@
...
+
+				$var      = $this->registry->css_var_for( (string) $binding->token );
+				$suffix   = $this->selector_suffix( $binding->css_selector() );
+
+				$by_suffix[ $suffix ][] = $prop . ':var(' . $var . ',' . $this->sanitize_value( $value ) . ')';
+			}
+
+			foreach ( $by_suffix as $suffix => $declarations ) {
+				$css .= $selector . $suffix . '{' . implode( ';', $declarations ) . ';}';
+			}
```
nikolaystrikhar, line 157: IMHO, such code blocks (easy to detect such when you do multiple loops) should be private methods for better testing. Not a blocker for this PR though, just for the future.

nikolaystrikhar (approved):

includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
d4mation: Yeah, the CSS Builder classes I want to do some serious rework to. I think they're definitely the easiest place to make improvements within across this module right now :) But for now, probably better to merge so we can see all the different cases side-by-side and can then try to write something to centralize the logic somehow.

PR #1060, opened 2026-06-15
fix: Advanced Heading dynamic link wrong inside FSE query loops

gilbert-hernandez:
## Summary

Fixes the Advanced Heading block's dynamic link resolving to the wrong URL when used inside an FSE archive Query Loop.

In FSE query loops the saved block markup is resolved **once** with the first/queried post's context and the baked HTML is reused for every item. For a plain heading (no icon/gradient), the dynamic link lives as a shortcode in the saved anchor's `href`, so every post after the first rendered with the *first* post's URL.

The fix re-resolves the link from the (unbaked) `link` attribute in the per-post render callback and overwrites the anchor's `href` — matching how the icon/gradient path and the single button block already build their links.

https://www.loom.com/share/08c5aea2a79f411bab2dd0237f4608b6

## Changes

- Re-resolve the href in the plain-heading branch of `build_html()`, gated on the link containing a shortcode (`[`), so static links pass through untouched.
- Use `WP_HTML_Tag_Processor` to rewrite the first `<a>`, matching the Optimizer's existing approach (min WP is 6.6, so the API is available). `esc_url_raw` + `set_attribute` avoids double-encoding ampersands.

## Tests

Adds `tests/wpunit/Blocks/AdvancedHeadingTest.php`:
- `testPlainHeadingResolvesDynamicLinkFromAttributeNotBakedContent` — simulates the query-loop bug (stale baked href + per-post shortcode attribute) and asserts the resolved per-post URL wins.
- `testPlainHeadingLeavesStaticLinkUntouched` — guardrail that a static link passes through unchanged.

### DIFF
--- a/includes/blocks/class-kadence-blocks-advanced-heading-block.php
+++ b/includes/blocks/class-kadence-blocks-advanced-heading-block.php
@@ -805,7 +805,7 @@ class Kadence_Blocks_Advancedheading_Block extends Kadence_Blocks_Abstract_Block
 			$textShadow['vOffset'] = $textShadow['vOffset'] ?? 1;
 			$textShadow['blur']    = $textShadow['blur'] ?? 1;
 			$textShadow['color']   = $textShadow['color'] ?? null;
-			$textShadow['opacity'] = $textShadow['opacity'] ?? 1.0; // Default is 0.2, but if it's undefined they set it at a time when the block defaults it to 1.0
+			$textShadow['opacity'] = $textShadow['opacity'] ?? 1.0; // Default is 0.2, but if it's undefed they set it at a time when the block defaults it to 1.0
 		}
 
 		if (!empty($attributes['textShadowTablet']) && is_array($attributes['textShadowTablet'][0])) {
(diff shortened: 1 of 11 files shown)


gilbert-hernandez:
There is an issue here. The cspell lints the whole changed file, but partial-phpcs only lints lines in the diff. So, if I fix the spelling error "undefed" (should be undefined) in line 808 of the class-kadence-blocks-advanced-heading-block.php file, phpcs suddenly want to fix the snake casing in legacy function. Every line I touch would then trigger phpcs to want to update all the snake casing for the rest of the function. This seems like a risky refactor that is out of scope and not worth it. The easier solution was to add "undefed" to the cspell.json file.

includes/blocks/class-kadence-blocks-advanced-heading-block.php
```diff
@@ -617,6 +620,21 @@ public function build_html( $attributes, $unique_id, $content, $block_instance )
 				$wrapper_attributes = implode( ' ', $wrapper_attributes );
 				$content = sprintf( '<div %1$s>%2$s</div>', $wrapper_attributes, $content );
 			}
+		} elseif ( ! empty( $attributes['link'] ) && false !== strpos( $attributes['link'], '[' ) ) {
+			// Plain heading: the dynamic link lives as a shortcode in the saved anchor's href.
```
WayneRocha, line 624: maybe we could shorten or remove this comment block. Except for that, LGTM

WayneRocha (approved):
LGTM.

PR #1061, opened 2026-06-15
feat(design-tokens): spacing/gap token delivery via the global-var slot bridge

d4mation:
🎫 [SOFT-3401](https://linear.app/nexcess/issue/SOFT-3401) · [SOFT-3402](https://linear.app/nexcess/issue/SOFT-3402)

https://www.loom.com/share/8cbaff72ae5243a9bd9c4a35d49e763f

Delivers the spacing/gap slot bridge and tokenizes Kadence Blocks' full spacing and gap scale, so a site owner can retune every size step (XS, MD, XL, XXL, ...) site-wide and any block already storing that slug follows the token with no per-block change.

### The problem

Kadence Blocks renders a dimension attribute holding a preset slug as `var(--global-kb-<family>-<slug>, <fallback>)`, but ships those slug values as plain literals with no filter to override - and for gap it does not define the variable at all. So unlike colors and font sizes, there is no seam for a token to retarget a stored spacing/gap slug.

### The mechanism

The `Css_Var` builder emits `--global-kb-<family>-<slug>: var(--kb-token--..., <literal>)` overrides under the same `:root` scope, later in source order than KB's own definitions:

```css
:root,:root:where(.kb-tokens){--global-kb-spacing-xl:var(--kb-token--primitive--dimension--spacing--xl,4rem);}
```

Every block already storing that slug then resolves to the token, with the resolved length as a literal fallback for contexts that lack the token vars. A token claims a slug through a projection key (`kb_spacing_slot` / `kb_gap_slot`); the families are described by a `Css_Var\Slot\Contracts\Target` contract with a shared `Css_Var\Slot\Contracts\Abstract_Target` base, so spacing and gap share one generic `slot_block()` path.

### The model (mirrors `radius`)

The token layering follows the same shape `radius` already uses - primitive = the scale, semantic = the usage:

- **Primitive = the scale.** `primitive.dimension.spacing.{xxs..5xl}` and `primitive.dimension.gap.{none,xs,sm,md,lg}` are the scale steps, and each **claims its Kadence Blocks slug** (the slug *is* a scale step). Retuning "XL spacing" site-wide means overriding `primitive.dimension.spacing.xl`. `ss-auto` is omitted - it resolves to `auto`, not a length.
- **Semantic = the usage.** `semantic.spacing.{inline,block,section}` alias the scale (md / lg / xl), exactly like `semantic.radius.media` aliases the radius scale. These are where intent-based delivery points - e.g. the Row Layout / Column `$default` padding presets resolve through `semantic.spacing.section` / `.block`. Gap has no usage consumer yet, so its primitive scale carries the slug claims and a `semantic.gap` usage layer lands with the per-block work.

`primitive.dimension.space` is renamed to `primitive.dimension.spacing` so every dimension family matches across layers (spacing/spacing, gap/gap, radius/radius, borderWidth/borderWidth).

Token defaults match KB's own runtime values, so registering them changes nothing until a site owner overrides one - the scale (`xxs 0.5rem ... 5xl 10rem`, gap `none 0 ... lg 4rem`) is KB's, matching the shipped design-system prototype.

**Note on `gap.xs`:** `--global-kb-gap-xs` (0.5rem) exists in KB's `$gap_sizes` map, so `primitive.dimension.gap.xs` is included for completeness - but Kadence Blocks does not currently expose a way to select it. The gutter controls surface only four steps: Row Layout stores `none/skinny/default/wider` (the button labels read None/SM/MD/LG) and Column stores `none/sm/md/lg`, all of which resolve to `--global-kb-gap-{none,sm,md,lg}`; the `custom` toggle is a raw numeric, not a preset slug. Nothing maps to `--global-kb-gap-xs`, so that step is inert until KB exposes it.

**Note for reviewers:** `primitive.dimension.spacing` previously held a different, abstract scale (0.25-4rem, renamed from `space`) that diverged from KB's runtime. Aligning it to KB's ramp shifts the existing `semantic.spacing.*` usage aliases - so the Row Layout / Column `$default` padding presets now resolve to KB's real 4rem / 3rem rather than the old 4rem / 2rem.

### Scope: what this covers for SOFT-3401 / SOFT-3402, and what's deferred

3401 (Row Layout) and 3402 (Section / Column) call for wiring those blocks to color, border, **and spacing** tokens. This PR is the **spacing/gap delivery** half:

- **Spacing/gap is complete for these blocks here.** Row Layout's padding default is `["sm", "", "sm", ""]` (slugs) and its gutter stores slugs too; Column is the same shape. Because the slug bridge retargets `--global-kb-spacing-*` / `--global-kb-gap-*`, a fresh Row Layout / Column already follows the spacing and gap tokens with **no per-block change** - the "no adapter needed" happy path the tickets describe.

**Deferred to the per-block wiring follow-up (SOFT-3401 / 3402):**

- **Color (background) and border wiring** for Row Layout / Column. These are *not* declarative: the block-preset seeder assigns a single flat value to one attribute, but these attributes are structured - `padding` is a `type: array` of four slugs, `border` is a `type: string`, and the background is not a flat attribute - so a flat token value cannot be seeded into them without reshaping. That needs the per-block adapter framework plus the per-block mapping spike, which is the substance of the 3401/3402 per-block work and belongs in its own PR.
- **Per-role empty-state defaults** (e.g. "a section's default padding when the attribute is left empty") - a different mechanism (the low-specificity block-default-CSS rule, as used for image border-radius), not a slug claim.

The baseline `$default` presets for `kadence/rowlayout` / `kadence/column` (background + padding) are placeholders for that follow-up: no variant set is registered for these blocks yet, so they are inert today.

### Changes

- `Projection/Css_Var/Slot/Contracts/Target.php` + `Abstract_Target.php` - the slot-family contract and shared base.
- `Projection/Css_Var/Slot/Spacing_Target.php` / `Gap_Target.php` - the spacing and gap families.
- `Projection/Css_Var/Css_Builder.php` - the generic `slot_block()`.
- `Registry/declarations.php` - registers the primitive spacing/gap scale steps, each claiming its slug.
- `Registry/Baseline/baseline.json` - `primitive.dimension.spacing` (renamed, KB ramp) + `primitive.dimension.gap`, usage-specific `semantic.spacing.{inline,block,section}`, repointed presets.
- `AGENTS.md` - how to flush the design-tokens baseline cache during local development.
- wpunit coverage for both targets, the builder's per-slug output, and an end-to-end check that the shipped declarations emit every slug.

### Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -143,6 +143,30 @@ bunx cspell lint -c .cspell.json --no-progress --no-must-find-files --dot <files
 Needs Node >= 22.18. If the default shell is on an older Node, switch first (e.g. via nvm:
 `nvm use 22`).
 
+### Flushing the design-tokens baseline cache
+
+The decoded `baseline.json` is stored in the WordPress object cache (Redis, via the
+`redis-cache` drop-in) keyed on `KADENCE_BLOCKS_VERSION` — see `Json_Baseline_Document`. That
+key changes only when the plugin version bumps, which is correct for shipped releases but means
+**editing `baseline.json` in place during development does NOT invalidate the cache**. The stale
+decoded baseline keeps being served while `declarations.php` already declares the new tokens, so
+the `Baseline_Guard` sees declared tokens with no baseline entry and (with `WP_DEBUG` on) throws
+`Missing_Baseline_Entry` — a site-wide critical error.
+
+Fix it by flushing Redis directly:
+
+```bash
+# from the lando app root (e.g. ~/stellarwp/sfwd-lms)
+lando ssh -s redis -c 'redis-cli flushall'
+```
+
+Do NOT reach for `lando wp cache flush` here: wp-cli boots WordPress, the guard fires on `init`,
+and it throws before the flush command runs. The out-of-band Redis flush is the escape hatch.
+(Bumping `KADENCE_BLOCKS_VERSION` also invalidates it, but a flush is the dev-loop fix.) Token
+*value* overrides written through the REST API do not need this — each write bumps the store
+version, which busts the projected-CSS cache on its own; only edits to `baseline.json` itself go
+stale this way.
+
 ## Git, commits, and PRs
 
 - **Backtick `@`-mentions in commit/PR text.** Wrap any `@`-prefixed token in backticks —
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -156,7 +156,7 @@ the `Baseline_Guard` sees declared tokens with no baseline entry and (with `WP_D
 Fix it by flushing Redis directly:
 
 ```bash
-# from the lando app root (e.g. ~/stellarwp/sfwd-lms)
+# from the lando app root
 lando ssh -s redis -c 'redis-cli flushall'
 ```
 
--- a/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
+++ b/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
@@ -57,23 +57,65 @@
 		},
 		"dimension": {
 			"space": {
-				"xs": {
-					"$type": "dimension",
-					"$value": "0.25rem"
-				},
-				"sm": {
+				"xxs": {
 					"$type": "dimension",
 					"$value": "0.5rem"
 				},
-				"md": {
+				"xs": {
 					"$type": "dimension",
 					"$value": "1rem"
 				},
-				"lg": {
+				"sm": {
+					"$type": "dimension",
+					"$value": "1.5rem"
+				},
+				"md": {
 					"$type": "dimension",
 					"$value": "2rem"
 				},
+				"lg": {
+					"$type": "dimension",
+					"$value": "3rem"
+				},
 				"xl": {
+					"$type": "dimension",
+					"$value": "4rem"
+				},
+				"xxl": {
+					"$type": "dimension",
+					"$value": "5rem"
+				},
+				"3xl": {
+					"$type": "dimension",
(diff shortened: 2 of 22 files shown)


includes/resources/Design_Tokens/Registry/Baseline/baseline.json
```diff
@@ -57,25 +57,67 @@
...
 				"md": {
 					"$type": "dimension",
-					"$value": "1rem"
+					"$value": "2rem"
 				},
 				"lg": {
 					"$type": "dimension",
-					"$value": "2rem"
+					"$value": "3rem"
 				},
```
d4mation, line 79: These values didn't match the existing ones in Kadence Blocks and now they do.

AGENTS.md
```diff
@@ -143,6 +143,30 @@ bunx cspell lint -c .cspell.json --no-progress --no-must-find-files --dot <files
...
+key changes only when the plugin version bumps, which is correct for shipped releases but means
+**editing `baseline.json` in place during development does NOT invalidate the cache**. The stale
+decoded baseline keeps being served while `declarations.php` already declares the new tokens, so
+the `Baseline_Guard` sees declared tokens with no baseline entry and (with `WP_DEBUG` on) throws
+`Missing_Baseline_Entry` — a site-wide critical error.
+
+Fix it by flushing Redis directly:
+
+```bash
+# from the lando app root (e.g. ~/stellarwp/sfwd-lms)
```
d4mation:
```suggestion
# from the lando app root
```

nikolaystrikhar (approved):

PR #1062, opened 2026-06-15
feat(design-tokens): wire Row Layout & Column background, border & radius to tokens

d4mation:
🎫 [SOFT-3401](https://linear.app/nexcess/issue/SOFT-3401) · [SOFT-3402](https://linear.app/nexcess/issue/SOFT-3402)

> Stacked on [the spacing/gap delivery PR](https://github.com/stellarwp/kadence-blocks/pull/1061) — base it there; review/merge that first. This branch adds the color/border/radius half of the Row Layout / Section wiring.

This is largely the same, mechanically, as https://github.com/stellarwp/kadence-blocks/pull/1059 and uses the same underlying code.

Wires `kadence/rowlayout` and `kadence/column` background, border color, and border radius to design tokens, completing the per-block color wiring 3401/3402 call for (the spacing half lands in the base PR via the slug bridge).

### How it works

These blocks store their background, border color, and border radius as a literal value — a hex, a palette slug, or a number+unit — **only when a user manually sets one**; an uncustomized block stores nothing. To make the **default** state follow a token, this reuses the block-default-CSS projector — the same low-specificity, block-scoped mechanism that delivers image border-radius. One rule per block points each bound property at a token variable:

```css
.wp-block-kadence-rowlayout{background-color:var(--kb-token--semantic--color--rowlayout-bg,transparent);border-color:var(--kb-token--semantic--color--border,#E2E8F0);border-radius:var(--kb-token--semantic--radius--rowlayout,0);}
.wp-block-kadence-column> .kt-inside-inner-col{background-color:var(--kb-token--semantic--color--column-bg,transparent);border-color:var(--kb-token--semantic--color--border,#E2E8F0);border-radius:var(--kb-token--semantic--radius--column,0);}
```

The row targets the block root; the column targets its inner `.kt-inside-inner-col` (where KB renders all three). The rule is a single class emitted earlier in source order than KB's per-instance CSS, so it supplies the default while any literal value the user sets (a per-instance rule of equal specificity, later in source order) overrides it. This is the same fit as image radius: a property KB renders as a literal when set and leaves empty otherwise, where only a token-backed default was missing.

### Defaults match KB's own, so nothing is restyled

A Row Layout / Column has no background and no rounding by default in KB. So the tokens are chosen to resolve to those same defaults, and an uncustomized block looks identical to today:

- **Background** binds to block-specific tokens — `semantic.color.rowlayout-bg` / `semantic.color.column-bg` — that **alias `primitive.color.transparent`**.
- **Border radius** binds to block-specific tokens — `semantic.radius.rowlayout` / `semantic.radius.column` — that **alias `radius.none`** (resolve to `0`), mirroring `semantic.radius.media` for the image.
- **Border color** follows `semantic.color.border`; it is invisible until a user adds a border (border-width defaults to 0), then defaults to the brand border color.

Each block type has its own override seam: a site owner brands all rows by overriding `rowlayout-bg` / `semantic.radius.rowlayout` (columns independently), without touching the shared `surface` colors, the `transparent` primitive, or the `radius` scale.

### Changes

- `Registry/declarations.php` — register the `rowlayout-bg` / `column-bg` / `border` color tokens and the `rowlayout` / `column` radius tokens (bare, so `Css_Var` emits their variables), and the `kadence/rowlayout` + `kadence/column` variant sets binding `background-color` / `border-color` / `border-radius` via `css_prop`.
- `Registry/Baseline/baseline.json` — `semantic.color.rowlayout-bg` / `column-bg` aliasing the transparent primitive and `semantic.radius.rowlayout` / `column` aliasing `radius.none`; the rowlayout/column `$default` presets reference them. `surface` / `surface-alt` keep their real values (matching the design-system Storybook Background scale: base `#FFFFFF`, subtle `#F7FAFC`).
- wpunit coverage that the shipped declarations emit both blocks' background + border + radius rules.

### Follow-up

- The Kadence color/radius controls won't show a swatch *selected* for these defaults — the value comes from the CSS rule, not a stored attribute. Showing it as the selected value is the adapter / variant path (needs a Kadence variant picker UI, which doesn't exist yet).
- `semantic.color.border` (`#E2E8F0`) has no exact match in the Storybook palette; left as the baseline neutral for now.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
+++ b/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
@@ -265,6 +265,14 @@
 			"icon": {
 				"$type": "color",
 				"$value": "{primitive.color.brand.primary}"
+			},
+			"rowlayout-bg": {
+				"$type": "color",
+				"$value": "{primitive.color.transparent}"
+			},
+			"column-bg": {
+				"$type": "color",
+				"$value": "{primitive.color.transparent}"
 			}
 		},
 		"spacing": {
@@ -424,7 +432,8 @@
 					"default": {
 						"label": "Default",
 						"tokens": {
-							"background": "{semantic.color.surface}",
+							"background": "{semantic.color.rowlayout-bg}",
+							"border": "{semantic.color.border}",
 							"padding": "{semantic.spacing.section}"
 						}
 					}
@@ -434,7 +443,8 @@
 					"default": {
 						"label": "Default",
 						"tokens": {
-							"background": "{semantic.color.surface-alt}",
+							"background": "{semantic.color.column-bg}",
+							"border": "{semantic.color.border}",
 							"padding": "{semantic.spacing.block}"
 						}
 					}
--- a/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
+++ b/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
@@ -297,6 +297,14 @@
 			"media": {
 				"$type": "dimension",
 				"$value": "{primitive.dimension.radius.none}"
+			},
+			"rowlayout": {
+				"$type": "dimension",
+				"$value": "{primitive.dimension.radius.none}"
+			},
+			"column": {
+				"$type": "dimension",
+				"$value": "{primitive.dimension.radius.none}"
 			}
 		},
 		"borderWidth": {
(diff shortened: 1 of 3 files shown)


dave-green-uk (approved):
LGTM, nothing to flag. Nice!

PR #1063, opened 2026-06-16
Button (kadence/advancedbtn + singlebtn) token wiring + variants

d4mation:
🎫 https://linear.app/nexcess/issue/SOFT-3406/button-kadenceadvancedbtn-singlebtn-token-wiring-variants

📹 https://www.loom.com/share/4ab4371b14694e5493af499c2ea1cc54

Wires the Button to design-token variants, models its colors as overridable per-variant semantics, and tokenizes its dimension defaults.

## What this does

- **Color variants (Primary / Secondary).** `kadence/singlebtn` gets a "Design Variant" picker (a `KadenceRadioButtons` control fed by a per-block catalog localized to the editor). Variants are a pure color axis: they retarget the button palette vars — `--global-palette-btn-bg`/`-btn` and their `-hover` counterparts — so they compose with the existing "Button Inherit Styles" shape control (Fill / Outline / Theme Base) rather than replacing it (e.g. Secondary + Outline = a secondary-colored ghost), and hover/focus follow the variant too. `core/button` gets the same variants through WordPress's native block Styles panel (the Block_Style projector: `register_block_style` + theme.json default).
  - **Primary** reproduces the block's existing default look: `brand.button` `#3633e1` resting → `brand.button-hover` `#2f2ffc` on hover, white text (the block's own `--global-palette-btn-bg` / `-bg-hover` SCSS fallbacks), so activation changes nothing.
  - **Secondary** is a dark/charcoal button: `neutral.900` `#1A202C` → `neutral.700` `#2D3748` on hover, white text (WCAG AAA), clearly distinct from the Primary.
- **Per-variant color semantics.** Each variant's colors come from named semantics (`semantic.color.button-{primary,secondary}-{bg,text,bg-hover,text-hover}`), each surfaced to the site editor, so a site owner recolors one variant through the standard token path without touching the global palette. Every semantic resolves to a primitive (no semantic-to-semantic chains); button-specific colors live as role-named primitives (`brand.button` / `brand.button-hover`, named for the var they back rather than the hue).
- **Global palette follows the primitives.** The brand + neutral primitives each claim a Kadence palette slot (`palette1..9`), so `--global-paletteN` (and the legacy `kadence_blocks_colors` palette) follows the primitive. Values mirror Kadence's default palette, so activation changes nothing until a primitive is overridden. Semantic colors deliberately claim **no** palette slot — they deliver at the block level — so writing a button semantic never re-skins the global palette.
- **Radius.** Registers `semantic.radius.control` and tokenizes the button's own default `border-radius` to `var(--kb-token--semantic--radius--control, 3px)` (resolves to the design-system control radius); an explicit per-button radius still wins by specificity.
- **Icon size.** Registers `semantic.icon-size.default` and tokenizes the button icon default.
- **Conventions.** Renames the camelCase `iconSize` baseline token to `icon-size` and documents in `AGENTS.md` that design-token ids must be kebab-case (camelCase fails registration). Built assets (`dist/`) are gitignored and rebuilt in CI.

## Deferred (follow-up tickets under SOFT-3366)

- **SOFT-3529** — Button typography token wiring (the `btns[]`-array shape + `render_font_family` indirection is bespoke, so it is split out).
- **SOFT-3533** — Grouped (orthogonal) variant axes. Variants are single-axis today; "Ghost" stays the native Outline shape (Tertiary/Ghost are V2 per the design spec). Filed now so the variant schema set here can grow into groups without a breaking migration.

### Checklist

- [ ] Tested with an existing instance of this block.

### DIFF
--- a/includes/blocks/class-kadence-blocks-singlebtn-block.php
+++ b/includes/blocks/class-kadence-blocks-singlebtn-block.php
@@ -5,6 +5,8 @@
  * @package Kadence Blocks
  */
 
+// cspell:ignore glight glightbox ktblocksvideopop plyr .
+
 // Exit if accessed directly.
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
@@ -333,10 +335,10 @@ class Kadence_Blocks_Singlebtn_Block extends Kadence_Blocks_Abstract_Block {
 	/**
 	 * Build HTML for dynamic blocks
 	 *
-	 * @param $attributes
-	 * @param $unique_id
-	 * @param $content
-	 * @param WP_Block   $block_instance The instance of the WP_Block class that represents the block being rendered.
+	 * @param array<string, mixed> $attributes The block attributes.
+	 * @param string               $unique_id The block's unique id.
+	 * @param string               $content The block inner content.
+	 * @param WP_Block             $block_instance The instance of the WP_Block class that represents the block being rendered.
 	 *
 	 * @return mixed
 	 */
@@ -361,7 +363,7 @@ class Kadence_Blocks_Singlebtn_Block extends Kadence_Blocks_Abstract_Block {
 			 * projector's scoped CSS hooks. This is a dynamic block, so the class is added here rather than by
 			 * the editor save filter; the sanitizer mirrors the projector's so the class matches the selector.
 			 */
-			$classes[] = 'kb-variant--' . preg_replace( '/[^A-Za-z0-9_-]+/', '-', $attributes['kbVariant'] );
+			$classes[] = 'kb-variant--' . preg_replace( '/[^A-Za-z0-9_-]+/', '-', $this->attribute_string( $attributes, 'kbVariant' ) );
 		}
 
 		if ( ! empty( $attributes['target'] ) && 'video' === $attributes['target'] ) {
@@ -374,13 +376,13 @@ class Kadence_Blocks_Singlebtn_Block extends Kadence_Blocks_Abstract_Block {
 			'class' => implode( ' ', $classes ),
 		];
 		if ( ! empty( $attributes['anchor'] ) ) {
-			$wrapper_args['id'] = $attributes['anchor'];
+			$wrapper_args['id'] = $this->attribute_string( $attributes, 'anchor' );
 		}
 		if ( ! empty( $attributes['label'] ) ) {
-			$wrapper_args['aria-label'] = $attributes['label'];
+			$wrapper_args['aria-label'] = $this->attribute_string( $attributes, 'label' );
 		}
 		if ( ! empty( $attributes['link'] ) ) {
-			$wrapper_args['href'] = esc_url( do_shortcode( $attributes['link'] ) );
+			$wrapper_args['href'] = esc_url( do_shortcode( $this->attribute_string( $attributes, 'link' ) ) );
 			$rel_add              = '';
 			if ( isset( $attributes['download'] ) && $attributes['download'] ) {
 				$wrapper_args['download'] = '';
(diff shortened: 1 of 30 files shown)


includes/resources/Design_Tokens/Registry/Baseline/baseline.json
```diff
@@ -466,15 +466,6 @@
...
 						}
-					},
-					"ghost": {
-						"label": "Ghost",
-						"tokens": {
-							"button-bg": "transparent",
-							"button-text": "{primitive.color.brand.primary}",
-							"button-border": "{primitive.color.brand.primary}",
-							"button-radius": "{semantic.radius.control}"
-						}
```
d4mation, line 497: According to the spec within the prototype from design, Ghost will be in v2. 

It is indirectly covered for `kadence/singlebtn` anyway by an existing setting which is compatible with our background color variants as-is:

image

src/blocks/advancedbtn/editor.scss
```diff
@@ -219,13 +219,13 @@
 	padding: 0.4em 1em;
 	cursor: pointer;
 	font-size: 1.125rem;
-	border-radius: 3px;
+	border-radius: var(--kb-token--semantic--radius--control, 3px);
```
d4mation, line 222: These defaults allow the original CSS to be used if the Token is undefined while still allowing our Tokens to be used.

includes/blocks/class-kadence-blocks-singlebtn-block.php
```diff
@@ -472,6 +482,23 @@ public function register_scripts() {
...
+	 * @param array<string, mixed> $attributes The block attributes.
+	 * @param string               $key        The attribute key to read.
+	 *
+	 * @return string
+	 */
+	private function attribute_string( array $attributes, string $key ): string {
+		$value = $attributes[ $key ] ?? '';
+
+		return is_scalar( $value ) ? (string) $value : '';
+	}
```
d4mation, line 501: I don't like this, but we added the `Cast` utility to the `Design_Tokens` namespace instead of something more general. I'll make another PR to move it (and everywhere it is used) to avoid this new method.

includes/resources/Design_Tokens/Registry/Baseline/baseline.json
```diff
@@ -9,11 +9,19 @@
...
 					"$type": "color",
 					"$value": "#ED8936"
+				},
+				"button": {
+					"$type": "color",
+					"$value": "#3633e1"
+				},
+				"button-hover": {
+					"$type": "color",
+					"$value": "#2f2ffc"
```
d4mation, line 24: I'm not a big fan of `primitives.color.brand.button`/`primitives.color.brand.button-hover`, but we can change it later. We realistically may want something like the `neutral` scale for `primary` and `secondary`.

To explain these changes in full:

1. `primitives.color.brand.secondary` was updated to `#2B6CB0` to better reflect the existing default for `palette2` in Kadence Blocks.
2. `primitives.color.brand.button` and `primitives.color.brand.button-hover` had to be added because the existing colors in Kadence Blocks don't match _any_ of the primitives and it is better to have a primitive hold the hex value with semantic values that exist to overwrite it. 
    - These are effectively the "default button colors" because the variants are able to pull from these and other primitives more cleanly.

includes/resources/Design_Tokens/Registry/Baseline/baseline.json
```diff
@@ -256,12 +268,44 @@
...
+				"$value": "{primitive.color.neutral.0}"
+			},
+			"button-secondary-bg-hover": {
+				"$type": "color",
+				"$value": "{primitive.color.neutral.700}"
+			},
+			"button-secondary-text-hover": {
+				"$type": "color",
+				"$value": "{primitive.color.neutral.0}"
+			},
```
d4mation, line 308: Secondary buttons point to neutrals to help ensure that they're visually distinct from the Primary button. 

If we used the `primitives.color.brand.secondary` Primitive, then they would both be blue and hard to distinguish when presented side-by-side for different actions ("Submit" vs "Close", etc.)

Again, not sold on the exact tokens here. But I wanted to get something _working_ and then we can easily change them later.

includes/resources/Design_Tokens/Registry/Baseline/baseline.json
d4mation: https://lw.slack.com/archives/C0BBB5R2LHF/p1781729131126329?thread_ts=1781729128.645469&cid=C0BBB5R2LHF

nikolaystrikhar (approved):
This PR is a beast and I'm not confident at all, but I also don't see anything insane, so... :D

PR #1064, opened 2026-06-18
SOFT-3533: design-token variant picker & composition infrastructure

d4mation:
🎫 [SOFT-3533](https://linear.app/nexcess/issue/SOFT-3533/grouped-orthogonal-design-token-variants-multi-axis-variant-model)

https://www.loom.com/share/410fa68ed392494e9a67497c862bd598

Refines the block-editor surface of the design-token variant system so a block's variant picker is reusable and lives in the block's Style tab.

- **Reusable `VariantPicker`** (`src/extension/variant-picker`) — reads the variant catalog and renders the radio control, identical wherever it surfaces.
- **Picker in the Style tab** — `kadence/singlebtn` renders it inside its own Style tab via `"kbVariant": { "inlinePicker": true }`; the generic inspector HOC uses `InspectorControls group="styles"` and skips inline-picker blocks to avoid a duplicate.
- **Declaration-driven control label** — a variant set may declare its picker label (e.g. `Style`) in `declarations.php`; `Variant_Set` carries it, `Variant_Catalog` surfaces it, the picker reads it (generic fallback otherwise).

Targets `feature/design-tokens-module`. The `core/button` rework ([SOFT-3576](https://linear.app/nexcess/issue/SOFT-3576/wire-corebutton-into-the-design-token-variant-system)) is stacked on top of this in a separate PR.

### Checklist


src/blocks/singlebtn/edit.js
```diff
@@ -375,13 +376,7 @@ export default function KadenceButtonEdit(props) {
 		{ value: 'full', label: __('Full', 'kadence-blocks') },
 	];
 	const defineWidthType = (type) => {
-		if ('full' === type) {
-			//updateParentBlock( 'forceFullwidth', true );
-			setAttributes({ widthType: type });
-		} else {
-			//updateParentBlock( 'forceFullwidth', false );
-			setAttributes({ widthType: type });
-		}
```
d4mation, line 384: These comments resulted in a cspell issue we didn't need to worry about. Since the conditional would have only mattered with the commented out code, it has been removed entirely and replaced with the identical `setAttributes()` line.

nikolaystrikhar (approved):

PR #1065, opened 2026-06-18
SOFT-3576: wire core/button into the design-token variant system

d4mation:
🎫 [SOFT-3576](https://linear.app/nexcess/issue/SOFT-3576/wire-corebutton-into-the-design-token-variant-system)

https://www.loom.com/share/79e8cd8b3b244e0fbc785d140b61d343

> **Stacked on #1064** (SOFT-3533) — review/merge that first. This PR targets the SOFT-3533 branch, so its own diff is just the `core/button` rework.

Wires `core/button` into the design-token variant system so its color axis composes with WordPress's built-in **Outline** instead of colliding with it (block styles are single-select).

- `core/button` reuses the Kadence button's additive `kb-variant--<slug>` class (opts in via `enableNativeBlockVariants`), so the color variant composes with `is-style-outline` rather than replacing it.
- Binds `core/button` to the same `--global-palette-btn-*` slots as the Kadence button, so the shared `Variant\Css_Builder` retarget serves both (block-aware selector: `core/button` → `.wp-block-button`).
- New `Projection\Native` module — a `Styles\Contracts\Styles` interface + `Styles\Button` companion stylesheet that makes the native button link consume those slots for Fill / Outline / hover, with a block-agnostic `Projector` iterating registered companions. Replaces the removed `Projection\Block_Style` (`register_block_style`) path.
- Default styling is gated on the Kadence palette `override` setting: when the theme palette is replaced (`kadence_blocks_colors` "override"), the design system owns every `core/button`'s default (Primary); otherwise an untouched `core/button` keeps its native theme styling.
- Relocates the `kb-variant--` class helper to `Projection\Variant\Style`.

### Checklist


includes/resources/Design_Tokens/Projection/Block_Style/Css_Builder.php
```diff
@@ -1,271 +0,0 @@
...
-use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Token_Registry;
-use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Variant_Resolver;
-use RuntimeException;
-
-/**
- * Builds the scoped CSS for native block-style variants on non-Kadence blocks.
- *
- * A native block (core/button and the like) reaches its variants through WordPress's own block-style
- * system: each named variant is registered with register_block_style(), so the editor adds an
- * "is-style-kb-<name>" class when it is picked, and this builder emits, per (block, variant), a rule that
```
d4mation, line 21: This is why we are avoiding this previous method. We cannot choose an existing Style along with our Variants which feels like a mistake.

nikolaystrikhar (approved):
I can only review it to a degree, without a deep understanding, I don't see any obvious problems with the code, and the loom looks good.

PR #1068, opened 2026-06-19
[SMTNC-771] Button (Adv) color settings aren’t working when using Shopkit + FSE theme.

coutinhocouto:
🎫  [SMTNC-771](https://linear.app/nexcess/issue/SMTNC-771/button-adv-color-settings-arent-working-when-using-shopkit-fse-theme)

### 🎥 Artifacts
[https://www.loom.com/share/07e166832b1444499d468a7a8cfd8cfd](https://www.loom.com/share/07e166832b1444499d468a7a8cfd8cfd)

### 🗒️ Description

Fixed Button (Adv) color settings (background, text, and border), and other block style settings, not applying on the frontend when the block is rendered in the page **body** under a block (FSE) theme, e.g. inside a Shop Kit Product Loop Item, a Query Loop, or an FSE template/template part.

...

### Checklist

### DIFF
--- a/changelog/smtnc-771-button-adv-color-settings-arent-working-when-using-shopkit-1782139769133.yaml
+++ b/changelog/smtnc-771-button-adv-color-settings-arent-working-when-using-shopkit-1782139769133.yaml
@@ -1,0 +1,5 @@
+significance: patch
+type: feature
+entry: "`kadence_blocks_render_footer_css` filter to control whether late,
+  body-rendered block CSS is output in the footer."
+timestamp: 2026-06-22T14:49:29.132Z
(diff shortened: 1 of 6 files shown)


gilbert-hernandez (changes requested):
For all new non-test methods, can we add `@since TBD` in the docblocks? Also, there is a new composer command for the changelogs. Use `composer run changelog:add` to add this fix as a patch. It should generate a yaml file that should now be committed into Kadence Blocks. Thanks!

PS, the CI tests should be fixed when illumise is merged to master and gets updated here.

gilbert-hernandez (approved):
LGTM.

PR #1073, opened 2026-06-29
feat: multi-set authoring (store + REST)

d4mation:
🎫 [SOFT-3583 — Multi-set authoring (store + REST)](https://linear.app/nexcess/issue/SOFT-3583/multi-set-authoring-store-rest)

First ticket of the multi-palette / token-set epic ([SOFT-3582](https://linear.app/nexcess/issue/SOFT-3582/multi-palette-token-set-support)). The storage layer was already slug-keyed; this exposes create / list / delete and lifts the REST single-set restriction so the module can hold more than one token set.

- `Token_Store`: adds `list_stores()`, `exists()`, `delete()`, and a `deleted` action. The default set is guarded from removal in the store — deleting it clears its overrides to baseline; any other set's row is dropped.
- `Token_History_Store::forget()`, wired to the `deleted` action so a removed set's history is purged with no orphaned snapshots.
- `Documents_Controller`: lists every set (the default is always present), reads / creates / updates / deletes any valid slug, and returns a WordPress-style `{ deleted, previous }` response. The single-set 404 / 422 rejections are gone.

### Checklist

### DIFF
--- a/tests/wpunit/Resources/Design_Tokens/Token_StoreTest.php
+++ b/tests/wpunit/Resources/Design_Tokens/Token_StoreTest.php
@@ -288,10 +288,20 @@ final class Token_StoreTest extends TestCase {
 		$this->assertSame( 0, $history->count() );
 	}
 
+	/**
+	 * With no rows stored, listing the sets yields an empty array.
+	 *
+	 * @return void
+	 */
 	public function testListStoresIsEmptyWhenNoRowsExist(): void {
 		$this->assertSame( [], $this->store->list_stores() );
 	}
 
+	/**
+	 * Listing returns one slug/title/version entry per stored set, ordered by slug ascending.
+	 *
+	 * @return void
+	 */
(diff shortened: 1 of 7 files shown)


tests/wpunit/Resources/Design_Tokens/Token_StoreTest.php
pauloiankoski, line 1: `AGENTS.md` is explicit:

> "Every test method has complete PHPDoc … Methods with no parameters still get a docblock with just `@return void`."

Should we strictly follow that, or just ignore it?

d4mation: I'm guessing it got skipped because the other tests in this file didn't have them :/ I'll fix that for these new ones. Good catch!

I've fixed this and updated `AGENTS.md` to hopefully prevent the issue in the future.

83d3123eea291a1d568e0fe4e1681e82b734f42d

pauloiankoski (approved):

PR #1075, opened 2026-06-29
feat: projection honors the active set

d4mation:
🎫  https://linear.app/nexcess/issue/SOFT-3585/projection-honors-the-active-set

> **Depends on #1074** (active-set selection). This is stacked on `soft-3584-active-set-selection`; merge #1074 first.

Make projection render the active token set instead of always the default. Each projector reads the active-set pointer at build time (`Active_Set_Store::get()`) and threads the resolved slug into the version lookup and the resolve/build call, replacing the hardwired `Token_Store::default_slug()` reads. Switching the active set then changes what is projected to the front end and editor, with cache invalidation falling out of the per-set version hash (no new change-signal wiring needed). The Native projector is untouched — it rides the Variant builder's now-active-set-aware slot retarget.

The test consolidation (the misnamed `Projection_ProviderTest` moved into a correctly named `Css_Var/ProjectorTest`) is isolated in its own commit ahead of the feature commit.

### Checklist

### DIFF
--- a/tests/wpunit/Resources/Design_Tokens/Projection/Css_Var/ProjectorTest.php
+++ b/tests/wpunit/Resources/Design_Tokens/Projection/Css_Var/ProjectorTest.php
@@ -1,7 +1,7 @@
 <?php declare( strict_types=1 );
 // cspell:ignore palette pagenow .
 
-namespace Tests\wpunit\Resources\Design_Tokens\Projection;
+namespace Tests\wpunit\Resources\Design_Tokens\Projection\Css_Var;
 
 use KadenceWP\KadenceBlocks\Design_Tokens\Projection\Css_Var\Projector;
 use KadenceWP\KadenceBlocks\Design_Tokens\Projection\Scope;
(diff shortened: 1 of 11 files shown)


tests/wpunit/Resources/Design_Tokens/Projection/Projection_ProviderTest.php
d4mation, line 1: It must have been too many changes for it to show as a "move", but this was moved to `tests/wpunit/Resources/Design_Tokens/Projection/Css_Var/ProjectorTest.php` so it was named correctly. New tests were then added afterwards.

tests/wpunit/Resources/Design_Tokens/Projection/Css_Var/ProjectorTest.php
d4mation, line 1: Contains new tests, plus the ones originally in `tests/wpunit/Resources/Design_Tokens/Projection/Projection_ProviderTest.php`. 

The new test methods can be more directly viewed here: da353c3c8defa8634e72c81fd67d98ddd39d0a76

pauloiankoski (approved):
Approved. Let's just ensure we are not missing some `@since TBD` tags.

PR #1077, opened 2026-06-29
feat(design-tokens): preserve alias var() indirection in the css-var projection

d4mation:
🎫 https://linear.app/nexcess/issue/SOFT-3586/preserve-alias-var-indirection-in-the-css-var-projection

📹 https://www.loom.com/share/8210a55d94cf46d08bd78077a520efa2

> [!IMPORTANT]
> Merge order: the PR for `soft-3585-projection-honors-the-active-set` (SOFT-3585) should be merged first — this branch is based on it.

The resolver flattened every alias to a literal before any CSS existed, so an alias-valued token emitted its resolved hex instead of pointing at the token it referenced. This preserves the indirection in the css-var projection so editing a referenced token updates dependents live, with no server re-resolve.

- A reference-valued token emits `--kb-token--<id>: var(--kb-token--<target>);`; a composite (shadow/typography) keeps a `var()` for any aliased field (e.g. `0px 2px 8px 0px var(--kb-token--primitive--color--ink)`).
- The decision is a function of the effective post-override value. A semantic token that has been explicitly set to a literal value is output as that literal, not as a `var()` reference — the indirection applies only while the token is still an alias. The hex then lives once, at the leaf.
- The literal maps stay literal for host-publishing surfaces (`kadence_blocks_colors`, `--global-palette*`, theme.json presets, REST preview). `Resolved_Tokens` now also exposes a var-preserving `projected_vars()` map plus the per-token alias `target()`.
- The variant projection chains too, and `Variant_Resolver` adopts the same literal/projected split with the **var-preserving form as the default**. `resolve()` now returns the projected values, so a variant token var (`--kb-token--variant--<block>--<variant>--<property>`) reads `var(--kb-token--<target>)` and chains variant to semantic to primitive instead of flattening to a hex. The flattened literal is the explicitly named opt-in, `resolve_literal()`, for the surfaces that cannot consume a `var()` chain: block-attribute presets (an editor control reads the attribute back), the `Block_Default_Css` `var(--token, <fallback>)` value, and the editor variant-catalog feed (swatches). `resolve_default()` delegates to `resolve_literal()` and documents why it must stay concrete.

### Checklist


tests/wpunit/Resources/Design_Tokens/Projection/Css_Var/Css_BuilderTest.php
```diff
@@ -25,8 +25,8 @@ private function builder(): Css_Builder {
 		return new Css_Builder( $this->registry );
 	}
 
-	private function resolved( array $by_id = [], array $by_var = [] ): Resolved_Tokens {
-		return new Resolved_Tokens( $by_id, $by_var );
+	private function resolved( array $by_id = [], array $by_var = [], array $by_id_target = [], array $by_var_target = [] ): Resolved_Tokens {
+		return new Resolved_Tokens( $by_id, $by_var, $by_id_target, $by_var_target );
```
pauloiankoski: Everything looks fine here. The only thing to note is that we are naming things differently. While here we are calling `$by_id_target`, the `Resolved_Tokens` constructor names it as `$by_var_projected`.

d4mation: Good catch! [e5febd6](https://github.com/stellarwp/kadence-blocks/pull/1077/commits/e5febd617b3d97484f0c153bfa0b1b4913f7de59)

pauloiankoski (approved):

PR #1081, opened 2026-07-01
[DTM] SOFT-3574 [2/11] Add User_Primitive_Document_Validator

pauloiankoski:
## Summary
- Adds `User_Primitive_Validation_Error`, a value object describing a single invariant violation.
- Adds `User_Primitive_Document_Validator`, which enforces the tree/envelope invariant for user-created primitives: every envelope entry has a matching valid `color` tree leaf with a literal (non-alias) `$value`, a non-empty label, and no collision with a baseline or system-registered token; every custom tree leaf under `primitive.color.custom.*` has a matching envelope entry.
- Adds `Token_Definition::is_user_created()` / `from_user_primitive()` groundwork needed by the validator's system-collision check (the registry-facing lifecycle methods land in the next phase).

## Test plan
- [x] `User_Primitive_Document_ValidatorTest` covers valid documents, orphaned tree leaves, orphaned envelope entries, namespace/type/alias/label violations, and baseline/system/user-created collision handling.
- [x] PHPStan clean
- [x] cspell clean


includes/resources/Design_Tokens/Document/User_Primitive_Document_Validator.php
```diff
@@ -0,0 +1,234 @@
...
+
+			return $errors;
+		}
+
+		$type = $leaf[ Token_Type::get_type_key() ] ?? null;
+
+		if ( $type !== Token_Type::get_type_color() ) {
+			$errors[] = new User_Primitive_Validation_Error(
+				$id,
+				sprintf( 'User primitive "%s" tree leaf has $type "%s"; only "color" is allowed in phase one.', $id, Cast::to_string( $type ) )
```
d4mation:
```suggestion
				sprintf( 'User primitive "%s" tree leaf has $type "%s"; only "color" is allowed at this time.', $id, Cast::to_string( $type ) )
```

includes/resources/Design_Tokens/Document/User_Primitive_Document_Validator.php
```diff
@@ -0,0 +1,234 @@
...
+			);
+		}
+
+		if ( ! array_key_exists( '$value', $leaf ) ) {
+			$errors[] = new User_Primitive_Validation_Error( $id, sprintf( 'User primitive "%s" tree leaf is missing $value.', $id ) );
+		} elseif ( Alias::is_alias( $leaf['$value'] ) ) {
+			$errors[] = new User_Primitive_Validation_Error(
+				$id,
+				sprintf( 'User primitive "%s" $value must be a literal; aliases are not allowed on user primitives in phase one.', $id )
```
d4mation:
```suggestion
				sprintf( 'User primitive "%s" $value must be a literal; aliases are not allowed on user primitives at this time.', $id )
```

includes/resources/Design_Tokens/Document/User_Primitive_Document_Validator.php
```diff
@@ -0,0 +1,234 @@
...
+		if ( $type !== Token_Type::get_type_color() ) {
+			$errors[] = new User_Primitive_Validation_Error(
+				$id,
+				sprintf( 'User primitive "%s" tree leaf has $type "%s"; only "color" is allowed in phase one.', $id, Cast::to_string( $type ) )
+			);
+		}
+
+		if ( ! array_key_exists( '$value', $leaf ) ) {
+			$errors[] = new User_Primitive_Validation_Error( $id, sprintf( 'User primitive "%s" tree leaf is missing $value.', $id ) );
+		} elseif ( Alias::is_alias( $leaf['$value'] ) ) {
```
d4mation, line 143: See my comment on https://github.com/stellarwp/kadence-blocks/pull/1082 about this. It looks like there is a bug related to this check when directly setting a Token.

pauloiankoski: `Documents_Controller::set_token()` does not use `Document_Write_Pipeline::validate_and_save()`. It uses its own smaller `validate_and_save()`. That method only checks DTCG grammar and does a resolver dry-run. It never calls `User_Primitive_Document_Validator` or `Token_Reference_Policy`.

This PR does not touch Documents_Controller. Neither does #1082. So there is no place in either diff to fix this.

I'm tracking the fix in #1087 instead. That PR already adds write guards to `Documents_Controller`. I think we can leave this PR as-is.

d4mation (approved):

PR #1082, opened 2026-07-01
[DTM] SOFT-3574 [3/11] Add Token_Reference and Token_Reference_Policy alias scanner

pauloiankoski:
## Summary
- Adds `Token_Reference`, a value object describing a single alias reference (kind, referencing path, whether the phase-1 cascade supports it).
- Adds `Token_Reference_Policy`, a full scanner that recursively walks the primitive/semantic token layers and every `$extensions` preset section for aliases pointing at a given token id.
- The scanner classifies a direct `$value` alias in the semantic layer as the only phase-1 "supported" (actionable) reference kind; everything else (primitive-layer aliasing, composite sub-field aliasing at any nesting depth, extension preset token maps) is unsupported and blocks deletion.
- Composite/nested `$value` scanning is a generic recursive walk rather than a fixed field-list lookup, so a future composite or list-shaped sub-field can't silently bypass detection.

## Test plan
- [x] `Token_Reference_PolicyTest` covers no-references, semantic direct alias (supported), primitive direct alias (unsupported), composite field alias, a nested composite field alias two levels deep, extension preset token map alias, `all_supported()`, `$`-prefixed key skipping, `$value: null` non-match, and multiple references to the same id.
- [x] PHPStan clean
- [x] cspell clean

### DIFF
--- a/includes/resources/Design_Tokens/Document/Token_Reference.php
+++ b/includes/resources/Design_Tokens/Document/Token_Reference.php
@@ -1,0 +1,105 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Document;
+
+/**
+ * A single alias reference from a stored token to another token id.
+ *
+ * @since TBD
+ */
+final class Token_Reference {
+
+	/**
+	 * Semantic layer override with a direct $value alias.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const KIND_SEMANTIC_OVERRIDE = 'semantic_override';
+
+	/**
+	 * Alias found inside a composite $value sub-field (e.g. shadow.color).
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const KIND_COMPOSITE_FIELD = 'composite_field';
+
+	/**
+	 * Alias found inside $extensions.variants or $extensions.foundationPresets.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const KIND_EXTENSION = 'extension';
+
+	/**
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public static function get_kind_semantic_override(): string {
+		return self::KIND_SEMANTIC_OVERRIDE;
+	}
+
+	/**
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public static function get_kind_composite_field(): string {
+		return self::KIND_COMPOSITE_FIELD;
+	}
+
+	/**
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public static function get_kind_extension(): string {
+		return self::KIND_EXTENSION;
+	}
+
+	/**
+	 * The reference kind. One of the KIND_* values.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public string $kind;
+
+	/**
+	 * The canonical dot-path of the token holding the reference.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public string $path;
+
+	/**
+	 * Whether the phase-1 cascade can automatically handle this reference.
+	 *
+	 * @since TBD
+	 *
+	 * @var bool
+	 */
+	public bool $supported;
+
+	/**
+	 * @since TBD
+	 *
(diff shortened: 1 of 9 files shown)


includes/resources/Design_Tokens/Document/Token_Reference.php
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	/**
+	 * The reference kind. One of the KIND_* values.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public string $kind;
```
d4mation: Methods should be below all class properties and consts.

includes/resources/Design_Tokens/Document/Token_Reference_Policy.php
```diff
@@ -0,0 +1,243 @@
...
+
+			if ( is_array( $value ) ) {
+				$this->scan_composite_value(
+					$value,
+					$path . '.$value',
+					$alias,
+					Token_Reference::get_kind_composite_field(),
+					$references
+				);
+			}
```
d4mation, line 131: There's a possible edge case here. It wouldn't be possible with the UI as it is, but if a Custom Primitive were updated to point to _another_ Primitive, it will think it is a `composite_field` when it isn't. The logic saying that it cannot be deleted works as expected, but it gets flagged as the wrong `kind`.

However, since guards against this seem to exist in `Document_Write_Pipeline::validate_and_save()`, the fact that you can do this _at all_ may be a bug.

Reproduction case:

```js
( async () => {
	const base = '/kb-design-tokens/v1/documents/default';

	// 1) Read the current store version (create is version-guarded) and grab a real
	//    baseline color primitive id to use as the alias holder (any non-custom primitive.color.*).
	const resolved = await wp.apiFetch( { path: `${ base }/resolved` } );
	const holderId = Object.keys( resolved.by_id ).find(
		( id ) => id.startsWith( 'primitive.color.' ) && ! id.includes( '.custom.' )
	);
	console.log( 'version', resolved.version, 'holder', holderId );

	// 2) Create the TARGET user primitive `blue` with a LITERAL value.
	//    (aliases are forbidden on user primitives, so this must be a hex.)
	const created = await wp.apiFetch( {
		path: `${ base }/user-primitives`,
		method: 'POST',
		data: { id: 'blue', '$type': 'color', '$value': '#3182ce', label: 'Blue', version: resolved.version },
	} );
	console.log( 'created', created ); // -> primitive.color.custom.blue

	// 3) Through the set_token side door, override a PRIMITIVE-LAYER token so its
	//    ENTIRE value is a direct alias to blue. No reserved-path guard, no reference
	//    guard, unconditional save -> no version needed.
	await wp.apiFetch( {
		path: `${ base }/tokens/${ holderId }`,
		method: 'PUT',
		data: { '$type': 'color', '$value': '{primitive.color.custom.blue}' },
	} );

	// 4) Preview references for blue. The holder is a whole-value, primitive-layer
	//    reference -> should be "direct primitive alias", but comes back mislabeled.
	const refs = await wp.apiFetch( {
		path: `${ base }/user-primitives/primitive.color.custom.blue/references`,
	} );
	console.log( 'references', refs.references );
	// Expect: [{ kind: "composite_field", path: "<holderId>", supported: false, action: "unsupported" }]
	//                  ^^^^^^^^^^^^^^^^  the bug: not composite, not a sub-field, and in the primitive layer
} )();
  ```

pauloiankoski: You're right, this is not a bug in `Token_Reference_Policy`. It's a bug in `set_token`. I confirmed `Documents_Controller::set_token()` uses its own small `validate_and_save()`, not the one in `Document_Write_Pipeline`. So it skips `User_Primitive_Document_Validator`, skips this reference check, and has no version check either.

I'm moving the fix to #1087, which already adds write guards to `Documents_Controller`.

Once that guard is in place, we may still want a new kind on `Token_Reference` for "whole-value alias outside the semantic layer," so this stops being labeled `composite_field`. That's a small follow-up, not a blocker for this PR.

includes/resources/Design_Tokens/Document/Token_Reference.php
pauloiankoski: Resolved in 0468d3a427f34965ce11d8ba791fcd1ca74e5a1c

d4mation (approved):

PR #1084, opened 2026-07-01
[DTM] SOFT-3574 [5/11] Add User_Primitive_Registrar and wire registry sync

pauloiankoski:
## Summary
- Adds `User_Primitive_Registrar`, which reads the committed active-set document, derives each user primitive's `$type` from its tree leaf (never trusting the envelope's provenance metadata for type), and registers/deregisters against `Token_Registry` accordingly. Corrupt entries (orphaned envelope, missing/unknown `$type`, system-token id collision) are logged and skipped rather than failing boot.
- Wires the registrar into `Registry\Provider`: boot registration on `init` (after `register_declarations`, same priority), a priority-0 subscriber on `Token_Store::changed_action()` that only re-syncs when the changed slug is the active one, and a priority-0 subscriber on `Active_Set_Store::changed_action()` that always re-syncs. Both run ahead of the default-priority projectors/caches that react to the same actions.

## Test plan
- [x] `User_Primitive_RegistrarTest` covers empty-active-set boot, valid registration with label/type sourced correctly, inactive-set entries being skipped, active-set-change re-sync, active vs. inactive slug writes, and each corrupt-entry path (orphaned envelope, missing type, unknown type, system collision), plus `sync()` idempotency and re-reading committed state.
- [x] PHPStan clean
- [x] cspell clean

### DIFF
--- a/includes/resources/Design_Tokens/Registry/User_Primitive_Registrar.php
+++ b/includes/resources/Design_Tokens/Registry/User_Primitive_Registrar.php
@@ -1,0 +1,186 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Registry;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Database\Active_Set_Store;
+use KadenceWP\KadenceBlocks\Design_Tokens\Database\Token_Store;
+use KadenceWP\KadenceBlocks\Design_Tokens\Document\User_Primitive_Index;
+use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Token_Type;
+use KadenceWP\KadenceBlocks\Psr\Log\LoggerInterface;
+
+/**
+ * Reads user-created primitive definitions from the store and registers them into Token_Registry.
+ * Also the authoritative re-sync point after each store write (see Registry\Provider).
+ *
+ * @since TBD
+ */
+final class User_Primitive_Registrar {
+
+	/**
+	 * @since TBD
+	 *
+	 * @var Token_Store
+	 */
+	private Token_Store $store;
+
+	/**
+	 * @since TBD
+	 *
+	 * @var Active_Set_Store
+	 */
+	private Active_Set_Store $active;
+
+	/**
+	 * @since TBD
+	 *
+	 * @var Token_Registry
+	 */
+	private Token_Registry $registry;
+
+	/**
+	 * @since TBD
+	 *
+	 * @var User_Primitive_Index
+	 */
+	private User_Primitive_Index $index;
+
+	/**
+	 * @since TBD
+	 *
+	 * @var LoggerInterface
+	 */
+	private LoggerInterface $logger;
+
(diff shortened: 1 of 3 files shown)


includes/resources/Design_Tokens/Registry/User_Primitive_Registrar.php
```diff
@@ -0,0 +1,186 @@
...
+	/**
+	 * @since TBD
+	 *
+	 * @param array<string, mixed> $document
+	 * @param string               $id
+	 *
+	 * @return string|null
+	 */
+	private function type_from_tree( array $document, string $id ): ?string {
+		$node = $document;
```
d4mation: Why do we need `$node = $document`? Couldn't the definition of the method just name it `$node` if we want to have it read that way? You've got two copies of `$document` in memory when doing this.

includes/resources/Design_Tokens/Registry/User_Primitive_Registrar.php
```diff
@@ -0,0 +1,186 @@
...
+		return is_string( $type ) && $type !== '' ? $type : null;
+	}
+
+	/**
+	 * @since TBD
+	 *
+	 * @return array<string, mixed>
+	 */
+	private function load_document(): array {
+		$raw = $this->store->get_document( $this->active->get() );
```
d4mation: What if different Token Sets have different User Primitives?

Token Set A:
    - custom-primitive-b: `#ffffff`
Token Set B:
    - custom-primitive-c: `#000000`
    
Since https://linear.app/nexcess/issue/SOFT-3587/simultaneous-multi-set-emission-active-alias-layer will emit _all_ Tokens in the rendered page to allow switching between them, we will need to have all possible User Primitives registered, right?

includes/resources/Design_Tokens/Registry/User_Primitive_Registrar.php
pauloiankoski: Resolved in fded8d715f9e97fa303f3cbe449d66311744fa41

includes/resources/Design_Tokens/Registry/User_Primitive_Registrar.php
pauloiankoski: `sync()` now loads every stored set via `Token_Store::list_stores()`, not just the active one, so each set’s user primitives remain registered regardless of which set is active.

One thing to note: `Token_Registry` uses a flat ID space with no per-set namespacing. If two sets define the same custom primitive ID, the default set’s definition takes precedence, and the other is skipped with a warning in the log. There is no silent overwrite, but only one definition is visible in the registry at a time for a conflicting ID.

Committed in 542e69542.

d4mation: Hmm, that may be an issue 🤔 We'll need to be sure to investigate it and whether it could cause issues in-practice. I'm going to go get the Multi-set PRs merged so that we can account for whatever we may need :)

d4mation (approved):

PR #1085, opened 2026-07-01
[DTM] SOFT-3574 [6/11] Add save_document_conditional and delete_document_conditional

pauloiankoski:
## Summary
- Adds `Token_Store::save_document_conditional()`: a compare-and-swap write. First write on a slug requires an empty `expected_version` and no existing row; subsequent writes require the caller's `expected_version` to match the stored version. The conditional `UPDATE ... WHERE slug = ? AND version = ?` closes the gap between the initial read and the write, so a concurrent write in between produces a definitive `false` (version conflict) rather than a silent overwrite. The first-write INSERT race (another request creating the same slug in between) is caught specifically and also returns `false`, not a generic failure.
- Adds `Token_Store::delete_document_conditional()` with the same CAS discipline for named-set deletion; deleting the default set conditionally is rejected (`false`) since the default set is reset rather than removed.
- `list_stores()`, `exists()`, unconditional `save_document()`/`delete()`, history archiving, and the `changed`/`superseded`/`deleted` action hooks are all unmodified — the new methods are pure additions alongside them.

## Test plan
- [x] `Token_StoreTest` adds first-write/update success and version-mismatch cases for both conditional methods, a concurrent-stale-update race case, version-change-on-success, and action-firing-only-on-success assertions.
- [x] PHPStan clean
- [x] cspell clean

### DIFF
--- a/includes/resources/Design_Tokens/Database/Token_Store.php
+++ b/includes/resources/Design_Tokens/Database/Token_Store.php
@@ -205,6 +205,147 @@ final class Token_Store extends Query {
 	}
 
 	/**
+	 * Write a document only when the stored version matches expected_version.
+	 *
+	 * First write: expected_version must be an empty string; the row must not yet exist.
+	 * Subsequent writes: expected_version must match the current stored version.
+	 *
+	 * Returns true on success, false on version mismatch (caller maps to 409).
+	 * Throws DatabaseQueryException on a real write failure.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $document         The raw DTCG JSON to persist.
+	 * @param string $expected_version The version the caller last read. Empty for first write.
+	 * @param string $slug             The token set slug.
+	 * @param string $title            Optional label; left untouched when empty.
+	 *
+	 * @return bool True on success; false when the version does not match.
+	 *
+	 * @throws DatabaseQueryException If the write fails.
+	 */
+	public function save_document_conditional(
+		string $document,
+		string $expected_version,
+		string $slug = self::DEFAULT_SLUG,
+		string $title = ''
+	): bool {
+		$previous = $this->qb()
+						->where( 'slug', $slug )
+						->get( ARRAY_A );
+
+		if ( ! is_array( $previous ) ) {
+			// First write — expected_version must be empty.
+			if ( $expected_version !== '' ) {
+				return false;
+			}
+
+			$data = [
+				'slug'       => $slug,
+				'document'   => $document,
+				'version'    => $this->hash_document( $document ),
+				'updated_at' => current_time( 'mysql', true ),
+			];
+
+			if ( $title !== '' ) {
+				$data['title'] = $title;
+			}
+
+			try {
+				$this->qb()->insert( $data );
+			} catch ( DatabaseQueryException $e ) {
+				// Duplicate-key on concurrent first write → conflict, not 500.
+				return false;
+			}
+
+			$this->changed( $slug );
+
+			return true;
+		}
+
+		// Subsequent write — must match.
+		if ( (string) ( $previous['version'] ?? '' ) !== $expected_version ) {
+			return false;
+		}
+
+		$new_version = $this->hash_document( $document );
+		$data        = [
+			'document'   => $document,
+			'version'    => $new_version,
+			'updated_at' => current_time( 'mysql', true ),
+		];
+
+		if ( $title !== '' ) {
+			$data['title'] = $title;
+		}
+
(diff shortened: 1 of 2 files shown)


includes/resources/Design_Tokens/Database/Token_Store.php
```diff
@@ -204,6 +204,147 @@ public function save_document( string $document, string $slug = self::DEFAULT_SL
...
+				return false;
+			}
+
+			$this->changed( $slug );
+
+			return true;
+		}
+
+		// Subsequent write — must match.
+		if ( (string) ( $previous['version'] ?? '' ) !== $expected_version ) {
```
d4mation: We can use the `Cast` utility for casting here.

includes/resources/Design_Tokens/Database/Token_Store.php
```diff
@@ -204,6 +204,147 @@ public function save_document( string $document, string $slug = self::DEFAULT_SL
...
+
+		// Zero affected rows means the version changed concurrently.
+		if ( (int) $affected === 0 ) {
+			return false;
+		}
+
+		$this->superseded( $slug, (string) ( $previous['document'] ?? '' ), $expected_version );
+		$this->changed( $slug );
+
+		return true;
```
d4mation: Maybe we should just call `save_document()` once we get to this point? Maintaining two copies of this will get annoying and may cause future issues.

includes/resources/Design_Tokens/Database/Token_Store.php
```diff
@@ -204,6 +204,147 @@ public function save_document( string $document, string $slug = self::DEFAULT_SL
...
+						->delete();
+
+		// Zero affected rows means the version changed concurrently.
+		if ( (int) $affected === 0 ) {
+			return false;
+		}
+
+		$this->deleted( $slug );
+
+		return true;
```
d4mation, line 317: Same here. Maybe we should just call `delete()`?

includes/resources/Design_Tokens/Database/Token_Store.php
pauloiankoski: Can't fully delegate to `save_document()` here without losing the guarantee this method exists for: `save_document()` is an unconditional upsert (SELECT-then-INSERT/UPDATE, no version guard), while this branch's whole point is the atomic `WHERE slug = ? AND version = ?` UPDATE. Between the version check above and a call to `save_document()`, a concurrent writer could update the row and `save_document()` would silently clobber it — exactly the lost-update race this endpoint prevents.

The duplication concern is fair at a narrower scope though: the `document`/`version`/`updated_at`/`title` array was built three separate times (`save_document()`, the first-write branch, and this branch). Extracted that into a private `build_document_data()` helper (d088c6264ab765d02aee8d363cbfb3272a47c705) used by all three call sites, so the CAS-guarded UPDATE and the plain upsert stay distinct but no longer duplicate the data-shaping logic.

includes/resources/Design_Tokens/Database/Token_Store.php
pauloiankoski: Same reasoning as above: `delete()` is unconditional (checks the row exists, then deletes with no version guard), while this branch's atomic `WHERE slug = ? AND version = ?` delete is the entire point of `delete_document_conditional()`. Calling `delete()` after the version check reopens the same race — a concurrent write could land between the check and the call, and `delete()` would remove the row regardless of whether it still matches `expected_version`, silently discarding whatever the concurrent writer just wrote.

Unlike the `save_document()` case, there isn't shared data-building logic here to extract — `delete()` also carries default-slug special-casing (it resets the default set to baseline instead of removing its row) that this method deliberately refuses to do. So I'd leave these as separate implementations rather than introduce a helper that would need its own branching to reconcile the two.

includes/resources/Design_Tokens/Database/Token_Store.php
pauloiankoski: Resolved in 6290b991c9aab0a8a7534b83eb1e2dcb97bf70cd

includes/resources/Design_Tokens/Database/Token_Store.php
d4mation: The change with `build_document_data()` solves my concern here :) Thanks!

includes/resources/Design_Tokens/Database/Token_Store.php
d4mation: The change with `build_document_data()` solves my concern here :) Thanks!

d4mation (approved):

PR #1086, opened 2026-07-01
[DTM] SOFT-3574 [7/11] Add Document_Write_Pipeline

pauloiankoski:
## Summary
- Adds `Document_Write_Pipeline`, a standalone shared write service for `User_Primitives_Controller`: load/decode, slug/version guards, and `validate_and_save()` (user-primitive invariant check → reference-policy check → DTCG schema validation → dry-run alias resolution → conditional persist via `Token_Store::save_document_conditional()`).
- Mirrors `Documents_Controller`'s existing private validate → dry-run-resolve → encode → persist pattern rather than extracting from it — `Documents_Controller`'s private `validate_and_save`, `persist`, `read_stored_document`, and `node_at` are untouched.
- Registers `Document_Write_Pipeline` as a container singleton in `Rest/V1/Provider`.

## Test plan
- [x] `Document_Write_PipelineTest` covers slug/version guard branches, each validation-failure path (invariant, unsupported reference, DTCG schema, alias cycle) returning 422, a version conflict returning 409, and both first-write (201) and update (200) success paths.
- [x] PHPStan clean
- [x] cspell clean

### DIFF
--- a/includes/resources/Design_Tokens/Rest/V1/Document_Write_Pipeline.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Document_Write_Pipeline.php
@@ -3,6 +3,7 @@
 namespace KadenceWP\KadenceBlocks\Design_Tokens\Rest\V1;
 
 use KadenceWP\KadenceBlocks\Design_Tokens\Database\Token_Store;
+use KadenceWP\KadenceBlocks\Design_Tokens\Document\Document_Path;
 use KadenceWP\KadenceBlocks\Design_Tokens\Document\Token_Reference_Policy;
 use KadenceWP\KadenceBlocks\Design_Tokens\Document\User_Primitive_Document_Validator;
 use KadenceWP\KadenceBlocks\Design_Tokens\Document\User_Primitive_Index;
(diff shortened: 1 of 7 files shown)


includes/resources/Design_Tokens/Rest/V1/Document_Write_Pipeline.php
```diff
@@ -0,0 +1,358 @@
...
+		foreach ( explode( '.', $path ) as $segment ) {
+			if ( ! is_array( $node ) || ! array_key_exists( $segment, $node ) ) {
+				return null;
+			}
+
+			$node = $node[ $segment ];
+		}
+
+		return is_array( $node ) ? $node : null;
+	}
```
d4mation, line 141: This seems like something that should exist elsewhere? Like maybe in `Token_Store`?

Maybe aspects of `Token_Resolver` would need updating to utilize this as well.

pauloiankoski: Pulled the dot-path walk into a new `Document_Path::node_at()` static (in the `Document` namespace, since it's a pure document-array operation, not DB access) and wired it into `Token_Resolver::lookup()`. It turned out to already be duplicated in two other places too — `Documents_Controller::node_at` and `User_Primitive_Document_Validator::node_at` — both now delegate to the same helper, and their private copies are gone.

`Document_Write_Pipeline::node_at()` stays as a thin public delegate rather than being removed, since `User_Primitives_Controller` (further down the stack) calls it as part of the pipeline's public surface.

Propagated this through `phase-8` → `phase-10` → `phase-11` via merge (no force-push) so the whole stack shares one implementation.

d4mation (approved):

PR #1087, opened 2026-07-01
[DTM] SOFT-3574 [8/11] Add write guards to Documents_Controller for reserved namespaces

pauloiankoski:
## Summary
- Injects `User_Primitive_Index` into `Documents_Controller` and adds three write guards so the generic multi-set document endpoints can't be used to bypass user-primitive lifecycle rules:
  - **PATCH** inspects the incoming partial payload (not the merged/stored candidate) for reserved paths — writes touching `primitive.<type>.custom.*` or the `userPrimitives` extensions section are rejected with 403, while unreserved partials succeed even when the stored document already has user primitives.
  - **PUT** is rejected with 409 while the stored document has any user primitive, since a full replace has no way to express "preserve these"; a PUT against a document with none proceeds normally.
  - **Single-token DELETE** rejects deleting a reserved custom path directly; deleting a normal (non-custom) token is unaffected.
- All three guards run before any read/merge/persist step, so a rejected request never touches storage.

## Test plan
- [x] `DocumentsControllerTest` adds the 8 guard cases (PATCH shallow/deep custom path, PATCH reserved extension, PATCH unreserved-with-existing-user-primitives, PUT-with-user-primitives, PUT-with-clean-doc, DELETE-reserved-path, DELETE-non-custom-path), and the full existing suite (47 tests) still passes, confirming no regression to the pre-existing multi-set document API.
- [x] PHPStan clean
- [x] cspell clean

### DIFF
--- a/includes/resources/Design_Tokens/Document/Reserved_Namespace.php
+++ b/includes/resources/Design_Tokens/Document/Reserved_Namespace.php
@@ -5,12 +5,14 @@ namespace KadenceWP\KadenceBlocks\Design_Tokens\Document;
 use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Token_Type;
 
 /**
- * The single source of truth for the user-primitive reserved namespace: primitive.color.custom.*.
+ * The single source of truth for the user-primitive reserved namespace: primitive.<type>.custom.*.
  *
  * Every write surface that must recognize this namespace (the user-primitive document invariant, the
  * user-primitives controller's id validation, and the generic documents controller's write guards)
- * defers to this class instead of re-deriving the shape on its own, so the phase-1 scope (color only)
- * cannot drift between call sites.
+ * defers to this class instead of re-deriving the shape on its own, so the set of reserved <type>
+ * segments cannot drift between call sites. The <type> segment is validated against every registered
+ * Token_Type rather than hardcoding "color", so a future custom primitive under an existing $type (e.g.
+ * a gap or radius primitive, both "dimension") is guarded automatically.
  *
  * @since TBD
  */
@@ -35,7 +37,7 @@ final class Reserved_Namespace {
 	private const SEGMENT = 'custom';
 
 	/**
-	 * Whether a canonical dot-path id names a user-primitive leaf: primitive.color.custom.<slug>.
+	 * Whether a canonical dot-path id names a user-primitive leaf: primitive.<type>.custom.<slug>.
 	 *
 	 * @since TBD
 	 *
@@ -45,7 +47,7 @@ final class Reserved_Namespace {
 	 */
 	public static function is_reserved_id( string $id ): bool {
 		return (bool) preg_match(
-			'/^' . self::LAYER . '\.' . Token_Type::get_type_color() . '\.' . self::SEGMENT . '\.[a-z0-9]+(?:-[a-z0-9]+)*$/',
+			'/^' . self::LAYER . '\.(?:' . implode( '|', Token_Type::all() ) . ')\.' . self::SEGMENT . '\.[a-z0-9]+(?:-[a-z0-9]+)*$/',
 			$id
 		);
 	}
@@ -67,7 +69,7 @@ final class Reserved_Namespace {
 
 		return count( $segments ) >= 3
 			&& $segments[0] === self::LAYER
-			&& $segments[1] === Token_Type::get_type_color()
+			&& in_array( $segments[1], Token_Type::all(), true )
 			&& $segments[2] === self::SEGMENT;
 	}
 
--- a/includes/resources/Design_Tokens/Document/Reserved_Namespace.php
+++ b/includes/resources/Design_Tokens/Document/Reserved_Namespace.php
@@ -1,0 +1,104 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Document;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Token_Type;
+
+/**
+ * The single source of truth for the user-primitive reserved namespace: primitive.color.custom.*.
+ *
+ * Every write surface that must recognize this namespace (the user-primitive document invariant, the
+ * user-primitives controller's id validation, and the generic documents controller's write guards)
+ * defers to this class instead of re-deriving the shape on its own, so the phase-1 scope (color only)
+ * cannot drift between call sites.
+ *
+ * @since TBD
+ */
+final class Reserved_Namespace {
+
+	/**
+	 * The token layer the reserved namespace lives under.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const LAYER = 'primitive';
+
+	/**
+	 * The tree segment, directly under the reserved $type, that holds user-created primitives.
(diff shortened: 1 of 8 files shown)


includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
```diff
@@ -607,8 +640,19 @@ public function set_token( $request ) {
...
+		if ( $this->is_reserved_custom_path( $path ) ) {
+			return new WP_Error(
+				'rest_design_tokens_reserved_path',
+				__( 'Use the user-primitives endpoint to delete a custom primitive.', 'kadence-blocks' ),
+				[
+					'status' => WP_Http::FORBIDDEN,
+					'path'   => $path,
+				]
+			);
+		}
```
d4mation: This change means the docblock needs updated, right? It is no longer an idempotent no-op in this particular case.

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
```diff
@@ -469,7 +490,19 @@ public function patch_item( $request ) {
...
+		if ( ! empty( $this->user_primitive_index->all( $stored ) ) ) {
+			return new WP_Error(
+				'rest_design_tokens_put_not_allowed',
+				__( 'Bulk replace is not allowed while user-created primitives exist. Delete them first or use PATCH.', 'kadence-blocks' ),
+				[
+					'status' => WP_Http::CONFLICT,
+					'slug'   => $slug,
+				]
+			);
+		}
```
d4mation, line 535: This also looks like it will require a change to the docblock.

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
```diff
@@ -1164,6 +1208,112 @@ private function prepare_item( string $slug ): array {
...
+	 * @return string[]
+	 */
+	private function collect_reserved_paths_in( array $node, string $prefix ): array {
+		$found    = [];
+		$segments = explode( '.', $prefix );
+
+		// At primitive.<type>.custom depth (3 segments: primitive, type, custom) or deeper: reserved.
+		if ( count( $segments ) >= 3 && $segments[2] === 'custom' ) {
+			return [ $prefix ];
+		}
```
d4mation: This seems fragile. We need a way to have an Object for the Token and determine if it is custom/user-provided instead.

pauloiankoski: What do you think about this approach? 95e1ba1ade20fb7bd6ba30ce36a6000a0afd05a7

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
pauloiankoski: 3b0e92d6f6c6c6ff5e0538e30110737ac395e0e5

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
pauloiankoski: b6b1c2c99be920bb69caf5d7aca24698de4a31be

pauloiankoski:
@d4mation Related to the comments for phase 2 and 3, here's the implementation: 17ecf2a240584167e34ca89062381bc658898adb

includes/resources/Design_Tokens/Document/Reserved_Namespace.php
```diff
@@ -0,0 +1,104 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string $id The canonical dot-path id.
+	 *
+	 * @return bool
+	 */
+	public static function is_reserved_id( string $id ): bool {
+		return (bool) preg_match(
+			'/^' . self::LAYER . '\.' . Token_Type::get_type_color() . '\.' . self::SEGMENT . '\.[a-z0-9]+(?:-[a-z0-9]+)*$/',
```
d4mation: This works for now if we're only doing Custom Color Primitives, but what if we support Custom Gap Primitives? Custom Radius Primitives?

includes/resources/Design_Tokens/Document/Reserved_Namespace.php
```diff
@@ -0,0 +1,104 @@
...
+	 * @param string $path The dot-path to test.
+	 *
+	 * @return bool
+	 */
+	public static function contains_reserved_path( string $path ): bool {
+		$segments = explode( '.', $path );
+
+		return count( $segments ) >= 3
+			&& $segments[0] === self::LAYER
+			&& $segments[1] === Token_Type::get_type_color()
```
d4mation: Same concern here.

includes/resources/Design_Tokens/Document/Reserved_Namespace.php
pauloiankoski: Good catch on both — `is_reserved_id()` and `contains_reserved_path()` were matching the `<type>` segment against `Token_Type::get_type_color()` specifically, so a custom primitive under any other `$type` wouldn't have been recognized as reserved. Fixed in 68b7ce572e6673253970d07b31d9add8db7fe900 by matching against `Token_Type::all()` instead: `is_reserved_id()` now builds its regex alternation from the full type list, and `contains_reserved_path()` uses `in_array()` against it. Note gap/radius primitives would land under `$type` `dimension` (see `baseline.json`'s `primitive.dimension.{gap,radius}`), not a new type of their own, so this already covers those cases without needing per-primitive-kind entries.

includes/resources/Design_Tokens/Document/Reserved_Namespace.php
pauloiankoski: See https://github.com/stellarwp/kadence-blocks/pull/1087#discussion_r3531818052

includes/resources/Design_Tokens/Document/Reserved_Namespace.php
```diff
@@ -0,0 +1,106 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string $id The canonical dot-path id.
+	 *
+	 * @return bool
+	 */
+	public static function is_reserved_id( string $id ): bool {
+		return (bool) preg_match(
+			'/^' . self::LAYER . '\.(?:' . implode( '|', Token_Type::all() ) . ')\.' . self::SEGMENT . '\.[a-z0-9]+(?:-[a-z0-9]+)*$/',
```
d4mation: While it _shouldn't_ be an issue here, I wonder if we should run the implode() call through `preg_quote( $implode, '/' )` to be safe.

d4mation (commented):
One small suggestion, but then I think we're good to merge this one.

includes/resources/Design_Tokens/Document/Reserved_Namespace.php
pauloiankoski: Updated `is_reserved_id()` to run each `Token_Type::all()` value through `preg_quote( $type, '/' )` before imploding with `|`, so the alternation stays correct even if a future type value contains a regex metacharacter or the `/` delimiter. Escaping element-by-element rather than the joined string avoids accidentally escaping the `|` separator itself. Committed in d7a7de0e4.

d4mation (approved):

PR #1089, opened 2026-07-01
[DTM] SOFT-3574 [10/11] Add create_item, delete_item, and rename_item to User_Primitives_Controller

pauloiankoski:
## Summary
- Adds the three mutation endpoints for user-created color primitives:
  - **Create**: client sends a kebab-case slug (`^[a-z0-9]+(?:-[a-z0-9]+)*$`, no dots) plus `$type`/`$value`/optional label; the server constructs the canonical path itself. Rejects non-`color` types, alias `$value`s, and slug collisions with an existing user primitive, baseline token, or system-registered token, all enforced centrally by `User_Primitive_Document_Validator` inside `Document_Write_Pipeline` (not re-implemented in the controller, so it can't be bypassed).
  - **Delete**: client sends the canonical full path; the server derives `$type` from the stored tree and never trusts a client-supplied type. Re-runs `Token_Reference_Policy` at mutation time (not the preview's cached result) and blocks with 422 if any unsupported reference exists. When safe, removes the tree leaf and envelope entry and reverts direct semantic-layer aliases to the deleted id in the same write.
  - **Rename**: moves the tree leaf and envelope entry and rewrites direct `{old_path}` alias strings in the semantic layer to `{new_path}`, in a single store write. Also derives `$type` from the stored tree rather than trusting the client.
- All three mutations go through exactly one `Document_Write_Pipeline::validate_and_save()` → one `save_document_conditional()` call — no multi-step writes, and version/CAS conflicts return 409 rather than silently overwriting.

## Known limitation carried from the design
- Unlike delete, rename does **not** run `Token_Reference_Policy` — it only rewrites direct semantic-layer aliases and leaves any `$extensions`/composite-field reference to the old id dangling (pointing at an id that no longer exists). This matches the locked design decision ("Rename semantics: move tree leaf + rewrite direct alias strings in the semantic layer... single store write") and is covered by `testRenameRewritesAliasesInSemanticLayerOnly`, but it's a materially weaker safety posture than delete's block-on-unsupported-reference behavior. Worth a deliberate call on whether Phase 11's UI should warn on rename the way it presumably will on delete.

## Test plan
- [x] `User_Primitives_ControllerTest` covers create (success, label truncation/derivation, non-color rejection, slug collisions, alias-value rejection, version conflicts, invalid slug formats), delete (success, cascade revert, re-running analysis rather than trusting a stale preview, unsupported-reference 422s, system-primitive 403, unknown-id 404, version conflict, invalid id format), and rename (success, type derivation from tree, semantic-layer-only alias rewrite, baseline collision 409, unknown id 404, version conflict, invalid id format, post-rename document validates cleanly).
- [x] PHPStan clean
- [x] cspell clean


includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
```diff
@@ -181,6 +189,45 @@ public function register_routes(): void {
 				'schema' => [ $this, 'get_references_schema' ],
 			]
 		);
+
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/' . self::SLUG_ROUTE . '/' . self::USER_PRIMITIVES_ROUTE,
+			[
+				[
+					'methods'             => WP_REST_Server::CREATABLE,
```
d4mation: `CREATABLE` only handles `POST`. What about `PUT`? `Documents_Controller` has different logic for these.

includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
```diff
@@ -239,6 +286,292 @@ public function get_references( $request ) {
...
+		$error = $this->pipeline->guard_slug( $slug )
+			?? $this->pipeline->guard_version( $slug, $version );
+
+		if ( $error instanceof WP_Error ) {
+			return $error;
+		}
+
+		$slug_input = Cast::to_string( $request->get_param( 'id' ) );
+
+		if ( ! preg_match( '/^[a-z0-9]+(?:-[a-z0-9]+)*$/', $slug_input ) ) {
```
d4mation: We need a centralized way to validate this. Having this RegEx in multiple places smells.

includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
```diff
@@ -239,6 +286,292 @@ public function get_references( $request ) {
...
+
+		$error = $this->pipeline->guard_slug( $slug )
+			?? $this->pipeline->guard_version( $slug, $version )
+			?? $this->validate_canonical_id( $id );
+
+		if ( $error instanceof WP_Error ) {
+			return $error;
+		}
+
+		if ( $this->baseline->has( $id ) ) {
```
d4mation: Maybe it should instead check if it is a user-created primitive? Checking the baseline directly may be less reliable overall.

includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
```diff
@@ -293,10 +626,88 @@ public function get_references_schema(): array {
...
+	 * @return array<string, mixed>
+	 */
+	private function rewrite_aliases( array $document, string $old_alias, string $new_alias, array &$rewritten ): array {
+		$semantic = $document['semantic'] ?? null;
+
+		if ( ! is_array( $semantic ) ) {
+			return $document;
+		}
+
+		$document['semantic'] = $this->rewrite_node( $semantic, 'semantic', $old_alias, $new_alias, $rewritten );
```
d4mation: What if a System Primitive was set to point to the Custom Primitive? Then it wouldn't be updated.

includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
pauloiankoski: Added PUT routes: c2bd1aeb7468086d5cf2989841d37f01ce40e426.

includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
pauloiankoski: Added a centralized validation helper: 7fca3a990ca25840d850a6d2ec8c54763e409327.

includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
pauloiankoski: Added registry check for system primitives: e1872d0b43354ab76742f691266c9c348cf737e1.

includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
pauloiankoski: `rename_item` now rejects the rename (422) when a reference exists outside the semantic layer (e.g. a primitive composite field or extension token pointing at the old id) instead of leaving it stale. Fixed in 461cfa2ddc879ac878b096ce1c36a1155209935b.

includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
```diff
@@ -182,6 +190,59 @@ public function register_routes(): void {
...
+					'callback'            => [ $this, 'rename_item' ],
+					'permission_callback' => [ $this, 'update_item_permissions_check' ],
+					'args'                => $this->get_rename_params(),
+				],
+				[
+					'methods'             => 'PUT',
+					'callback'            => [ $this, 'rename_item' ],
+					'permission_callback' => [ $this, 'update_item_permissions_check' ],
+					'args'                => $this->get_rename_params(),
+				],
```
d4mation, line 232: This could be a single definition with `'methods' => [ WP_REST_Server::CREATABLE, 'PUT' ]`, right?

d4mation: Or similar. Cannot remember the exact syntax 😅

includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
d4mation: This feels like the wrong solution. There could be a reason why a site owner would create a Custom Color Primitive and then later decide to point a System Color Primitive to it. Completely blocking that flow to avoid changing this rewrite logic to support scanning Primitives feels weird. As long as this scanned both Primitives and Semantics the same way that `Token_Reference_Policy` does, then we could remove those guards.

pauloiankoski: Agreed, and done in 18bdb639a. `Token_Reference_Policy` now classifies a direct `$value` alias
match in the `primitive` layer as a supported `primitive_override` reference (previously it was
lumped in with `composite_field` and marked unsupported), matching how it already treats
semantic-layer aliases. `User_Primitives_Controller::rewrite_aliases()` now walks
`Layers::token_layers()` (both `primitive` and `semantic`) instead of hardcoding the semantic
tree, so a system primitive pointing at the custom primitive being renamed gets its alias
rewritten to the new id instead of being blocked. The 422 guard in `rename_item` is unchanged in
code, but with the reclassification it no longer fires for this case — it still fires for the
references it should (composite fields and extension-preset aliases), which the phase-1 cascade
still can't rewrite. Added `testRenameRewritesPrimitiveLayerDirectAliasReference` to cover the
scenario you described, and updated the `Token_Reference_PolicyTest` case for the primitive-layer
direct alias to assert `supported` instead of blocking.

includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
pauloiankoski: Done in 73b30f6e7. Collapsed the two-entry `create_item` and `rename_item` route registrations
into a single handler each using `'methods' => [ WP_REST_Server::CREATABLE, 'PUT' ]`, since both
handlers shared an identical callback, permission callback, and args in both cases. Verified the
existing route-registration tests and the full `User_Primitives_ControllerTest` suite still pass
with the array form.

includes/resources/Design_Tokens/Rest/V1/User_Primitives_Controller.php
d4mation: Nice!

This unfortunately adds a small gap 😅 

- The `delete_item()` docblock is now stale because it says that it "reverts any supported semantic references" when now it also covers primitive ones.
- We should add a test for the same scenario you have for rename but for delete to ensure that it is covered properly.

pauloiankoski: Reworded the `delete_item()` docblock to state that it reverts supported references regardless of layer (semantic or primitive-layer direct `$value` aliases), since it's driven by `all_supported()` rather than a semantic-only check. Also added `testDeleteRevertsPrimitiveLayerDirectAliasReference`, mirroring `testRenameRewritesPrimitiveLayerDirectAliasReference`: a primitive-layer token aliasing the custom primitive being deleted now has that reference reverted (path in `revertedPaths`, node removed from the document) instead of being blocked. Committed in 556de3d6c.

d4mation (approved):

pauloiankoski:
Superseded by #1117 — this PR's base branch (`SOFT-3574/phase-8-write-guards`) was deleted when #1087 merged, which auto-closed this PR. Recreated with the same head against `SOFT-3574/user-primitive-tokens`.

PR #1090, opened 2026-07-01
[DTM] SOFT-3574 [11/11] Add Style Book UI for user-created color primitives

pauloiankoski:
## Summary
- Extends the Style Book (from the `SOFT-3407` token-catalog stack) with add/delete/rename controls for user-created color primitives: `AddPrimitiveDialog`, `DeletePrimitiveDialog`, `RenamePrimitiveDialog`, and the REST client/path helpers and `useUserPrimitiveEditor` hook that back them.
- `DeletePrimitiveDialog` fetches the references preview on open and disables confirmation while any unsupported reference exists, mirroring the backend's block-on-unsupported-reference behavior. `RenamePrimitiveDialog` intentionally makes no equivalent safety claim, since the backend's rename only rewrites direct semantic-layer aliases and does not scan for unsupported references (a limitation called out in the phase-10 PR).
- Fixes a reachability bug found during review: `FoundationPage` always rendered the token list in its flat (non-grouped) mode, which never mounted the `TokenGroup` component holding the add/rename/delete controls — so the mutation UI shipped in this phase's original commit was unreachable through the app regardless of custom-color count. The ungrouped path now routes through `TokenGroup` (with its heading suppressed, since the page already shows one), scoped to the Custom Colors section only. The Custom Colors section is also now always present in the sidebar, even before any custom color exists, so there's a way to create the first one.
- Fixes a version-conflict (409) UX gap: `AddPrimitiveDialog` previously had no conflict handling (a stale version would silently retry-and-fail forever); `RenamePrimitiveDialog` mislabeled a conflict as an ID collision. Both now prompt a reload. `DeletePrimitiveDialog`'s existing silent preview-refresh-on-conflict now surfaces an interim notice.
- Also adds a PHP `Localizer` test proving a user primitive's `userCreated`/`label` metadata survives `User_Primitive_Registrar::sync()` into the localized feed the UI consumes, and a `test` script in `package.json`.

## Test plan
- [x] Added `navigation.test.js` covering that the Custom Colors section is always present (including with zero user-created tokens) and correctly flagged `isUserCreated`.
- [x] `use-user-primitive-editor.test.js`, `tokens.test.js`, `paths.test.js` (existing) plus the new test all pass via `bun run test`.
- [x] `LocalizerTest` (PHP) passes.
- [x] ESLint clean on all touched JS.
- [x] cspell clean.
- [x] Full wpunit/acceptance/snapshot suites pass (pre-push CI-equivalent hook).

### DIFF
--- a/src/style-book/components/pages/TokensPage.js
+++ b/src/style-book/components/pages/TokensPage.js
@@ -6,7 +6,6 @@ import { useCallback, useState } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import { DEFAULT_TOKEN_SET_SLUG } from '../../constants';
 import { CUSTOM_COLORS_GROUP_LABEL, SECTION_OVERVIEW } from '../../constants/navigation';
 import { findSection } from '../../helpers/navigation';
 import { useDesignTokensFeed } from '../../hooks/use-design-tokens-feed';
@@ -55,6 +54,7 @@ export function TokensPage() {
 		values: feedValues,
 		rest,
 		version: initialVersion,
+		slug,
 	} = useDesignTokensFeed();
 
 	// Shared across every write surface on this page: the document version changes on any
@@ -63,7 +63,7 @@ export function TokensPage() {
 	// tripping a false-positive conflict when one surface's write is followed by another's.
 	const [version, setVersion] = useState(initialVersion);
 
-	const { values, saveToken, getFieldState, refreshValues } = useTokenEditor(rest, feedValues, setVersion);
+	const { values, saveToken, getFieldState, refreshValues } = useTokenEditor(rest, feedValues, setVersion, slug);
 
 	const [localTokens, setLocalTokens] = useState(null);
 	const tokens = localTokens ?? feedTokens;
@@ -72,7 +72,7 @@ export function TokensPage() {
 
 	const { createPrimitive, deletePrimitive, renamePrimitive, fetchPreview } = useUserPrimitiveEditor(
 		version,
-		DEFAULT_TOKEN_SET_SLUG,
+		slug,
 		setVersion
 	);
 
--- a/src/style-book/components/pages/TokensPage.js
+++ b/src/style-book/components/pages/TokensPage.js
@@ -1,14 +1,41 @@
 /**
+ * WordPress dependencies
+ */
+import { useCallback, useState } from '@wordpress/element';
+
+/**
  * Internal dependencies
  */
-import { SECTION_OVERVIEW } from '../../constants/navigation';
+import { DEFAULT_TOKEN_SET_SLUG } from '../../constants';
+import { CUSTOM_COLORS_GROUP_LABEL, SECTION_OVERVIEW } from '../../constants/navigation';
 import { findSection } from '../../helpers/navigation';
 import { useDesignTokensFeed } from '../../hooks/use-design-tokens-feed';
 import { useStyleBookNavigation } from '../../hooks/use-style-book-navigation';
 import { useTokenEditor } from '../../hooks/use-token-editor';
+import { useUserPrimitiveEditor } from '../../hooks/use-user-primitive-editor';
 import { FoundationPage } from '../pages/FoundationPage';
 import { OverviewPage } from '../pages/OverviewPage';
 import { StyleBookShell } from '../templates/StyleBookShell';
+
+/**
+ * Build an optimistic token definition from a create payload.
+ *
+ * @param {object} payload Create request payload { id, label, $value }.
+ * @return {object} Token definition ready for the flat token list.
+ */
+function tokenFromCreatePayload(payload) {
+	const id = `primitive.color.custom.${payload.id}`;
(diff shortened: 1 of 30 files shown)


src/style-book/hooks/use-user-primitive-editor.js
```diff
@@ -0,0 +1,132 @@
...
+ * Manage user-primitive mutations (create, delete, rename) and version tracking.
+ *
+ * @since TBD
+ *
+ * @param {string} initialVersion Feed version string from the localized payload.
+ * @param {string} slug           Token set slug.
+ * @return {{ version: string, isPending: boolean, fetchPreview: Function, createPrimitive: Function, deletePrimitive: Function, renamePrimitive: Function }}
+ */
+export function useUserPrimitiveEditor(initialVersion, slug) {
+	const [version, setVersion] = useState(initialVersion);
```
d4mation: The `version` only updates here from the Primitive Editor's own CRUD responses, but the version is actually dependent on the _entire Document_ so an edit of a Semantic Token or something similar would update that Version without sending it back here. That could be an issue, as it could start to claim that a page reload is required in cases like this:

1. Edit an existing Semantic Token
2. Create/Rename a custom Primitive Token
3. Click submit

The Add/Rename will fail saying that the token set has changed since the page was loaded. 

`DeletePrimitiveDialog` currently doesn't hit this problem because it doesn't trust the `version` used here and instead grabs the actual version from a separate `GET` request, but other cases will hit things like this.

I think in all Primitive CRUD cases, instead of trying to do a `GET` to retrieve the actual version, we should instead share a version for the entire application. This way the guard that is in-place for cases like another tab/user editing the Document are caught properly still (as the version would be different) and we won't hit this issue. 

But if you think delete should still `GET` the actual version and bypass the rest of the system, I guess that could work. We just need to make it very clear _why_ it does that.

src/style-book/components/pages/TokensPage.js
```diff
@@ -1,25 +1,95 @@
...
+ */
+function tokenFromCreatePayload(payload) {
+	const id = `primitive.color.custom.${payload.id}`;
+	const label = payload.label?.trim() || payload.id.replace(/-/g, ' ').replace(/\b\w/g, (c) => c.toUpperCase());
+
+	return {
+		id,
+		type: 'color',
+		label,
+		cssVar: `--kb-${id.replace(/\./g, '-')}`,
```
d4mation: Is this the correct placement? This should replace to `--` right? See `Css_Var::from_id()`.

src/style-book/hooks/use-user-primitive-editor.js
d4mation, line 1: I tested this and it looks like created User Primitives are not shown on reload of the page. I guess they're not being read correctly? I can see them in the Document itself.

src/style-book/hooks/use-user-primitive-editor.js
pauloiankoski: Committed as f2ebb30de167b47e29b87aef3bff760e56071606. `TokensPage` now owns a single version state shared by both `useTokenEditor` (semantic edits) and `useUserPrimitiveEditor` (create/rename/delete), syncing whenever either surface's write returns a fresh version — so editing a semantic token no longer leaves the primitive editor holding a stale version that would trip a false-positive 409 on the next create/rename.

src/style-book/components/pages/TokensPage.js
pauloiankoski: Committed as cb5121868c2707c1d846e7d9f48d11f10f13d344. The optimistic `cssVar` built right after creating a primitive was wrong on two counts — it used a single - instead of -- between path segments, and dropped the `token--` segment entirely — so it didn't match `Css_Var::from_id()`'s actual format (`--kb-token--` + dots joined by `--`). Fixed to reconstruct the same rule (e.g. `primitive.color.custom.my-blue` → `--kb-token--primitive--color--custom--my-blue`).

src/style-book/hooks/use-user-primitive-editor.js
pauloiankoski: `Admin\Feed\Localizer` hardcoded `Token_Store::default_slug()` when resolving values/version, while `User_Primitive_Registrar` (which drives the schema) and every projector (CSS vars, theme.json, presets, variants) already resolve against `Active_Set_Store::get()`. A newly created primitive is written wherever the caller's slug points, but its *existence* in the schema is entirely sourced from whichever document the registrar synced from — so once the active set diverges from `default`, the write and the read disagree and the primitive never resurfaces after a reload, even though it's sitting in the document. System tokens don't show this because their structure is declared unconditionally, independent of any document.

e575bd1 makes `Localizer` resolve against `Active_Set_Store::get()` (matching the registrar/projectors), exposes that resolved `slug` on the feed, and threads it through `useDesignTokensFeed` → `TokensPage` → `useTokenEditor`/`useUserPrimitiveEditor` instead of the hardcoded `DEFAULT_TOKEN_SET_SLUG`, so every write from the Style Book targets the same document the page is displaying.

d4mation (approved):
This should work for now :) 

I think we'll need to support multi-token set editing here, but that should probably wait until we have Figmas for it.

PR #1093, opened 2026-07-01
feat(design-tokens): namespaced css-var naming and resolution (foundation)

d4mation:
🎫 [SOFT-3587](https://linear.app/nexcess/issue/SOFT-3587/simultaneous-multi-set-emission-active-alias-layer)

📹 https://www.loom.com/share/b367269167694d2fb17386b695609880

> Depends on #1077 (SOFT-3586); merge that first.

Foundation for simultaneous multi-set emission: a set-namespaced mode for css-var naming and token resolution, with no projection wiring yet.

- `Css_Var::from_id()` gains an optional `$namespace` that inserts the set slug as a leading segment after the `--kb-token--` prefix, so each set's tokens occupy their own namespace while sharing the one derivation rule. `from_id('semantic.color.text')` still yields `--kb-token--semantic--color--text`; `from_id('semantic.color.text', 'dark')` yields `--kb-token--dark--semantic--color--text`.
- `Token_Resolver::resolve_namespaced($slug)` mirrors `resolve()` but returns the set's resolved maps with every css-var name namespaced to `$slug`. A semantic that aliases a primitive points at that primitive's namespaced var in the same set (`--kb-token--dark--semantic--color--button-bg: var(--kb-token--dark--primitive--color--brand)`), so the alias indirection survives into CSS and each set's chain stays self-contained -- no cross-set references. A shared `load()` path memoizes the canonical and namespaced forms separately under version-qualified cache keys, so a store write self-invalidates both.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Registry/Css_Var.php
+++ b/includes/resources/Design_Tokens/Registry/Css_Var.php
@@ -2,6 +2,8 @@
 
 namespace KadenceWP\KadenceBlocks\Design_Tokens\Registry;
 
+use KadenceWP\KadenceBlocks\Design_Tokens\Projection\Traits\Sanitizes_Css_Identifier;
+
 /**
  * Derives the canonical CSS custom-property name from a token id.
  *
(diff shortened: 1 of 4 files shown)


pauloiankoski (approved):
This looks good to go. Optionally, we could validate the `$namespace` prop in the `from_id` method with `Sanitizes_Css_Identifier` but not a blocker for this PR.

PR #1103, opened 2026-07-02
Custom Variant Creation [7/11]: add the save-as-new-variant editor UI

d4mation:
> [!IMPORTANT]
> Stacked PR — #1102 must be merged first.

🎫  https://linear.app/nexcess/issue/SOFT-3575/user-created-variants-clone-an-existing-variants-surface-no-token

The editor affordance for user-created variants.

A new `SaveVariantModal` clones an existing variant's surface: the user picks a variant to clone from, edits the value of each bound property (rendered by its `kind`), names the variant, and saves through `POST /variants/{block}?set=`. Because it seeds from an existing variant's values it needs no per-block knowledge, so it works for every block that registers a variant set. The server aliases matching literals, enforces the full surface, and rejects dangling aliases; on success the new variant is appended to the in-memory catalog and selected on the block.

A small variants REST client (`src/extension/variants`) wraps the block routes over the editor's `apiFetch`, and `src/extension/variants/slug.js` derives a kebab slug from the label and de-duplicates it against the block's existing slugs (the reserved `default` included).

The variant picker is now set-aware: it reads the per-set catalog and shows the variants for whichever set the block is on (its `kbTokenSet`, or the active set). The "Save as new variant" affordance is wired into the generic Design Tokens inspector panel and the singlebtn inline panel, gated on the REST descriptor being present.

Verified with `wp-scripts lint-js` and a full `wp-scripts build` (the repo has no JS test harness).

Stacks on the per-set catalog work.

### Checklist

### DIFF
--- a/src/blocks/singlebtn/edit.js
+++ b/src/blocks/singlebtn/edit.js
@@ -84,7 +84,9 @@ import {
 } from '@wordpress/components';
 import { addFilter, applyFilters, doAction } from '@wordpress/hooks';
 import BackendStyles from './components/backend-styles';
-import { VariantPicker, blockVariants } from '../../extension/variant-picker';
+import { VariantPicker, blockVariants, activeSet } from '../../extension/variant-picker';
+import { SaveVariantModal } from '../../extension/variant-picker/SaveVariantModal';
+import { hasVariantsRest } from '../../extension/variants/api/client';
 import { TokenSetPicker, selectableSets } from '../../extension/token-set-picker';
 
 export default function KadenceButtonEdit(props) {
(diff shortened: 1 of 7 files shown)


pauloiankoski (approved):
@d4mation I see the variant edit form shows the raw text (like `{semantic.color.button-primary-bg}`) instead of the real color. Is that just because the final UI for this isn't done yet, right? If so, I'm fine approving this as is.

Image

d4mation:
> @d4mation I see the variant edit form shows the raw text (like `{semantic.color.button-primary-bg}`) instead of the real color. Is that just because the final UI for this isn't done yet, right? If so, I'm fine approving this as is.
> 
> Image

Correct :) I flagged this in my demo video. I figure it is something we'll improve later once we have proper designs.

PR #1106, opened 2026-07-04
[DTM] SOFT-3398 [2/5]: Value translator

pauloiankoski:
## Summary

Adds a `Value_Translator` for converting Site Editor (Global Styles) preset values into DTCG token leaves.

- `Value_Translator` interface: translates a literal preset value for a given `wp_preset` category (`color`, `spacing`, `font-family`) into a `['\$type' => ..., '\$value' => ...]` DTCG leaf. Deliberately excludes `shadow` — the Site Editor's shadow preset is a CSS shorthand string, while the DTCG `shadow` `\$type` is a structured, alias-able composite; that conversion is out of scope here and is skipped upstream by the sync listener instead of being guessed at.
- `Default_Value_Translator`: the default implementation.
  - `color` — passed through unchanged (hex, `rgb()`/`rgba()`, `hsl()`/`hsla()`); the write pipeline's own DTCG validator re-validates it before commit, so this only rejects an empty value.
  - `spacing` — passed through unchanged as a `dimension` leaf; theme.json `spacingSizes` values are already valid CSS length strings.
  - `font-family` — wrapped into the array form the DTCG `fontFamily` `\$type` requires; a comma-separated stack is split and trimmed into its individual families.
  - Any other category, or an empty/malformed literal, throws `Untranslatable_Value_Exception`.
- `Untranslatable_Value_Exception`: thrown for an unsupported category or a malformed literal (empty value, or all-comma font-family list).

## Test plan

- [x] `Default_Value_TranslatorTest` covers color (hex, `rgb()`, empty), spacing (`px`, `rem`, `clamp()`, empty), font-family (single name, comma stack, whitespace trimming, empty, all-commas), and the unsupported-category cases (`shadow`, an arbitrary unknown category).

### DIFF
--- a/includes/resources/Design_Tokens/Global_Styles/Default_Value_Translator.php
+++ b/includes/resources/Design_Tokens/Global_Styles/Default_Value_Translator.php
@@ -1,0 +1,120 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Global_Styles;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Token_Type;
+
+/**
+ * Default Value_Translator: color, spacing (dimension), font-family only.
+ *
+ * @since TBD
+ */
+final class Default_Value_Translator implements Value_Translator {
+
+	/**
+	 * @inheritDoc
+	 *
+	 * @since TBD
+	 *
+	 * @param string $category       The wp_preset category ("color", "spacing", "font-family").
+	 * @param string $literal_value  The literal value read from the Global Styles preset entry.
+	 *
+	 * @return array<string, mixed>
+	 *
+	 * @throws Untranslatable_Value_Exception When the category is unsupported or the value is malformed.
+	 */
+	public function translate( string $category, string $literal_value ): array {
+		switch ( $category ) {
+			case 'color':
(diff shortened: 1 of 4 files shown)


includes/resources/Design_Tokens/Global_Styles/Default_Value_Translator.php
```diff
@@ -0,0 +1,120 @@
...
+	 * @param string $value The font family value to translate.
+	 *
+	 * @return array<string, mixed>
+	 *
+	 * @throws Untranslatable_Value_Exception When the value is empty or contains only whitespace.
+	 */
+	private function font_family( string $value ): array {
+		$families = array_values(
+			array_filter(
+				array_map( 'trim', explode( ',', $value ) ),
```
d4mation, line 106: If a font-family were set to something like `-apple-system, "Segoe UI", Roboto`, then it would keep the surrounding quotes around `"Segoe UI"`. Since it then gets re-wrapped in `Css_Renderer::font_family()` due to containing a space, you'd get `""Segoe UI""` and it will break.

pauloiankoski:
Superseded by #1118 — this PR's base branch (`SOFT-3398/phase-1-preset-locator`) was deleted when #1105 merged, which auto-closed this PR. Recreated with the same head against `SOFT-3398/base`.

PR #1107, opened 2026-07-04
[DTM] SOFT-3398 [3/5]: Global styles sync listener

pauloiankoski:
## Summary

Adds `Global_Styles_Sync_Listener`, which hooks `wp_after_insert_post` to detect when the Site Editor's `global-styles` custom post is saved and fires a dedicated synced action so downstream consumers can react to global-styles updates without re-implementing the post-type/status checks.

- `Global_Styles_Sync_Listener` filters for the `wp_global_styles` post type and a published/non-revision save, then fires the synced action via `do_action_ref_array()` (rather than `do_action()`) so WordPress does not unwrap the single-object payload array before it reaches listeners.
- `Global_Styles\Provider` wires the listener into the container and registers it on `wp_after_insert_post`.
- `Design_Tokens_Provider` boots the `Global_Styles\Provider`.
- `Global_Styles_Sync_ListenerTest` covers the listener's filtering and action-firing behavior.

### DIFF
--- a/includes/resources/Design_Tokens/Global_Styles/Global_Styles_Sync_Listener.php
+++ b/includes/resources/Design_Tokens/Global_Styles/Global_Styles_Sync_Listener.php
@@ -209,7 +209,11 @@ final class Global_Styles_Sync_Listener {
 			 * @param array<int, Preset_Target> $synced The presets that were synced.
 			 * @param WP_Post                   $post   The wp_global_styles post.
 			 */
-			do_action( self::SYNCED_ACTION, $synced, $post );
+			// do_action_ref_array(), not do_action(): do_action()'s single-object-array backward-
+			// compatibility unwrap (`array( &$this )`, wp-includes/plugin.php) silently collapses
+			// $synced to its bare element whenever exactly one preset is synced, handing subscribers
+			// a lone Preset_Target instead of a one-item list.
+			do_action_ref_array( self::SYNCED_ACTION, [ $synced, $post ] );
 		}
 	}
 
--- a/includes/resources/Design_Tokens/Global_Styles/Global_Styles_Sync_Listener.php
+++ b/includes/resources/Design_Tokens/Global_Styles/Global_Styles_Sync_Listener.php
@@ -1,0 +1,341 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Global_Styles;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Database\Active_Set_Store;
+use KadenceWP\KadenceBlocks\Design_Tokens\Database\Token_Store;
+use KadenceWP\KadenceBlocks\Design_Tokens\Document\Mutator;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Exception\Alias_Cycle_Exception;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Exception\Dangling_Alias_Exception;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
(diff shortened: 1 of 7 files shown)


includes/resources/Design_Tokens/Global_Styles/Global_Styles_Sync_Listener.php
```diff
@@ -0,0 +1,346 @@
...
+ * Token_Store::save_document() commits it.
+ *
+ * Only tokens Site_Editor_Preset_Locator returns (wp_preset + site_editor => true) are ever
+ * touched; every other preset in the payload — ad-hoc or token-backed-but-not-opted-in — is left
+ * exactly as WordPress wrote it. A changed preset whose category has no Value_Translator support
+ * (currently "shadow") is skipped and logged, not guessed at.
+ *
+ * @since TBD
+ */
+final class Global_Styles_Sync_Listener {
```
d4mation:
```suggestion
final class Sync_Listener {
```

includes/resources/Design_Tokens/Global_Styles/Global_Styles_Sync_Listener.php
```diff
@@ -0,0 +1,346 @@
...
+						'Global Styles token sync failed for "%s" (%s): %s',
+						$target->token->id,
+						$target->category,
+						$e->getMessage()
+					)
+				);
+			}
+		}
+
+		if ( $synced !== [] ) {
```
d4mation: Why not return early on `empty( $synced )`? Then we aren't nesting the `do_action()` call.

includes/resources/Design_Tokens/Global_Styles/Provider.php
```diff
@@ -0,0 +1,34 @@
...
+
+	/**
+	 * @inheritDoc
+	 *
+	 * @since TBD
+	 */
+	public function register(): void {
+		$this->container->singleton( Site_Editor_Preset_Locator::class, Site_Editor_Preset_Locator::class );
+		$this->container->singleton( Value_Translator::class, Default_Value_Translator::class );
+		$this->container->singleton( Global_Styles_Sync_Listener::class, Global_Styles_Sync_Listener::class );
```
d4mation: If we're not binding a specific implementation for something like an Interface, then both params are not needed.

includes/resources/Design_Tokens/Global_Styles/Global_Styles_Sync_Listener.php
pauloiankoski: Applied — renamed the class, its file, and its test class (`Global_Styles_Sync_ListenerTest` → `Sync_ListenerTest`), and updated every reference (`Provider`, `Value_Translator`'s docblock). Fixed in commit 1104ccff4, pushed to `SOFT-3398/phase-3-sync-listener`.

includes/resources/Design_Tokens/Global_Styles/Global_Styles_Sync_Listener.php
pauloiankoski: Applied — inverted the `if ( $synced !== [] ) { ... }` block at the end of `on_after_insert_post()` into an early `if ( $synced === [] ) { return; }`, dropping a level of nesting around the `do_action_ref_array()` call. Fixed in commit 1104ccff4.

includes/resources/Design_Tokens/Global_Styles/Provider.php
pauloiankoski: Applied — `Site_Editor_Preset_Locator::class` and `Sync_Listener::class` were both self-bound (`singleton( X::class, X::class )`); dropped the redundant second argument on both. Left `Value_Translator::class, Default_Value_Translator::class` as-is since that one *is* a real interface-to-implementation binding. Fixed in commit 1104ccff4.

d4mation (approved):

PR #1108, opened 2026-07-04
[DTM] SOFT-3398 [4/5]: Detachment detector

pauloiankoski:
## Summary

Adds `Detached_Token_Detector`, which derives whether a design token is "detached from brand": its baseline `\$value` is an alias (the normal shape for a semantic token pointing at a primitive/brand value), but its effective value — baseline merged with the currently stored overrides — is a literal instead.

Detachment is computed as a pure function of `(baseline, overrides)` on every call, with no state stored, following the same approach as other projectors in the design tokens module (e.g. `Css_Var_Projector`, `Palette_Builder`).

Key behaviors:
- A token whose baseline alias has been overridden with a literal is detached.
- An aliased token with no override, or re-pointed to a different alias, is not detached.
- A RESET sentinel override (`\$value: null`) falls back to the baseline alias and is not detached.
- A token whose baseline is already a literal has no alias relationship to lose, so overriding it with another literal is an ordinary edit, not a detachment.
- An unknown token id is never detached.

## Test plan

- [x] `Detached_Token_DetectorTest` covers each transition against a synthetic baseline, plus a check against the real shipped baseline for a real semantic token.

## Note

Renamed `Detachment_Detector` → `Detached_Token_Detector` per review feedback. This branch (and phase 5) were forked from phase 3 before phase 3's latest commit (`f3480b987`, "skip re-sync when a preset's literal is unchanged since the last save") landed — needs a rebase onto phase 3's tip before final merge so that fix isn't lost.

### DIFF
--- a/includes/resources/Design_Tokens/Global_Styles/Detachment_Detector.php
+++ b/includes/resources/Design_Tokens/Global_Styles/Detachment_Detector.php
@@ -5,6 +5,7 @@ namespace KadenceWP\KadenceBlocks\Design_Tokens\Global_Styles;
 use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Contracts\Baseline_Document;
 use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Effective_Document;
 use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Alias;
+use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Sentinels;
 
 /**
  * Derives "detached from brand": a token whose baseline $value is an alias (the normal shape for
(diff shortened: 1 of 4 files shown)


includes/resources/Design_Tokens/Global_Styles/Detachment_Detector.php
```diff
@@ -0,0 +1,105 @@
...
+ * every call, so it can never drift from what the token store actually holds — the same ideology
+ * every other projector in this module already follows (see Css_Var_Projector, Palette_Builder).
+ *
+ * A token whose baseline is ALREADY a literal (most primitives) is never "detached" — there was
+ * no alias relationship to lose, so overriding it with a different literal is an ordinary edit,
+ * not a demotion.
+ *
+ * @since TBD
+ */
+final class Detachment_Detector {
```
d4mation: Maybe `Detached_Token_Detector` is better?

pauloiankoski: Applied — renamed the class, its file, and its test class (`Detachment_DetectorTest` → `Detached_Token_DetectorTest`). Not wired into `Provider` yet, so no other references to update. Fixed in commit c4bba6c34.

d4mation (approved):

PR #1109, opened 2026-07-04
[DTM] SOFT-3398 [5/5]: Override stripper

pauloiankoski:
## Summary

Adds `Global_Styles\Reference\Restorer`, the final phase of the design-tokens Global Styles sync/detach round-trip. After `Global_Styles_Sync_Listener` writes a user's literal preset value into the token store, this class rewrites the `wp_global_styles` post's `post_content` to restore the synced entries back to `var(--kb-token--*)`, keeping the CSS variable the single source of truth once the sync has captured the user's value.

- `Restorer::restore_synced()` takes the `Preset_Target` list that was just synced plus the `wp_global_styles` `WP_Post`, decodes `post_content` (as objects, not associative arrays, so an empty JSON object round-trips correctly), restores each target's value back to its `var(--kb-token--*)` reference, and persists via `wp_update_post()` only when something actually changed.
- `Reference\Listener` wires the restorer to `Global_Styles_Sync_Listener::synced_action()`, so the restore runs immediately after a sync with the exact set of targets that were synced — no re-derivation of which presets changed.
- Registered in `Global_Styles` `Provider`.
- The resulting `wp_update_post()` re-fires `wp_after_insert_post`, but that second pass is a guaranteed no-op: the sync listener compares against the canonical `var(--kb-token--*)` form (which the restorer just restored), so nothing "changes" on re-entry.

## Test plan

- [x] `RestorerTest` covers restoring single and multiple synced targets, no-op paths (empty synced list, malformed `post_content`, no matching entries), persistence behavior, and an empty-JSON-object round-trip regression case.
- [x] `Global_Styles_Sync_ListenerTest` extended with a full-chain assertion verifying sync → restore round-trips back to `var(--kb-token--*)`.

## Note

Per review feedback: moved `Override_Stripper` → `Global_Styles\Reference\Restorer` and `Override_Stripper_Hook_Listener` → `Global_Styles\Reference\Listener`; fixed a real bug where `json_decode( ..., true )` could flatten an empty JSON object (`"settings": {}`) to `[]` and corrupt the document shape on write. This branch (and phase 4) were forked from phase 3 before phase 3's latest commit (`f3480b987`, "skip re-sync when a preset's literal is unchanged since the last save") landed — needs a rebase onto phase 3's tip before final merge so that fix isn't lost.


includes/resources/Design_Tokens/Global_Styles/Provider.php
```diff
@@ -20,6 +20,8 @@ public function register(): void {
 		$this->container->singleton( Site_Editor_Preset_Locator::class, Site_Editor_Preset_Locator::class );
 		$this->container->singleton( Value_Translator::class, Default_Value_Translator::class );
 		$this->container->singleton( Global_Styles_Sync_Listener::class, Global_Styles_Sync_Listener::class );
+		$this->container->singleton( Override_Stripper::class, Override_Stripper::class );
+		$this->container->singleton( Override_Stripper_Hook_Listener::class, Override_Stripper_Hook_Listener::class );
```
d4mation: If we're not binding a specific implementation for something like an Interface, then both params are not needed.

includes/resources/Design_Tokens/Global_Styles/Reference/Restorer.php
```diff
@@ -0,0 +1,140 @@
...
+			if ( $this->restore( $decoded, $target ) ) {
+				$changed = true;
+			}
+		}
+
+		if ( ! $changed ) {
+			return;
+		}
+
+		$encoded = wp_json_encode( $decoded );
```
d4mation, line 79: This could cause issues with `json_decode( $post->content, true )` because an empty object `{}` would get decoded as `[]` and then when re-encoded would remain an empty array which is incorrect. We should avoid decoding to an associative array if possible to get around this.

includes/resources/Design_Tokens/Global_Styles/Override_Stripper_Hook_Listener.php
d4mation, line 1: Based on this class name, it sounds like maybe we should have another namespace here.

Suggestion:

```
Global_Styles\Reference
    - Restorer.php
    - Listener.php
```

Then `$stripper`, `strip()`, etc. would be renamed accordingly.

includes/resources/Design_Tokens/Global_Styles/Override_Stripper.php
```diff
@@ -0,0 +1,140 @@
...
+
+		$result = wp_update_post(
+			[
+				'ID'           => $post->ID,
+				'post_content' => wp_slash( $encoded ),
+			],
+			true
+		);
+
+		if ( is_wp_error( $result ) ) {
```
d4mation: We could return early on `! is_wp_error( $result )`

includes/resources/Design_Tokens/Global_Styles/Provider.php
pauloiankoski: Applied — `Override_Stripper::class` and `Override_Stripper_Hook_Listener::class` (now `Restorer` and `Listener`, see the namespace comment below) were self-bound; dropped the redundant second argument on both, same as the sibling fix on #1107. Fixed across commits 5a5bed509 and bd5b6bf0f.

includes/resources/Design_Tokens/Global_Styles/Reference/Restorer.php
pauloiankoski: Confirmed real bug and fixed. `Restorer::restore_synced()` (formerly `Override_Stripper::strip()`) now decodes `post_content` as objects (`json_decode( $post->post_content )`, no `true`) instead of associative arrays. JSON arrays (the palette's list of preset entries) still decode to PHP arrays either way, so the list-traversal logic is unaffected — only the object-shaped nodes (`settings`, `color`, etc.) change, and objects mutate by handle, so the per-target restore helper no longer needs its by-reference (`&$cursor`) traversal.

Added `RestorerTest::testEmptyJsonObjectRoundTripsAsObjectNotArray` covering the round-trip, and had a second, independent review pass confirm the fix by actually reverting it in an isolated worktree and re-running the suite to see it fail. Fixed across commits 5a5bed509 and bd5b6bf0f, pushed to `SOFT-3398/phase-5-override-stripper`.

includes/resources/Design_Tokens/Global_Styles/Override_Stripper_Hook_Listener.php
pauloiankoski: Applied as suggested, including the method-level rename ("$stripper, strip(), etc. would be renamed accordingly"):
- `Override_Stripper` → `Global_Styles\Reference\Restorer`
- `Override_Stripper_Hook_Listener` → `Global_Styles\Reference\Listener`
- `Override_Stripper::strip()` → `Restorer::restore_synced()` (the public entry point). The pre-existing private per-target helper — previously also named `restore()` — was renamed to `restore_entry()` rather than left colliding with the newly-renamed public method.

`$stripper` properties/variables renamed to `$restorer` throughout, and every reference (`Provider`, docblocks elsewhere) updated. Fixed across commits 5a5bed509 and bd5b6bf0f.

includes/resources/Design_Tokens/Global_Styles/Override_Stripper.php
pauloiankoski: Applied — inverted the trailing `if ( is_wp_error( $result ) ) { ...log... }` into an early `if ( ! is_wp_error( $result ) ) { return; }`, removing the nesting around the log call. Fixed in commit 5a5bed509.

d4mation (approved):

PR #1111, opened 2026-07-07
[DTM] SOFT-3405 [2/4]: Icon size adapter

pauloiankoski:
## Summary

Adds the first concrete `Adapter_Interface`/`Abstract_Adapter` implementation in this module, `Icon_Size_Adapter`, which gap-fills `kadence/single-icon`'s `size` attribute from the resolved `semantic.icon-size.default` token.

- `Design_Tokens\Projection\Adapter\Icon_Size_Adapter` resolves `semantic.icon-size.default`, converts the resulting CSS length (e.g. `"1.5rem"`) to the raw pixel number `size` stores, and applies it only when the stored attributes genuinely omit `size` — an explicit value, including `0`, always wins.
- Registered in `Adapter\Provider::ADAPTERS` alongside the module's other adapters.
- Because `size` has no `source` key and defaults to `50`, WordPress's block serializer omits it from saved content for any block that was never customized away from the default — so this fires for the large majority of currently-published icon blocks, not a narrow legacy-content edge case. This affects the render-path/CSS-build gap-filling case broadly.
- Does not cover a freshly inserted `single-icon` block: that's a static (client-rendered) block, so a fresh insert's `size` comes from `block.json`'s JS default at insert time, before this render-path filter ever runs. A later PR in this stack closes that complementary "freshly inserted block" gap on the editor JS side.

## Test plan

- [x] `Icon_Size_AdapterTest` covers gap-filling when `size` is absent, leaving an explicit stored value (including `0`) untouched, unit conversion (`px`/`rem`/`em`), and the null/unresolvable-token fallback.

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Adapter/Icon_Size_Adapter.php
+++ b/includes/resources/Design_Tokens/Projection/Adapter/Icon_Size_Adapter.php
@@ -1,0 +1,113 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection\Adapter;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Projection\Adapter\Contracts\Abstract_Adapter;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
+
+/**
+ * Gap-fills kadence/single-icon's `size` attribute with the resolved semantic.icon-size.default
+ * token value, converted to the raw pixel number the attribute stores, when the stored attributes
+ * genuinely omit `size`. Because `size` has no `source` key and defaults to `50`, the block
+ * serializer omits it from saved content whenever it was never customized — so this fires for the
+ * large majority of currently-published icon blocks, not a narrow legacy case. Intentional: this
+ * mirrors the retroactive re-skin behavior kadence/image's borderRadius token already has in
+ * production.
+ *
+ * The rem/em-to-px conversion assumes a 16px root font size — the same assumption an unstyled
+ * `rem` makes in a browser. Nothing in this module tracks a site's actual root font size, so this
+ * is a known, accepted simplification rather than a silent guess.
+ *
+ * Does NOT affect a freshly inserted icon block: `single-icon` is a static (client-rendered) block,
+ * so a fresh insert's `size` comes from block.json's JS default at insert time, before this
+ * render-path filter ever runs. See Phase 3 (the editor-default catalog) for that case.
+ *
+ * @since TBD
+ */
+final class Icon_Size_Adapter extends Abstract_Adapter {
+
+	/**
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	protected const BLOCK = 'kadence/single-icon';
+
+	/**
+	 * The resolved-token dot-path this adapter reads.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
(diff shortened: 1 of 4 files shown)


includes/resources/Design_Tokens/Projection/Adapter/Icon_Size_Adapter.php
```diff
@@ -0,0 +1,112 @@
...
+	 */
+	private function to_px( string $length ): ?float {
+		if ( ! preg_match( '/^(-?[0-9.]+)(px|rem|em)$/', trim( $length ), $matches ) ) {
+			return null;
+		}
+
+		$number = (float) $matches[1];
+
+		return $matches[2] === 'px' ? $number : $number * 16.0;
+	}
```
d4mation, line 111: Do we want to have something like this? A site owner/designer could want to use `em` or `rem` intentionally here. Is the Block incompatible with anything other than `px`?

Assuming the root font size is always `16px` could result in some issues.

includes/resources/Design_Tokens/Projection/Adapter/Icon_Size_Adapter.php
```diff
@@ -0,0 +1,112 @@
...
+		$px = $this->to_px( $length );
+
+		if ( $px === null ) {
+			return $attributes;
+		}
+
+		$attributes['size'] = $px;
+
+		return $attributes;
+	}
```
d4mation, line 89: This ends up rendering 

https://github.com/stellarwp/kadence-blocks/blob/2242fe080e254d3a32509feb38bde4129f644e81/includes/blocks/class-kadence-blocks-single-icon-block.php#L64-L67

effectively dead code. Maybe we should remove it to avoid confusion?

pauloiankoski:
Superseded by 1119 — this PR's base branch (`SOFT-3405/phase-1-spike-and-bindings`) was deleted when #1110 merged, which auto-closed this PR. Recreated with the same head against `SOFT-3405/base`.

includes/resources/Design_Tokens/Projection/Adapter/Icon_Size_Adapter.php
pauloiankoski: No code change here — this is already an intentional, documented simplification, not an oversight. `Icon_Size_Adapter`'s class docblock and `to_px()`'s docblock both call out the 16px assumption explicitly and note nothing in this module tracks a site's actual root font size.

More importantly: the block is architecturally incapable of rendering anything but px. `src/blocks/single-icon/block.json`'s `size` attribute is `"type": "number"` with no unit companion attribute, and `render_responsive_size()` always appends `px` when there's no unit attribute — so this isn't a judgment call the adapter is making, it's the only value the attribute can hold. This adapter only ever reads the token's baseline value to seed a numeric default; it never constrains what a site owner can type into the block's own size control afterward.

includes/resources/Design_Tokens/Projection/Adapter/Icon_Size_Adapter.php
pauloiankoski: Agreed and fixed — removed the isset-guarded fallback in `class-kadence-blocks-single-icon-block.php::build_css()`. `Icon_Size_Adapter::apply()` runs on `kadence_blocks_block_default_attributes`, so `merge_attributes_with_defaults()` guarantees `size` is always present on the attributes `build_css()` receives — the fallback could never execute.

Fixed in commit 3677663, pushed to `SOFT-3405/phase-2-icon-size-adapter` (same branch content carried into #1119).

PR #1112, opened 2026-07-07
[DTM] SOFT-3405 [3/4]: Editor default catalog

pauloiankoski:
## Summary

Adds the editor-side attribute-default catalog for design-token-backed blocks, and the `blocks.registerBlockType` filter that applies it.

- `Design_Tokens\Editor\Block_Preset_Catalog` resolves a small per-block, per-attribute map of token dot-paths (today just `kadence/single-icon`'s `size` → `semantic.icon-size.default`) to numeric pixel values via the `Token_Resolver`, converting `rem`/`em` lengths against the CSS default root size. A block/attribute whose token doesn't resolve, or whose value isn't a convertible length, is omitted so the caller falls back to `block.json`'s own default.
- `Design_Tokens\Editor\Localizer` now attaches a third window global, `window.kadenceDesignTokensPresetDefaults`, alongside the existing `kadenceDesignTokensVariants` and `kadenceDesignTokensSets` catalogs — same fail-closed gate and inline-script attachment as the other two.
- `blockPresetAttributeDefault` in `early-filters.js` hooks `blocks.registerBlockType` and, for any block/attribute the catalog covers, overrides the registered attribute's `default` with the resolved value. This means a freshly inserted `kadence/single-icon` starts at the brand's token-resolved size instead of the hardcoded `block.json` default, and — because block registration runs once per editor load — any previously saved post that re-opens (and re-saves) in the editor picks up the same resolved default for that attribute, the same broad, intentional scope as the render-path fix in the prior PR.

There's no JS unit test for `blockPresetAttributeDefault` — this repo has no existing test infrastructure for `early-filters.js`. Coverage is via `Block_Preset_CatalogTest` on the PHP side; manual Lando verification of the editor behavior is still pending.

## Test plan

- [x] `Block_Preset_CatalogTest` covers resolution, unit conversion (px/rem/em), and omission when a token doesn't resolve or convert.
- [ ] Manual Lando verification that a freshly inserted `kadence/single-icon` starts at the brand-resolved size in the editor.

### DIFF
--- a/includes/resources/Design_Tokens/Editor/Block_Preset_Catalog.php
+++ b/includes/resources/Design_Tokens/Editor/Block_Preset_Catalog.php
@@ -2,7 +2,7 @@
 
 namespace KadenceWP\KadenceBlocks\Design_Tokens\Editor;
 
-use KadenceWP\KadenceBlocks\Design_Tokens\Projection\Traits\Converts_Length_To_Px;
+use KadenceWP\KadenceBlocks\Design_Tokens\Projection\Traits\Converts_Number_To_Px;
 use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
 
 /**
@@ -20,7 +20,7 @@ use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
  */
 final class Block_Preset_Catalog {
 
-	use Converts_Length_To_Px;
+	use Converts_Number_To_Px;
 
 	/**
 	 * Block => attribute => resolved-token dot-path. Each entry's value is looked up via the
--- a/includes/resources/Design_Tokens/Editor/Block_Preset_Catalog.php
+++ b/includes/resources/Design_Tokens/Editor/Block_Preset_Catalog.php
@@ -1,0 +1,113 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Editor;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
+
+/**
+ * Builds the compact per-block attribute-default catalog the `blocks.registerBlockType` filter in
+ * early-filters.js reads to seed a freshly inserted block's attribute default from a resolved
+ * token, instead of the hardcoded static default in block.json.
+ *
+ * Scoped to kadence/single-icon's `size` today — the only block/attribute pair this ticket needs.
+ * Not a general "any block, any attribute" registry: extend the ENTRIES map (or promote to a
+ * declarations-driven shape) only when a second real consumer needs it, matching this module's
+ * existing preference for composable-but-not-speculative catalogs (see Variant_Catalog, which
+ * itself started scoped to what the variant picker needed).
+ *
+ * @since TBD
+ */
+final class Block_Preset_Catalog {
+
+	/**
+	 * Block => attribute => resolved-token dot-path. Each entry's value is looked up via the
+	 * resolver and, when present, exposed to JS as a raw number (this catalog only supports
+	 * numeric attribute defaults today — the one case this ticket has).
+	 *
+	 * @since TBD
+	 *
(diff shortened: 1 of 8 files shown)


includes/resources/Design_Tokens/Editor/Block_Preset_Catalog.php
```diff
@@ -0,0 +1,113 @@
...
+	 */
+	private function to_px( string $length ): ?float {
+		if ( ! preg_match( '/^(-?[0-9.]+)(px|rem|em)$/', trim( $length ), $matches ) ) {
+			return null;
+		}
+
+		$number = (float) $matches[1];
+
+		return $matches[2] === 'px' ? $number : $number * 16.0;
+	}
```
d4mation: Similar to my comment on #1111

If we do need this, then I think we should instead move it into a Trait instead of defining it in multiple places.

src/early-filters.js
```diff
@@ -130,6 +130,40 @@ export function blockVariantAttribute(settings) {
...
+ * output for existing content (which already has an explicit stored value).
+ *
+ * @param {Object} settings The block settings.
+ * @param {string} name     The block name.
+ *
+ * @since TBD
+ *
+ * @return {Object} The block settings, with any catalog-covered attribute defaults overridden.
+ */
+export function blockPresetAttributeDefault(settings, name) {
```
d4mation, line 147: This is nice! We should double-check Button Border Radius to see if it should/could use this system too. I handled it a bit differently there.

includes/resources/Design_Tokens/Editor/Block_Preset_Catalog.php
d4mation: I approved this on the last PR, but I still think we should move this to a Trait. It will be needed any case the type is a `number` and the block expects it to be a `px` value.

pauloiankoski: Extracted to a Trait and implemented it where we had that method being repeated. Refactored in 5adbb2c375022ed44907e04e96a165915232e3d8.

src/early-filters.js
pauloiankoski: Yes, Button Border Radius uses the same `Block_Preset_Catalog` mechanism, so I went ahead and prototyped it: [#1125](https://github.com/stellarwp/kadence-blocks/pull/1125).

The main difference from icon size is that `borderRadius` is a 4-corner array instead of a single value. So this PR updates the catalog to support both single values and arrays. The button maps `borderRadius` to `semantic.radius.control`, which resolves to `radius.md` (8px) by default, so a new button gets [8, 8, 8, 8].

`early-filters.js` didn’t need any changes since it already assigns the catalog value directly.

I only included the base Border Radius. I left the tablet/mobile values and the hover, transparent, and sticky radius settings unchanged since your question was only about the base control.

Take a look when you have a chance.

includes/resources/Design_Tokens/Projection/Traits/Converts_Length_To_Px.php
```diff
@@ -0,0 +1,35 @@
...
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection\Traits;
+
+/**
+ * Shared converter from a resolved CSS length to a raw pixel number, used by every adapter/catalog
+ * that overlays a token-resolved default onto a block attribute typed `number` where the block
+ * expects a `px` value.
+ *
+ * @since TBD
+ */
+trait Converts_Length_To_Px {
```
d4mation: Minor thing, but maybe `Converts_Number_To_Px`? Because this could be needed for font size, etc. in the future and that's not a `length`.

d4mation (approved):

includes/resources/Design_Tokens/Projection/Traits/Converts_Length_To_Px.php
pauloiankoski: Implemented a simple and direct refactor in 79ec286a14baf2ac53a20c6c1f0a2ce4c88fae71.

PR #1116, opened 2026-07-07
Custom Variant Creation [11/11]: let a button variant set its own border-radius

d4mation:
> [!IMPORTANT]
> Stacked PR — #1115 must be merged first.

🎫  https://linear.app/nexcess/issue/SOFT-3575/user-created-variants-clone-an-existing-variants-surface-no-token

🎥  https://www.loom.com/share/ad75b63e1b624659983ffbcb2d753a3b

Makes a button variant able to set its own border-radius, and fixes an editor-only leak that was masking it. Two commits.

**1. `fix(advanced-form)`: scope the submit field's editor button styles.** The submit field's editor stylesheet styled bare, unscoped `.kt-button`, so its hardcoded `border-radius: 3px` leaked onto every Kadence button in the editor iframe and overrode the design-token radius at equal specificity + later source order (colors survived only because those rules token-reference the palette vars). Scoped the bare `.kt-button` rules to `.wp-block-kadence-advanced-form-submit`, matching the already-scoped rules in the same file and the front-end's dedicated class.

**2. `feat(design-tokens)`: per-variant button border-radius.** Border-radius was a single shared token across all variants — the variant CSS builder only projected `kadence_slot` (color) bindings, so the `button-radius` `css_var` binding was dropped and a per-variant radius never rendered. Now:
- the `css_var` binding carries the target variable name (`button-radius` drives `--kb-btn-radius`), with a `Binding::css_var()` accessor;
- the variant builder projects `css_var` bindings: a selected variant emits `.kb-variant--<slug>{ --kb-btn-radius: var(<per-variant var>) }` plus the per-variant value in the namespaced block;
- the button reads `border-radius: var(--kb-btn-radius, var(--kb-token--semantic--radius--control, 3px))`, so with no variant it falls through to the shared token exactly as before (no circularity, no regression), and a variant overrides only its own buttons.

Verified on a live site: a variant's radius now emits `--kb-btn-radius:var(--kb-token--variant--kadence-singlebtn--<variant>--button-radius)` and applies live in the editor via the projected-CSS re-injection. wpunit updated (Css_Builder + Binding) and green; PHPStan clean.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Registry/Binding.php
+++ b/includes/resources/Design_Tokens/Registry/Binding.php
@@ -66,8 +66,10 @@ final class Binding {
 	private const BLOCK_ATTR = 'block_attr';
 
 	/**
-	 * Inline target (boolean flag): the property surfaces only as its `--kb-token--*` var, with no
-	 * preset/slot bucket (e.g. border-radius, which WordPress has no preset category for).
+	 * Inline target: a KB-owned CSS custom property this property drives, named without the leading `--`
+	 * (e.g. "kb-btn-radius" → `--kb-btn-radius`). For a property with no Kadence palette slot or WordPress
+	 * preset bucket (e.g. border-radius): the block reads `var(--<css_var>, <token fallback>)`, and a selected
+	 * variant sets `--<css_var>` on its scope so the value can vary per variant.
 	 *
 	 * @since TBD
 	 *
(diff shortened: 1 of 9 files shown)


includes/resources/Design_Tokens/Registry/Binding.php
```diff
@@ -258,6 +260,21 @@ public function css_selector(): ?string {
...
+	 *
+	 * @since TBD
+	 *
+	 * @return string|null
+	 */
+	public function css_var(): ?string {
+		$name = $this->projections[ self::CSS_VAR ] ?? null;
+
+		return is_string( $name ) ? $name : null;
+	}
```
pauloiankoski, line 276: The test for this in [tests/wpunit/Resources/Design_Tokens/Projection/Block_Preset/ProjectorTest.php:128](https://github.com/stellarwp/kadence-blocks/blob/33d781b2e3df4179977624ebc327db864a2292a2/tests/wpunit/Resources/Design_Tokens/Projection/Block_Preset/ProjectorTest.php#L128) still uses the old form `'css_var' => true`. This class now needs
`css_var` to be a real string, like `'kb-btn-radius'`, not `true`. That line will now throw an error and break that test file. Every other place in the code that used true got updated to a string. Can you change it to a string too?

d4mation: Fixed! [`54f106f` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1116/commits/54f106fe36c33c487938ad5252db6d03551afba6)

pauloiankoski (approved):

PR #1118, opened 2026-07-08
[DTM] SOFT-3398 [2/5]: Value translator

pauloiankoski:
## Summary

Adds a `Value_Translator` for converting Site Editor (Global Styles) preset values into DTCG token leaves.

- `Value_Translator` interface: translates a literal preset value for a given `wp_preset` category (`color`, `spacing`, `font-family`) into a `['\$type' => ..., '\$value' => ...]` DTCG leaf. Deliberately excludes `shadow` — the Site Editor's shadow preset is a CSS shorthand string, while the DTCG `shadow` `\$type` is a structured, alias-able composite; that conversion is out of scope here and is skipped upstream by the sync listener instead of being guessed at.
- `Default_Value_Translator`: the default implementation.
  - `color` — passed through unchanged (hex, `rgb()`/`rgba()`, `hsl()`/`hsla()`); the write pipeline's own DTCG validator re-validates it before commit, so this only rejects an empty value.
  - `spacing` — passed through unchanged as a `dimension` leaf; theme.json `spacingSizes` values are already valid CSS length strings.
  - `font-family` — wrapped into the array form the DTCG `fontFamily` `\$type` requires; a comma-separated stack is split and trimmed into its individual families.
  - Any other category, or an empty/malformed literal, throws `Untranslatable_Value_Exception`.
- `Untranslatable_Value_Exception`: thrown for an unsupported category or a malformed literal (empty value, or all-comma font-family list).

## Test plan

- [x] `Default_Value_TranslatorTest` covers color (hex, `rgb()`, empty), spacing (`px`, `rem`, `clamp()`, empty), font-family (single name, comma stack, whitespace trimming, empty, all-commas), and the unsupported-category cases (`shadow`, an arbitrary unknown category).

### DIFF
--- a/includes/resources/Design_Tokens/Global_Styles/Default_Value_Translator.php
+++ b/includes/resources/Design_Tokens/Global_Styles/Default_Value_Translator.php
@@ -92,6 +92,11 @@ final class Default_Value_Translator implements Value_Translator {
 	 * (Schema\Validation\Values\Font_Family_Value), so a plain family name is wrapped, and a
 	 * comma-separated stack is split into its elements.
 	 *
+	 * Each element also has its surrounding quotes stripped (e.g. `"Segoe UI"` -> `Segoe UI`):
+	 * Css_Renderer::font_family() re-wraps any element containing a space in quotes when
+	 * rendering back to CSS, so a literal that already arrived quoted would otherwise come out
+	 * double-quoted (`""Segoe UI""`).
+	 *
 	 * @since TBD
 	 *
 	 * @param string $value The font family value to translate.
@@ -103,7 +108,10 @@ final class Default_Value_Translator implements Value_Translator {
 	private function font_family( string $value ): array {
 		$families = array_values(
 			array_filter(
-				array_map( 'trim', explode( ',', $value ) ),
+				array_map(
+					static fn( string $family ): string => trim( $family, " \t\n\r\0\x0B\"'" ),
+					explode( ',', $value )
+				),
 				static fn( string $family ): bool => $family !== ''
 			)
 		);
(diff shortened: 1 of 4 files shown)


includes/resources/Design_Tokens/Global_Styles/Default_Value_Translator.php
```diff
@@ -0,0 +1,120 @@
...
+	 * @param string $value The font family value to translate.
+	 *
+	 * @return array<string, mixed>
+	 *
+	 * @throws Untranslatable_Value_Exception When the value is empty or contains only whitespace.
+	 */
+	private function font_family( string $value ): array {
+		$families = array_values(
+			array_filter(
+				array_map( 'trim', explode( ',', $value ) ),
```
pauloiankoski: **Originally posted by @d4mation on #1106:**

If a font-family were set to something like `-apple-system, "Segoe UI", Roboto`, then it would keep the surrounding quotes around `"Segoe UI"`. Since it then gets re-wrapped in `Css_Renderer::font_family()` due to containing a space, you'd get `""Segoe UI""` and it will break.

pauloiankoski:
Recreated from #1106 (auto-closed when #1105 merged and deleted its base branch). Review discussion above is copied over, attributed to the original author.

includes/resources/Design_Tokens/Global_Styles/Default_Value_Translator.php
pauloiankoski: Confirmed real bug and fixed. `Default_Value_Translator::font_family()` split/trimmed each comma-separated element but didn't strip surrounding quotes, so a literal like `-apple-system, "Segoe UI", Roboto` stored `"Segoe UI"` (quotes intact) in the DTCG `$value` array. `Css_Renderer::font_family()` then re-wraps any element containing a space in quotes on render, producing `""Segoe UI""`.

Fixed by trimming quote characters alongside whitespace when splitting the family stack. Added `testFontFamilyStripsSurroundingQuotes` and `testFontFamilyStripsSingleQuotes` regression tests. Fixed in commit 9f489da60.

d4mation (approved):

PR #1125, opened 2026-07-09
Seed button border radius from the control-radius token via the preset catalog

pauloiankoski:
## Summary

Follow-up to the review question on PR #1112 (`discussion_r3539263596`): can the Kadence button's **Border Radius** seed its editor default from a design token via the new `Block_Preset_Catalog`, the same way `kadence/single-icon`'s `size` does?

Yes. This generalizes the catalog from scalar-only to per-entry **shape** and wires the button's `borderRadius` to `semantic.radius.control`.

## What changed

- **`Block_Preset_Catalog`** grows a per-entry `shape` (`SHAPE_SCALAR` / `SHAPE_CORNERS`, both private consts with accessors per convention). `all()` now emits either a bare float (icon size — unchanged output) or a 4-element float array for corner attributes.
- New entry: `kadence/singlebtn` → `borderRadius` → `semantic.radius.control` (corners shape). In the shipped baseline this resolves to `radius.md` = 0.5rem = 8px, so a freshly-inserted button seeds `[8, 8, 8, 8]`.
- Class docblock updated to record the button as the anticipated second consumer.

## Deliberately out of scope

- `tabletBorderRadius` / `mobileBorderRadius` — left empty; the block already inherits desktop when a breakpoint value is empty.
- `borderHoverRadius` / `borderTransparentRadius` / `borderStickyRadius` — the review question was about the base Border Radius, not hover/transparent/sticky states.
- `borderRadiusUnit` — untouched; the resolved value is already px, matching its own default.

## JS

No change to `src/early-filters.js`. `blockPresetAttributeDefault()` assigns the catalog value verbatim into the attribute default, so an array works unmodified.

## Testing

- `Block_Preset_CatalogTest` rewritten to a `Generator`-based provider covering both shapes: rem→px and px pass-through, each entry in isolation (scalar float vs. corner array), empty baseline, and unconvertible units. The real-container test asserts both the icon-size scalar (`24.0`) and the button corners (`[8, 8, 8, 8]`).
- PHPStan (level max): clean.
- slic wpunit `Block_Preset_CatalogTest`: `OK (7 tests, 7 assertions)`.
- cspell on changed files: clean.

## Base branch

Branched off `SOFT-3405/phase-3-editor-default-catalog` (PR #1112), which is the intended integration point.

### DIFF
--- a/includes/resources/Design_Tokens/Editor/Block_Preset_Catalog.php
+++ b/includes/resources/Design_Tokens/Editor/Block_Preset_Catalog.php
@@ -10,11 +10,12 @@ use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
  * early-filters.js reads to seed a freshly inserted block's attribute default from a resolved
  * token, instead of the hardcoded static default in block.json.
  *
- * Scoped to kadence/single-icon's `size` today — the only block/attribute pair this ticket needs.
- * Not a general "any block, any attribute" registry: extend the ENTRIES map (or promote to a
- * declarations-driven shape) only when a second real consumer needs it, matching this module's
- * existing preference for composable-but-not-speculative catalogs (see Variant_Catalog, which
- * itself started scoped to what the variant picker needed).
+ * Covers kadence/single-icon's `size` (a scalar pixel default) and kadence/singlebtn's
+ * `borderRadius` (a 4-corner pixel default) — the anticipated second consumer this module's
+ * composable-but-not-speculative preference (see Variant_Catalog) was waiting on before
+ * generalizing. Not a general "any block, any attribute" registry: keep extending the ENTRIES
+ * map (or promote to a declarations-driven shape) only when a further real consumer needs it,
+ * rather than opening it up ahead of demand.
  *
  * @since TBD
  */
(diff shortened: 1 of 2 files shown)


includes/resources/Design_Tokens/Editor/Block_Preset_Catalog.php
```diff
@@ -10,11 +10,12 @@
...
- * Not a general "any block, any attribute" registry: extend the ENTRIES map (or promote to a
- * declarations-driven shape) only when a second real consumer needs it, matching this module's
- * existing preference for composable-but-not-speculative catalogs (see Variant_Catalog, which
- * itself started scoped to what the variant picker needed).
+ * Covers kadence/single-icon's `size` (a scalar pixel default) and kadence/singlebtn's
+ * `borderRadius` (a 4-corner pixel default) — the anticipated second consumer this module's
+ * composable-but-not-speculative preference (see Variant_Catalog) was waiting on before
+ * generalizing. Not a general "any block, any attribute" registry: keep extending the ENTRIES
+ * map (or promote to a declarations-driven shape) only when a further real consumer needs it,
+ * rather than opening it up ahead of demand.
```
d4mation, line 18: I played around with this and I actually think this could be a bad idea in general. It sounded nice at first, but there's a legitimate issue with it:

Scenario:

1. Token is set to a scalar, example: `5`
2. A new block utilizing that Token is inserted into the Editor
    - This also impacts existing blocks _without_ the setting explicitly defined.
3. It populates the Editor _with the matching scalar value_
4. The Post is saved, thereby "freezing" that scalar value
5. If the Token is updated in the future, the inserted Block doesn't inherit it because "it had a hardcoded value for that in the Editor"

The way this was handled before for Button Border Radius would allow it to change with the Token value if one wasn't specifically defined in the Editor, but now it wouldn't be able to do that. This could be the case with the change for Icon Size as well, so it would be worth double-checking. Unfortunately, this could be considered a regression :( 

This may mean removing this code entirely, or reworking it so that it can work in a way similar to how Button Border Radius worked before where it doesn't _actually_ set the value for the Block but rather targets the CSS Variable. If one doesn't exist, it can be added (which is what I had to do for Button Border Radius).

See https://github.com/stellarwp/kadence-blocks/pull/1063/files#diff-85d12f9ae93a9d8abe07894cf761686b579f18746bf85f671a1bedc500f611db

d4mation (changes requested):

pauloiankoski:
Closing as unmergeable rather than unwanted.

This PR edits `includes/resources/Design_Tokens/Editor/Block_Preset_Catalog.php`, which no longer exists — it was replaced by `Attribute_Default_Catalog.php` when SOFT-3992 (#1193) landed. GitHub reports the PR as `CONFLICTING`, and there is no diff left to salvage.

**The capability it added has not shipped elsewhere.** It is not superseded by #1173 or #1202: those cover *presets* (saved styles a user applies), while this covered *editor defaults* (what a freshly inserted block starts with). `Attribute_Default_Catalog` today holds a single scalar entry —

```php
'kadence/single-icon' => [ 'size' => 'semantic.icon-size.default' ],
```

— and its own docblock notes it supports numeric defaults only. Nothing seeds the button's `borderRadius` from `semantic.radius.control`, and the per-corner shape support this PR introduced was dropped in the rewrite.

So the original question from the #1112 review — can the button's Border Radius seed its editor default from a token the way `single-icon`'s size does — is still unanswered in the code. Re-doing it against `Attribute_Default_Catalog` is a fresh, small piece of work rather than a rebase of this branch.

PR #1129, opened 2026-07-13
SOFT-3834 [1/2]: Collapse variant sets — PHP backend

d4mation:
🎫  https://linear.app/nexcess/issue/SOFT-3834/collapse-variant-sets-into-a-single-per-block-variant-list-per-variant

Primarily reverses the recently-shipped grouped / multi-axis variant work (SOFT-3533, SOFT-3577), collapsing the design-token variant model from multiple combinable axes (grouped sets keyed by block + group) back to one flat variant list per block. Backend half.

- Registry: one `Variant_Set` per block (the block's full bindable surface); the `group` concept and the `$single` sentinel are removed.
- Resolvers, projection, REST, catalog and admin feed drop all group branching. The emitted class collapses to `kb-variant--<variant>` and the CSS var to `--variant--<block>--<variant>--<property>`.
- A variant may define any subset of the block's bound properties; an unbound property is still rejected. The `variant_set` REST param is gone.
- `baseline.json` / `declarations.php` migrated off the grouped `.style.` shape.

Also fixes a **pre-existing** bug (present before this ticket, unrelated to the collapse): the per-variant scoped retarget rules were built from the active token set only, so a variant defined solely on a non-active set (e.g. a user-created variant on `dark`) never applied in the editor or on the front end. They are now emitted from every set's fragment.

PHPStan (level max) and the design-tokens wpunit suite are green.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
+++ b/includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
@@ -10,7 +10,6 @@ use KadenceWP\KadenceBlocks\Design_Tokens\Projection\Scope;
 use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Binding;
 use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Css_Var;
 use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Token_Registry;
-use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Variant_Set;
 use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Variant_Resolver;
 use RuntimeException;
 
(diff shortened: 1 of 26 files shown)


includes/resources/Design_Tokens/Projection/Variant/Css_Builder.php
```diff
@@ -504,57 +480,88 @@ private function point_canonical_at_set( array $collected, string $slug ): strin
...
+				if ( $declarations !== '' ) {
+					$css .= $data['selector'] . '.' . Style::variant_class( $variant ) . '{' . $declarations . '}';
+				}
+			}
+		}
 
-				foreach ( $group_data['variants'] as $variant => $properties ) {
-					$declarations = '';
+		return $css;
+	}
```
d4mation, line 509: This is related to the bugfix mentioned in the PR description.

`testItEmitsTheScopedRuleForAVariantOnlyInANonActiveSet()` has been added to `Css_BuilderTest` to exercise it.

pauloiankoski (approved):

PR #1133, opened 2026-07-14
[DTM] SOFT-3849 [3/5]: indicator UI + singlebtn wiring

pauloiankoski:

Phase 3 of 4 (stacked). Base: `SOFT-3849/phase-2-editor-binding-state`.

## Summary

The indicator UI and its wiring into the Single Button controls.

- **Components** (`src/extension/token-indicators/components/`): `TokenIndicator` (design-system icon when bound; override dot + reset affordance when overridden; nothing when unmapped), `TokenLabel` (label text + indicator as a React node), `TokenControlRow` (renders the indicator adjacent to a control).
- **Highlight-edits store** (`src/extension/token-indicators/store.js`) — a small `@wordpress/data` store; `TokenIndicator` highlights overridden controls when it is on. (The toggle that flips it ships in Phase 4.)
- **Two wiring mechanisms, by control type.** `ResponsiveMeasurementControls` (Border Radius) renders its `label` prop as content, so the indicator is injected as a `TokenLabel` node. `PopColorControl` (the color swatches) forwards `label` to a `Button` **tooltip**, not content — so a node there would not render and would corrupt the aria label. The 4 color swatches instead get a `TokenControlRow` sibling, leaving each swatch's own `label`/`value`/`onChange` untouched.
- **Single Button wiring** (`src/blocks/singlebtn/edit.js`): `borderRadius` via `TokenLabel`; `color`, `background`, `colorHover`, `backgroundHover` via `TokenControlRow`. The unmapped hover-radius control is deliberately left alone.

`// TODO: confirm exact placement vs Figma node 160-12291` markers sit at each swatch wiring site — the adjacent-placement default is implemented; the exact visual is to be confirmed against the Figma.

## Test plan

- `npm run lint-js` (eslint) and cspell green on all changed files.
- Production `npm run build` compiles; the new `kb-token-*` classes are present in `dist/blocks-singlebtn.css`.
- Editor check (after the chain merges + baseline flush): select a variant on a Single Button and confirm the design-system icon appears on the 5 bound controls; override one and confirm it flips to the override state and reset returns it to the variant value.

### DIFF
--- a/src/blocks/singlebtn/edit.js
+++ b/src/blocks/singlebtn/edit.js
@@ -87,6 +87,9 @@ import BackendStyles from './components/backend-styles';
 import { VariantPicker, blockVariants, activeSet } from '../../extension/variant-picker';
 import { VariantActions } from '../../extension/variant-picker/VariantActions';
 import { TokenSetPicker, selectableSets } from '../../extension/token-set-picker';
+import { useVariantBinding, resetAttr } from '../../extension/token-indicators';
+import { TokenLabel } from '../../extension/token-indicators/components/TokenLabel';
+import { TokenControlRow } from '../../extension/token-indicators/components/TokenControlRow';
 
 export default function KadenceButtonEdit(props) {
 	const { attributes, setAttributes, isSelected, context, clientId, name } = props;
@@ -263,6 +266,11 @@ export default function KadenceButtonEdit(props) {
 	);
 	const marginMouseOver = mouseOverVisualizer();
 	const paddingMouseOver = mouseOverVisualizer();
+
+	// Design-token indicators: the per-attribute bound/overridden state for the selected variant, plus a
+	// reset that clears the mapped attribute back to the variant value (served by the existing scoped CSS).
+	const tokenBinding = useVariantBinding('kadence/singlebtn', attributes);
+	const resetToken = (attr) => resetAttr(attr, setAttributes, tokenBinding[attr]?.kind);
 
 	useEffect(() => {
 		setAttributes({ inQueryBlock: getInQueryBlock(context, inQueryBlock) });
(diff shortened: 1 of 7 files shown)


src/extension/token-indicators/components/TokenControlRow.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * A row wrapper that places the design-system indicator BESIDE a control whose `label` prop is not
+ * rendered as visible content (e.g. `PopColorControl`, which forwards `label` to a Button tooltip). The
+ * wrapped control is passed through as children with all its own props intact; only an adjacent indicator
+ * (and an optional heading) is added. Renders just the children when the attribute is not mapped for the
+ * selected variant, so an unmapped control looks identical to today.
```
d4mation: It looks like currently it always renders `__header` + the heading if nothing is mapped for the variant.

We should ensure that the doc here matches reality to prevent confusion. Or would this be fixed by your TODOs?

src/blocks/singlebtn/edit.js
```diff
@@ -1124,15 +1158,23 @@ export default function KadenceButtonEdit(props) {
...
+															// TODO: confirm exact placement vs Figma node 160-12291
+															<TokenControlRow
+																heading={__('Background Color', 'kadence-blocks')}
+																attr="background"
+																binding={tokenBinding}
+																onReset={resetToken}
+															>
+																<PopColorControl
+																	key={'btnbg'}
+																	label={__('Background Color', 'kadence-blocks')}
```
d4mation: Should this still have a `label`? Maybe the `<TokenControlRow>` could be mapped via `<label for="">` to these elements so that it replaces the built-in label? Or if `<TokenControlRow>` itself rendered as a `<label>` wrapper for this element, it may be able to accomplish the same. It would be worth testing.

src/extension/token-indicators/components/TokenControlRow.js
pauloiankoski: Fixed in e3f2edbff — doc now matches reality: the header always renders; only the indicator is conditional.

src/blocks/singlebtn/edit.js
pauloiankoski: Fixed in e3f2edbff — dropped the duplicate label (`label` → `swatchLabel`), so the heading is the single label.

d4mation (approved):

PR #1135, opened 2026-07-14
Remove Design System impact on WP Core Blocks (retire the Native Projector)

d4mation:
🎫 https://linear.app/nexcess/issue/SOFT-3836/remove-design-system-impact-on-wp-core-blocks-retire-the-native

Removes the design system's footprint on non-Kadence blocks, so tokens only affect Kadence blocks.

- Deletes the Native projector (the `.wp-block-button` styler), the `core/button` variant (declarations + baseline), the `core` selector special-case, and the JS opt-in.
- Retires the Theme.json token projection and the `Global_Styles` two-way sync subsystem, and removes the now-unused `wp_preset` projection vocabulary. theme.json presets are inherently global; the Style Book edits tokens over the `kb-design-tokens/v1` REST API with no theme.json dependency, so token editing is preserved.

The KB custom-color palette feature ("Use only Custom Colors") and Kadence-block token/variant behavior are unchanged. The keep-vs-restrict decision is recorded on the ticket. PHPStan (max), wpunit, and snapshot suites pass.

### Checklist


pauloiankoski:
@d4mation The `site_editor` projection is now obsolete because its only consumer was the deleted `Global_Styles` subsystem, so `site_editor => true` on the button tokens is now inert. Maybe we could remove it (like `wp_preset`) or document why it's being kept.

pauloiankoski (approved):

d4mation:
> @d4mation The `site_editor` projection is now obsolete because its only consumer was the deleted `Global_Styles` subsystem, so `site_editor => true` on the button tokens is now inert. Maybe we could remove it (like `wp_preset`) or document why it's being kept.

@pauloiankoski Good catch! Fixed in [7f64bbf](https://github.com/stellarwp/kadence-blocks/pull/1135/commits/7f64bbf7538887c7b1cdd483a9b44a68bea19a74)

pauloiankoski (approved):

PR #1136, opened 2026-07-14
docs: document the JS `@since TBD` JSDoc convention

pauloiankoski:
Documents the JavaScript `@since TBD` JSDoc convention in `AGENTS.md` (surfaced as `CLAUDE.md`), so it is enforced in review the same way the PHP rule already is.

## Why

`AGENTS.md` spelled out `@since TBD` only for PHP, even though the repo's JS already uses it (`src/early-filters.js`, `src/style-book/*`). New JS could silently miss the tag. This adds a **JavaScript coding conventions → Docblocks** section that:

- requires `@since TBD` on every new JS function, hook, component, store selector/action/reducer, and exported helper (only the code a change introduces, not pre-existing functions);
- documents the repo-specific placement: the `@since TBD` line goes **after** the `@param` block and **before** `@return` (unlike PHP, which puts it before `@var`), matching `src/early-filters.js`.

Docs only — no code changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -96,6 +96,33 @@ the team enforces in review; follow them exactly.
   `kadence/single-icon`'s `color`) — that shape fits the low-specificity `Block_Default_Css`
   binding directly, with no adapter and no unit conversion needed.
 
+## JavaScript coding conventions
+
+### Docblocks
+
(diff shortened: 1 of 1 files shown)


pauloiankoski:
Folded into the Phase 4 PR (#1134) via cherry-pick — no separate branch needed.

PR #1137, opened 2026-07-14
[DTM] SOFT-3849 [5/5]: indicator design alignment (Figma)

pauloiankoski:

Phase 5 of 5 (stacked). Base: `SOFT-3849/phase-4-picker-actions`.

## Summary

Presentation-only refinement that aligns the token indicators with the Figma design reference (node `160-12291`) and resolves the swatch-placement TODOs left in Phase 3. No detection/logic changes — it reuses the Phase 2 `useVariantBinding` hook unchanged.

- **Bound state** now renders a small design-system **icon** (right-aligned at the end of each control row) instead of a colored dot next to the section label — matching how the Figma places the indicator on every bound control.
- **Overridden state** keeps the icon with a small **red corner dot**, and the reset affordance is an icon-only **undo-arrow** button instead of a "Reset" text link.
- Right-alignment is handled in CSS (`justify-content: space-between` on the control-row header); `TokenControlRow` markup is unchanged.
- Removed the 4 `// TODO: confirm exact placement vs Figma` markers in `singlebtn/edit.js`.

**Icon choice:** the bound state uses `@wordpress/icons`' `styles` glyph (WordPress's own Global Styles mark) as a design-system stand-in, and `undo` for reset — both noted in code as swappable once design exports the exact Figma asset.

**Deliberately deferred** (documented in the phase plan, not built here): a per-section override **count badge** (the block's panels don't map 1:1 to the Figma sub-sections, and only 5 controls are mapped this pass); the `Stylebook`/`Custom` control dropdowns; showing the variant's value-name inside controls; the top-of-card variant selector; and the "Variant" → "Preset" wording (a product-copy decision).

## Test plan

- `npm run lint-js` (eslint) + cspell green on all changed files.
- Production `npm run build` compiles and regenerates `dist/blocks-singlebtn.*`.
- Editor check: select a variant on a Single Button and confirm the design-system icon sits right-aligned on each bound control; override one and confirm the red dot + undo-arrow reset appear and reset works.

### DIFF
--- a/src/blocks/singlebtn/edit.js
+++ b/src/blocks/singlebtn/edit.js
@@ -913,7 +913,6 @@ export default function KadenceButtonEdit(props) {
 															/>
 														)}
 														{'normal' === textBackgroundHoverType && (
-															// TODO: confirm exact placement vs Figma node 160-12291
 															<TokenControlRow
 																heading={__('Color Hover', 'kadence-blocks')}
 																attr="colorHover"
@@ -949,7 +948,6 @@ export default function KadenceButtonEdit(props) {
 															/>
 														)}
 														{'normal' === backgroundHoverType && (
-															// TODO: confirm exact placement vs Figma node 160-12291
 															<TokenControlRow
 																heading={__('Background Color', 'kadence-blocks')}
 																attr="backgroundHover"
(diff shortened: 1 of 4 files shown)


pauloiankoski:
Reverting Phase 5 — keeping SOFT-3849 to the original 4 phases. The design-alignment work is dropped for now.

PR #1140, opened 2026-07-15
SOFT-3529 [3/3] Wire kadence/singlebtn typography to a control token

d4mation:
🎫  https://linear.app/nexcess/issue/SOFT-3529/button-typography-token-wiring-kadenceadvancedbtn-singlebtn

Wires `kadence/singlebtn` typography to a new `semantic.typography.control` token, seeded to the on-brand font family. A render-instance adapter fills the button's `typography[0]` bundle from the resolved token with field-level local-wins on the per-instance render filter, so an uncustomized button follows the token while any per-button value still wins — with no change to the button's render path or stylesheet. Also adds a render-instance adapter category alongside the existing registration-default one, and guards the variant value normalizer against aliasing a value onto an empty-resolving semantic.

### Checklist


d4mation:
Superseded by #1141 — reworked from a front-end-only render adapter to CSS-variable delivery so button typography renders in the editor as well, with no frozen values.

PR #1144, opened 2026-07-15
[DTM] SOFT-3849 [5/5]: preset button, dropdown, and inline indicator redesign

pauloiankoski:
🎥  https://www.loom.com/share/5ca3795e4f4143b7b5203e6900e76e03

Phase 5 of 5 (stacked). Base: `SOFT-3849/phase-4-picker-actions`.

## Summary

Redesigns the design-token controls in the Single Button inspector.

- Removes the old "Design Tokens" panel from the Style tab. A single **preset button** now sits above the inspector tabs, right after the block card: it shows the current preset name, the variant icon, and an "(Edited)" label plus a red edit dot when a control overrides the preset.
- Clicking the button opens a **dropdown**: the block's presets (the current one checked) above Highlight Edits / Reset / Save As a New Preset. A separate **reset button** sits next to the preset button and clears every override.
- Switching a preset — or saving a new one — clears the overrides, so edits never carry over to the next preset.
- **Save As a New Preset** captures the block's current values (the selected preset plus your edits) and stores them as the new preset. The modal is only a name field; value editing lives elsewhere.
- Reworks the **inline control indicators**: a 4×4 red dot marks an overridden control, the per-control reset is an icon in the control-row header, and Highlight Edits tints the whole overridden control with a warning color. The color pickers' native "clear" button is hidden so they rely on the header reset.
- All user-facing wording is **"preset"** instead of "variant"; internal identifiers are unchanged.

## Test plan

- eslint + cspell green on every changed file; production build compiles.
- Editor: on a Single Button, open the preset dropdown, switch presets, edit a color, and confirm the red dot and "(Edited)" appear. Reset via the header icon and via the reset button. Save As a New Preset and confirm the new preset matches the edit with no leftover "(Edited)" state.


src/blocks/singlebtn/edit.js
```diff
@@ -921,7 +898,8 @@ export default function KadenceButtonEdit(props) {
 																onReset={resetToken}
 															>
 																<PopColorControl
-																	key={'btncolorhover'}
+																	key={`btncolorhover-${colorResetNonce}`}
+																	hideClear
```
d4mation: While unlikely to happen, what would happen if this control is not actually backed by a Token in the future or due to some bug? Then the built-in reset button wouldn't show. 

We should probably have some kind of check here to ensure it is Token-backed before hiding the built-in reset button.

src/extension/variant-picker/VariantButton.js
```diff
@@ -0,0 +1,235 @@
...
+	const variants = blockVariants(blockName, tokenSet);
+
+	if (!variants.length) {
+		return null;
+	}
+
+	const selected = get(attributes, 'kbVariant', '');
+	const currentSlug = selected || blockDefaultVariant(blockName, tokenSet);
+	const label = currentVariantLabel(blockName, tokenSet, selected);
+	const edited = Object.values(useVariantBinding(blockName, attributes)).some((entry) => entry.overridden);
```
d4mation: `useVariantBinding()` re-derives the token set from the attributes rather than using the `set` property. Maybe we should use it instead to ensure consistency so there's never a case where the "edited" state ends up mismatching reality.

src/blocks/singlebtn/edit.js
```diff
@@ -1203,7 +1184,6 @@ export default function KadenceButtonEdit(props) {
 																	text={__('Border Radius', 'kadence-blocks')}
 																	attr="borderRadius"
 																	binding={tokenBinding}
-																	onReset={resetToken}
```
d4mation, line 1206: If we're removing this, the control needs to be wrapped in `TokenControlRow` so that it has the reset control, right?

src/extension/variant-picker/VariantButton.js
```diff
@@ -0,0 +1,235 @@
...
+										</MenuItem>
+									);
+								})}
+							</MenuGroup>
+							<MenuGroup>
+								<MenuItem
+									className="kb-variant-button__action"
+									role="menuitemcheckbox"
+									aria-checked={highlighting}
+									disabled={!edited}
```
d4mation, line 187: What happens if highlighting is on but then the user resets back to the chosen Variant? They won't be able to disable highlighting again until a change has been made. This could be confusing UX.

src/extension/variant-picker/VariantButton.js
d4mation: This is called after the early return via `if  (!variants.length)` which would violate react hook rules. If `react-hooks/rules-of-hooks` were enabled in `.eslintrc.js` it probably could have caught this 🤔 

https://react.dev/warnings/invalid-hook-call-warning#breaking-rules-of-hooks

src/extension/token-indicators/token-indicators.scss
```diff
@@ -54,17 +59,33 @@
...
+.kb-token-control-row--highlight {
+	&::before {
+		background: rgba(240, 184, 73, 0.25);
+		content: '';
+		height: calc(100% + 16px);
+		left: -16px;
+		position: absolute;
+		top: -8px;
+		width: calc(100% + 32px);
+		z-index: -1;
```
d4mation: This `z-index` is strange. It could result in the highlight being painted below another opaque background and then becoming invisible.

Maybe it would be better to have something like this:

```scss
.kb-token-control-row {
	position: relative;

	&__highlight {
		position: absolute;
		top: 0;
		right: 0;
		bottom: 0;
		left: 0;
		background: rgba(240, 184, 73, 0.25);
		content: '';
	}
	
	&__content {
		position: relative;
		padding: 8px 16px; // This way you're not doing the weird left/top positioning.
	}
}
```

src/extension/token-indicators/components/TokenIndicator.js
```diff
@@ -1,64 +1,71 @@
...
-				variant="tertiary"
-				isSmall
-				onClick={onReset}
-				label={__('Reset to variant value', 'kadence-blocks')}
-				showTooltip
-			>
-				{__('Reset', 'kadence-blocks')}
-			</Button>
+		<span className="kb-token-indicator">
+			<span className="kb-token-indicator__dot" aria-hidden="true" />
```
d4mation: This dot will not play nicely with screenreaders. We should remove `aria-hidden` and add an `aria-label` describing what it signifies. However, that alone would only solve for screenreaders. Truly, a tooltip would be best. Because while _we_ know what the dot means, our users will probably have no idea.

src/blocks/singlebtn/edit.js
```diff
@@ -838,49 +858,6 @@ export default function KadenceButtonEdit(props) {
 
 						{activeTab === 'style' && (
 							<>
-								{(blockVariants(name, attributes.kbTokenSet || activeSet()).length > 0 ||
```
d4mation, line 841: Something we need to consider here (maybe it would be something we talk to Donaldo about today in our call) is that by removing this, we now have no way to tell this system "I want to use a Variant from this Token Set for this specific button" which the old interface _did_ provide. We will need a per-Block way to change the Token Set to be used.

src/extension/variant-picker/capture.js
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	return normalizeText(value);
+}
+
+/**
+ * The block's current values across its preset surface, as a `{ propertyKey: literal }` token map: each
+ * mapped control's edited value when it has one, else the selected preset's value. Feeds the "save as a
+ * new preset" write so the new preset matches what the editor currently shows.
+ *
```
d4mation, line 45: Saving to a literal should only be done if a non-Token value was chosen. If a new Variant had chosen the Secondary Primitive Color as its Text Color, for instance, then in the future changing that Secondary Primitive Color should automatically update this new Variant. If we store the literal and assign that in the editor, then we lose that reference.

This may require other refactoring, which we're discussing with Donaldo today. Just calling this out as I see it :)

src/blocks/singlebtn/edit.js
```diff
@@ -290,6 +287,27 @@ export default function KadenceButtonEdit(props) {
 		}
 	}, [isSelected]);
 
+	// A remount key for the token-mapped color pickers. PopColorControl keeps the picked color in its own
+	// internal state, so clearing the attribute (a per-control revert or the variant Reset-all) updates the
+	// block but not the swatch. Bumping this key when a mapped color attribute goes from set -> empty
+	// remounts the picker, which re-reads the now-empty attribute — the missing wire for token reset.
+	const [colorResetNonce, setColorResetNonce] = useState(0);
```
d4mation, line 296: This is quite the sledgehammer to get around the fact that `hideReset` effectively removes our ability to alter the internal state.

I'm already looking at a need for us to update `@kadence/components` due to https://linear.app/nexcess/issue/SOFT-3870/token-backed-palette-colors-hidden-in-kadence-color-controls-when-use, so we may as well add something like this to the `PopColorControl` component so that changing the value to match our Tokens would forcibly reload the control.

```js
componentDidUpdate( prevProps ) {
      const prevValue = prevProps.colorValue ? prevProps.colorValue : prevProps.value;
      const value     = this.props.colorValue ? this.props.colorValue : this.props.value;

      // An external reset cleared the value out from under our internal swatch state; re-read it.
      if ( prevValue && ! value && ! this.state.reload ) {
              this.setState( { reload: true } );
      }
}
```

src/extension/variant-picker/VariantButton.js
pauloiankoski: Fixed in bb1221c. `useVariantBinding` now accepts an explicit set and `VariantButton` passes its resolved `tokenSet`, so the edited state and the rest of the dropdown always agree. Callers that don't pass a `set` keep the old fallback.

pauloiankoski: Fixed in df0a025. The binding call now runs before the early return. Also re-enabled react-hooks/rules-of-hooks as an error scoped to the design-token folders in d3639e8, so lint catches this class of bug going forward.

src/extension/variant-picker/VariantButton.js
pauloiankoski: Fixed in 39c263a. When no overrides remain, the highlight flag now clears itself — with nothing to highlight there's nothing for the toggle to do, so it can never be stuck on in the disabled state.

src/extension/token-indicators/components/TokenIndicator.js
pauloiankoski: Fixed in 2ee63e9. The dot now has a `Tooltip` and an `aria-label ("Overrides the preset value")`, same treatment as the matches-preset mark.

src/blocks/singlebtn/edit.js
pauloiankoski: Fixed in c34df1e. All four color pickers now only hide the native clear when the attribute is actually bound (`tokenBinding.<attr>?.bound`), so an unbound control keeps its own clear as the fallback.

src/extension/token-indicators/token-indicators.scss
pauloiankoski: Fixed in 88e5c31. The tint is now an inset overlay at `z-index: 0` with the row content lifted to `z-index: 1`, so there's no negative z-index and no `calc()` sizing. Kept it CSS-only to avoid touching the `TokenControlRow` markup.

src/blocks/singlebtn/edit.js
pauloiankoski: Agreed the right fix is in `PopColorControl` alongside the SOFT-3870 work. Added a TODO in 2bd0388 marking the nonce as a stopgap to delete once that component update lands.

src/blocks/singlebtn/edit.js
pauloiankoski: Restored in 45fe65f. Border Radius now carries the indicator and a one-click reset (all devices + unit) inline in its own header row, next to the device tabs.

CleanShot 2026-07-17 at 15 08 38

d4mation (approved):

PR #1148, opened 2026-07-16
SOFT-3857 [3/8] Publish the responsive/clamp shape in the DTCG JSON schema

d4mation:

Part of the SOFT-3857 stack (3/8) — stacked on 2/8.

Publishes the responsive/clamp shape on the dimension and lineHeight branches of the generated DTCG JSON schema and regenerates the committed `dtcg.schema.json`. Byte-exact conformance holds and the shipped baseline still validates against both engines.

### Checklist

### DIFF
--- a/dev_scripts/src/Design_Tokens/Schema/Dtcg_Schema_Generator.php
+++ b/dev_scripts/src/Design_Tokens/Schema/Dtcg_Schema_Generator.php
@@ -21,6 +21,8 @@ namespace KadenceWP\KadenceBlocks\Dev\Design_Tokens\Schema;
 // cspell:ignore justinrainbow .
 
 use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Alias;
+use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Extensions;
+use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Responsive;
 use KadenceWP\KadenceBlocks\Design_Tokens\Schema\Vocabulary\Token_Type;
 
 /**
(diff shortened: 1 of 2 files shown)


dev_scripts/src/Design_Tokens/Schema/Dtcg_Schema_Generator.php
```diff
@@ -247,6 +257,69 @@ private function composite_shape( string $type ): array {
...
+							Responsive::get_clamp_key()      => [
+								'type'                 => 'object',
+								'required'             => [
+									Responsive::get_clamp_min_key(),
+									Responsive::get_clamp_preferred_key(),
+									Responsive::get_clamp_max_key(),
+								],
+								'properties'           => [
+									Responsive::get_clamp_min_key()       => $slot,
+									Responsive::get_clamp_preferred_key() => [ 'type' => 'string' ],
```
pauloiankoski, line 309: For lineHeight, `min`/`max` accept a number but `preferred` is string-only, while PHP accepts a numeric preferred. Should `preferred` also allow a number here?

d4mation: `preferred` in `clamp()` should always be a string because it is the part that actually changes as the container width changes. `min` is a bottom clamp to ensure it can never go smaller than that and `max` is so it can't go larger than that, but the `preferred` bit is what actually does the scaling. 

Because of this, `preferred` always needs to be a relative expression like `0.5vw`, `50%`, or even something like `0.25vw + 2` which is a string by nature. 

While the way we have it written means `preferred` cannot be validated as an Alias to another value, that may be more edge case and we could look at supporting it in the future if needed similar to how `min` and `max` support either an Alias or a Literal.

pauloiankoski (approved):

PR #1151, opened 2026-07-16
SOFT-3857 [6/8] Surface authored responsive shapes in the admin feed

d4mation:

Part of the SOFT-3857 stack (6/8) — stacked on 5/8.

Surfaces each token's raw authored responsive/clamp shape in the admin feed (`Responsive_Feed` + `Localizer`/`Builder`) so the Style Book editor can hydrate its per-breakpoint inputs exactly as authored rather than from the flat resolved value.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
@@ -473,7 +473,9 @@ final class Documents_Controller extends Controller {
 				'version'    => $this->store->get_version( $slug ),
 				'by_id'      => $resolved->by_id(),
 				'by_var'     => $resolved->by_var(),
-				'responsive' => $responsive,
+				// An empty PHP array JSON-encodes as [] rather than {}; cast the empty case to an object so
+				// the responsive map keeps a stable object wire type whether or not a token carries the shape.
+				'responsive' => empty( $responsive ) ? (object) [] : $responsive,
 			],
 			WP_Http::OK
 		);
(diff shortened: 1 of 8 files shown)


includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
```diff
@@ -446,12 +465,15 @@ public function get_resolved( $request ) {
...
 			[
-				'slug'    => $slug,
-				'version' => $this->store->get_version( $slug ),
-				'by_id'   => $resolved->by_id(),
-				'by_var'  => $resolved->by_var(),
+				'slug'       => $slug,
+				'version'    => $this->store->get_version( $slug ),
+				'by_id'      => $resolved->by_id(),
+				'by_var'     => $resolved->by_var(),
+				'responsive' => $responsive,
```
pauloiankoski: When empty, `responsive` is serialized as `[]` even though the schema declares it as an object. JS clients won’t notice, but casting the empty case to an object would keep the wire type stable. Up to you, though. This only matters for REST response schema validation, which we don’t do (the schema is only served for route discovery), so it’s more documentation drift than a bug.

pauloiankoski (approved):

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
d4mation: 615d03e0bcc6eedda7d983e6a99d87de01a09a09

pauloiankoski (approved):
LGTM!

PR #1152, opened 2026-07-16
SOFT-3857 [7/8] Per-breakpoint token editor in the Style Book

d4mation:

Part of the SOFT-3857 stack (7/8) — stacked on 6/8.

Adds a responsive-aware token field to the Style Book: per-breakpoint (desktop/tablet/mobile) inputs plus an optional fluid clamp helper. `buildTokenLeaf` serializes the `$extensions` shape and omits it entirely when empty, so a desktop-only edit round-trips as a clean flat leaf.

### Checklist


pauloiankoski (approved):
This only affects the Style Book, which will almost certainly be reworked soon. Still, just as a heads-up: if `Fluid` is enabled but one of the clamp values is left empty, `Save` writes `$value`: `'clamp(, , )'`, which is invalid CSS. Also, there are a few new functions missing `@since TBD` tags.

Image

d4mation:
@pauloiankoski I've tweaked the Style Guide page to ideally not allow this, but more importantly I tightened up the REST surface itself so that it should reject these cases :) That way it won't be possible once we redo the Style Guide page.

pauloiankoski (approved):

PR #1154, opened 2026-07-17
[DO NOT USE]License modal that allows user to enter both legacy and unified license key 

pramodjodhani:
🎫  [SMTNC-1800](https://linear.app/nexcess/issue/SMTNC-1800/kadence-blocks-license-popup)

...

Functionality demo: 
1. https://www.loom.com/share/e0870d74982a4dd9830e8471c03abe26
2. https://www.loom.com/share/008933a1a0274df582f70d015e4c1a50

Sorry for the large PR, I understand it will be a little hard to review that's why I recorded a code demo so you know which code is doing what: https://www.loom.com/share/5bd3c2f5199a40789e14d047af390147

## Updates
- Single Modal to add Legacy and Unified License
- For unified license:
	- one click install and activate pro version
	- one click activate the site domain
	- Guide user towards the legacy field if they try to enter the legacy license in unified license field
- License information widget on the sidebar

See loom or screenshots below for details. 

## Related
This PR is related with this PR for Blocks Pro: https://github.com/stellarwp/kadence-blocks-pro/pull/298

## Popup states
1. No license key added
image

2. Legacy license added to Unified field
image

3. Legacy key field
image

4. Post License activation
image
 
5. All set
image

## License Widget Statuses
1. When Unified License is not active
image

2. When Unified License is active and domain is connected
image

3. When License is connected but domain is not connected:

4. When License is connected, Blocks pro is installed but not active:
image

5. When License is active but Kadence block Pro is not installed
image

6. When Legacy License is connected
image


includes/settings/class-kadence-blocks-settings.php
```diff
@@ -1453,28 +1506,98 @@ public function config_page() {
...
+				</ul>
+				<a href="https://www.kadencewp.com/kadence-blocks/pro/?utm_source=in-app&utm_medium=kadence-blocks&utm_campaign=dashboard" target="_blank" class="sidebar-btn-link"><?php esc_html_e( 'Upgrade Kadence Blocks', 'kadence-blocks' ); ?></a>
+			</div>
+		</div>
+		<?php
+	}
+	/**
+	 * Activate the Kadence Blocks Pro plugin via ajax.
+	 */
+	public function ajax_activate_pro(): void {
```
redscar: **blocking**

Pretty sure this whole AJAX path is dead. The JS binds to `.kt-activate-pro-button` and posts the `kadence_blocks_activate_pro` action, but nothing in this PR renders that class anywhere, and the nonce it's checking against (`kadence-blocks-activate-pro`) doesn't match the one you're actually localizing (`kadence-blocks-manage`). I searched the rest of the repo too and came up empty. My guess is this got superseded by the Harbor `installAndActivateFeature` flow in `active-license-view.js` and just never got cleaned up. Can we pull it, or if there's markup for it planned somewhere I'm missing, let's wire it up before this merges — as-is it's just confusing dead weight.

src/assets/js/admin-scripts.js
```diff
@@ -560,3 +560,45 @@ jQuery(document).ready(function ($) {
...
 	});
 })(jQuery, window, document);
+/**
+ * Ajax activate the Kadence Blocks Pro plugin.
+ */
+(function ($, window, document, undefined) {
+	'use strict';
+
+	$(function () {
+		$('.kt-activate-pro-button').on('click', function (event) {
```
redscar: **blocking**

Same issue as the `ajax_activate_pro` handler on the PHP side — this click handler has no matching markup anywhere in the repo, and the nonce action name it's posting doesn't match anything we localize. Looks like dead code left over from an earlier iteration.

includes/resources/Harbor/Licensing/Abstract_License_Strategy.php
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * Mask a license key for display (e.g. LWSW-••••••••••••-9KQ2).
+	 *
+	 * @since TBD
+	 *
+	 * @param string $key The full license key.
+	 *
+	 * @return string
+	 */
+	protected function mask_key( string $key ): string {
```
redscar, line 46: **should fix**

None of the new license logic has tests — `mask_key()` here, `License_Status::resolve()`, `get_expires_label()`, `is_harbor_rest_request()`/`get_harbor_rest_path()`. These are small and pure, which makes them cheap to cover, and also exactly the kind of thing that quietly breaks on an edge case (short keys, both-licenses-active, a filtered `rest_url_prefix`). Would like to see at least unit tests on these before merge.

src/license-modal/active-license-view.js
```diff
@@ -0,0 +1,215 @@
...
+				: sprintf(
+						/* translators: %s: plugin name */
+						__('Preparing to install %s…', 'kadence-blocks'),
+						PRO_FEATURE_NAME
+					)
+		);
+
+		try {
+			await installAndActivateFeature(PRO_FEATURE_SLUG, {
+				onStatus: (status) => {
```
redscar: **nit**

The install/activate status copy (`installing` / `activating` / `already_active`) is duplicated near verbatim between this file and `unified-license-view.js`. Feels like it wants to be one shared hook in `src/harbor` so the two views can't drift from each other next time this copy changes.

includes/resources/Harbor/Licensing/Unified_License_Strategy.php
```diff
@@ -0,0 +1,114 @@
...
+	 * {@inheritdoc}
+	 */
+	public function get_type(): string {
+		return 'unified';
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function get_key(): string {
```
redscar, line 45: **nit**

`get_ui_status()` and `License_Status::resolve()`/`get_unified_key()` each end up calling `get_key()`/`is_active()` independently, and `is_active()` calls `get_key()` again itself — so a single request can hit `lw_harbor_is_product_license_active()`/`lw_harbor_get_unified_license_key()` multiple times. Probably fine today, but worth memoizing on the strategy instance before one of those calls gets heavier.

includes/resources/Harbor/Harbor_Provider.php
```diff
@@ -35,14 +34,8 @@ public function register(): void {
...
 		add_filter( 'kadence_blocks_ai_disabled_message', [ $this, 'ai_disabled_message' ] );
 
-		foreach ( ( new Get_Known_Plugins() )() as $slug => $plugin ) {
-			add_action(
-				"stellarwp/uplink/{$slug}/license_field_after_form",
-				new Render_Harbor_License_Notice( $plugin['name'] )
-			);
-
-			add_filter( "stellarwp/uplink/{$slug}/plugin_notices", [ $this, 'suppress_inline_license_notices' ] );
-		}
```
jonwaldstein, line 45: @pramodjodhani can you confirm this is intentional? in case you don't know this, every kadence add-on has its own license field from uplink.  This filter is how we're able to hook into all of them from the kadence blocks library and not have to update each one individually.

I recommend installing all kadence add-ons while working on this so you can see the effect they all have

includes/resources/Harbor/Harbor_Provider.php
```diff
@@ -124,6 +123,75 @@ class_exists( ITSEC_Core::class ) &&
...
+	 * Allowed:
+	 * - /liquidweb/harbor/v1/license[/*] (license modal get/store/refresh)
+	 * - /liquidweb/harbor/v1/features/kadence-blocks[/*]
+	 * - /liquidweb/harbor/v1/features/kadence-blocks-pro[/*]
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	private function is_harbor_rest_request(): bool {
```
jonwaldstein, line 161: What's the purpose of this exactly? We just want to be careful of hardcoding references to harbor as they can change between versions & leaders

includes/resources/Harbor/Licensing/Unified_License_Strategy.php
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * Formatted expiration label for the active unified Kadence entitlement.
+	 *
+	 * @since TBD
+	 *
+	 * @return string Localized "Expires on …" label, or empty string when unknown.
+	 */
+	private function get_expires_label(): string {
+		$state = get_option( 'lw_harbor_licensing_products_state', [] );
```
jonwaldstein: Do you think this would be valuable to come from the harbor package itself? we want to be careful about hardcoding references to harbor itself as it was intended to supply global functions to plugin consumers as needed

includes/resources/Harbor/Harbor_Provider.php
```diff
@@ -124,6 +123,75 @@ class_exists( ITSEC_Core::class ) &&
...
+
+	/**
+	 * Extract a normalized Harbor REST path from the current request, if any.
+	 *
+	 * @since TBD
+	 *
+	 * @return string Path like "liquidweb/harbor/v1/features/kadence-blocks-pro", or empty.
+	 */
+	private function get_harbor_rest_path(): string {
+		$namespace = 'liquidweb/harbor/';
```
jonwaldstein, line 187: This is another place I don't love that we're hardcoding, feels risky

includes/helper-functions.php
```diff
@@ -228,18 +223,7 @@ function kadence_blocks_get_authorized_license_key(): string {
...
-		&& function_exists( 'lw_harbor_is_product_license_active' )
-		&& lw_harbor_is_product_license_active( 'kadence' )
-	) {
-		$unified_key = lw_harbor_get_unified_license_key();
-		if ( ! empty( $unified_key ) ) {
-			return $unified_key;
-		}
-	}
-
-	return '';
```
jonwaldstein, line 242: Are you sure these should be returning the same logic?   There are some slight differences to confirm.

includes/settings/class-kadence-blocks-settings.php
```diff
@@ -802,6 +815,45 @@ public function scripts() {
...
+		$license_modal_meta = kadence_blocks_get_asset_file( 'dist/admin-license-modal' );
+		wp_enqueue_script( 'admin-kadence-license-modal', KADENCE_BLOCKS_URL . 'dist/admin-license-modal.js', $license_modal_meta['dependencies'], $license_modal_meta['version'], true );
+		wp_enqueue_style( 'admin-kadence-license-modal', KADENCE_BLOCKS_URL . 'dist/admin-license-modal.css', [ 'wp-components' ], $license_modal_meta['version'] );
+		wp_localize_script(
+			'admin-kadence-license-modal',
+			'kadenceLicenseModalParams',
+			[
+				'ajaxurl'         => admin_url( 'admin-ajax.php' ),
+				'wpnonce'         => wp_create_nonce( 'kadence-blocks-manage' ),
+				'licensePageUrl'  => function_exists( 'lw_harbor_get_license_page_url' ) ? esc_url( lw_harbor_get_license_page_url() ) : '',
```
jonwaldstein: We're doing alot of function_exists call throughout this PR, honestly the function should always exist because we're pulling it into this version via composer.  If we want to be extra careful then we should create a simpler decorator so we don't have to keep doing these checks everywhere 😄

jonwaldstein (commented):
@pramodjodhani I like the updated UI but I think we need to consider the future of harbor, how other plugins will do something similar to this in the future, and the effect on existing Kadence add-ons who all come with their own license key fields. I added some notes related to all this throughout the codebase.  I also recommend installing all the Kadence add-ons defined in `Get_Known_Plugins` while you're working on this to see the full picture 😄

src/license-modal/active-license-view.js
pramodjodhani: Good point. I have created a common hook and a helper function. [`15df124` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1154/commits/15df12450dc12066e5a13c7ac826f0a2807880ac)

includes/resources/Harbor/Licensing/Abstract_License_Strategy.php
pramodjodhani: Right. Added unit tests: [`d46cb8b` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1154/commits/d46cb8b7d55f8af031ea5f30fbe22c031004272e)

includes/settings/class-kadence-blocks-settings.php
pramodjodhani: That's a good one. This is one of the last changes. Removed unused code [`7a014a6` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1154/commits/7a014a6ee70bd482687dc6ab3eb04731365cef5c)

src/assets/js/admin-scripts.js
pramodjodhani: Removed unused code [`7a014a6` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1154/commits/7a014a6ee70bd482687dc6ab3eb04731365cef5c)

includes/resources/Harbor/Licensing/Unified_License_Strategy.php
pramodjodhani: This is fixed here: [`73899ea` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1154/commits/73899ea80572639e0421cddd067a8e052a405ce8)

includes/resources/Harbor/Harbor_Provider.php
pramodjodhani: When we only have the Kadence block installed, and after the user enters the license key, we provide them a button to automatically install or activate the Kadence Blocks Pro. However, the problem in this scenario, when no other premium plugin exists, is that the Harbor REST API is not loaded. That's why we are not able to use the Harbor API to automatically install or activate the Kadence Blocks Pro. 

This is a hacky solution to load Harbor code for specific rest API endpoints.

includes/resources/Harbor/Licensing/Unified_License_Strategy.php
pramodjodhani: Good idea, yes, I have made that change: [`6cc670e` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1154/commits/6cc670ef8799ff3035d4106a80cfada99aac75f9)

includes/settings/class-kadence-blocks-settings.php
pramodjodhani: I have removed the unnecessary function_checks [`6cc670e` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1154/commits/6cc670ef8799ff3035d4106a80cfada99aac75f9)

includes/resources/Harbor/Harbor_Provider.php
pramodjodhani: I'm afraid this is not available publicly. It's a protected property under the [License_Controller class](https://github.com/stellarwp/harbor/blob/567d5371c5693248e9a1376202d338acb5de576f/src/Harbor/API/REST/V1/License_Controller.php#L35). I'll take a note of this and fix it in the next PR.

includes/helper-functions.php
pramodjodhani: Its correct, the lw_harbor_is_product_license_active exists within the get_unified_key()

https://github.com/stellarwp/kadence-blocks/pull/1154/changes#diff-a847b82978481ad6fd5be49980348d3c62342b2978c0877597ce4fe2c8f7e4baR50

includes/resources/Harbor/Harbor_Provider.php
pramodjodhani: Good catch. I have restored this code, but skipped it for `kadebce-blocks` [`1fc9f67` (this PR)](https://github.com/stellarwp/kadence-blocks/pull/1154/commits/1fc9f67a3d44b088f55e8a3984c069dfcceee082)

pramodjodhani:
@jonwaldstein @redscar all the changes have been done. Can you please give it another look?

redscar (approved):
I'm happy from my side. I would see if Jon can give his approval - and then you should be good to go.

includes/resources/Harbor/Harbor_Provider.php
jonwaldstein: @pramodjodhani that's good context, I think we should take a moment to think through this because i'm concerned about releasing more _hacky_ solutions - especially in Kadence 😏 

I'd love to get @d4mation's thoughts here - perhaps we have a quick chat about this today

jonwaldstein:
@pramodjodhani I requested @d4mation take a look at this with me because of the way we're interacting with harbor is making me nervous 😅

includes/resources/Harbor/Licensing/Unified_License_Strategy.php
jonwaldstein: actually i'm not sure about pulling in that class because harbor changes the [leader](https://github.com/stellarwp/harbor/blob/main/docs/architecture/fat-leader-thin-instance.md) meaning this logic would change depending on versions/leader presence which would cause the strauss class reference to become stale.

What i'm really getting at here is stopping for a moment and asking some questions:
- is hardcoding this a reliable solution between harbor versions?
- should harbor expose this information to plugins?

jonwaldstein (changes requested):
I set up a meeting with @pramodjodhani and @d4mation to chat through the engineering direction - we can circle back to this PR afterwards

pramodjodhani:
Closing this PR, as have ported over this code to https://github.com/stellarwp/kadence-blocks-pro/pull/299

PR #1164, opened 2026-07-23
[DTM] SOFT-3902 [2/2]: JS token-picker accessor

pauloiankoski:
## Summary

Adds the JS accessor half of the pickable-token surface: a new `src/extension/token-picker/` module. Pure functions that turn the `window.kadenceDesignTokensPickable` pool (from #1163) into a per-control pickable list — a hard type filter by control kind, semantic-layer tokens ranked before primitives, each entry carrying a resolved preview `value`. No UI and no attribute writes (those belong to the picker ticket).

Stacked on #1163 — review/merge that first.

## Changes

- `pickableTokenPool()` — reads the pool global, fail-soft on a missing/partial payload.
- `pickableTokensFor(kind, set)` — kind-to-type hard filter (a radius control never lists color or font tokens), semantic-before-primitive stable rank, and each entry's `value` resolved from the requested set with an active-set fallback. Returns `[{ id, alias, label, value, type }]`.
- `pickableTokensForControl(blockName, controlAttr, set)` — resolves the control kind from the variant-catalog surface and delegates; the entry point the picker UI will call.
- jest tests: per-kind filtering with no cross-type leakage, semantic-first ordering, fail-closed on an unknown kind, set override + active-set fallback, and fail-soft on a missing pool.

`value` is preview-only; a pick will write the `alias`. Precise per-control narrowing within `dimension` (radius vs spacing) and role-pinning are a documented follow-up — this surface ranks semantic before primitive.

## Test plan

- `bun run test src/extension/token-picker` green (12 tests).
- `bun run lint-js src/extension/token-picker` clean.

### DIFF
--- a/src/extension/token-picker/index.js
+++ b/src/extension/token-picker/index.js
@@ -21,6 +21,8 @@ import { activeSet, blockProperties } from '../variant-picker';
  * kinds; values are DTCG $type lists. `shadow` is inert until a shadow control kind exists — it is
  * mapped now so a future shadow control lights up with no change here. An unknown kind yields no
  * types, so the filter fails closed.
+ *
+ * @since TBD
  */
 const KIND_TYPES = {
 	color: ['color'],
--- a/src/extension/token-picker/index.js
+++ b/src/extension/token-picker/index.js
@@ -1,0 +1,118 @@
+/**
+ * Pickable-token accessor for the editor token picker.
+ *
+ * The pool is printed by the server-side editor localizer to `window.kadenceDesignTokensPickable`:
+ * `{ tokens: [{ id, alias, label, type, layer }], values: { <setSlug>: { <id>: literal } } }`.
+ * This module turns it into the per-control PICKABLE list — hard type filter by the control's kind
+ * (a radius control never lists color or font tokens), semantic-layer tokens ranked before
+ * primitives, and each entry carrying the resolved literal `value` for the preview swatch/number.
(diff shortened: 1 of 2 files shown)


src/extension/token-picker/index.js
```diff
@@ -0,0 +1,118 @@
...
+import { get } from 'lodash';
+import { activeSet, blockProperties } from '../variant-picker';
+
+/**
+ * Token $types compatible with each control kind. Keys are the variant catalog's coarse control
+ * kinds; values are DTCG $type lists. `shadow` is inert until a shadow control kind exists — it is
+ * mapped now so a future shadow control lights up with no change here. An unknown kind yields no
+ * types, so the filter fails closed.
+ */
+const KIND_TYPES = {
```
d4mation, line 27: It would be worth adding `@since TBD` here too.

d4mation (approved):
> value is preview-only; a pick will write the alias. Precise per-control narrowing within dimension (radius vs spacing) and role-pinning are a documented follow-up — this surface ranks semantic before primitive.

Make sure to create a follow-up ticket for this, please :)

src/extension/token-picker/index.js
pauloiankoski: https://github.com/stellarwp/kadence-blocks/pull/1164/commits/c1061870c4c48bf6b8dbd9a63275c3103afb2c36

d4mation (approved):

PR #1165, opened 2026-07-23
[DTM] SOFT-3403: Advanced Text (advancedheading) token bindings

pauloiankoski:
## Summary

Wires the Advanced Text block (`kadence/advancedheading`) to the design-tokens system, so a fresh block follows the site's tokens for its core design properties while any per-instance value the user sets still wins.

Every bound attribute is empty-by-default in the block, so all of them flow through the existing low-specificity `Block_Default_Css` projector — the same mechanism already used for `kadence/image` and `kadence/single-icon` color. The per-block mapping spike found **no deviation**, so there is **no adapter, no `build_css()` change, no SCSS change, and no editor-JS change**. (The button block delivers the same tokens through SCSS `var()` fallbacks because its stylesheet loads on the front end; this block's does not — all its front-end CSS comes from PHP — so it uses the PHP projector to get the identical low-specificity-default semantics.)

## What changed

- **`baseline.json`** — 6 new `semantic.*` leaves: `color.heading-bg` (transparent), `font-size.heading` (`2rem`), `line-height.heading` (`1.2`), `font-weight.heading` (`700`), `spacing.heading-padding` (0), `radius.heading` (0). Plus a `kadence/advancedheading` `$default` variant entry mapping the 12 bound properties to their tokens.
- **`declarations.php`** — a `kadence/advancedheading` `variant_sets` entry with 12 `token` + `css_prop` bindings (no `css_selector`, no picker `label`).
- Reuses existing leaves for color, font-family, letter-spacing, text-transform, border color, and border width — no duplication.
- **No `tokens` registration** for any of these (the `:root` vars emit from the baseline document regardless, and overrides work without it — matching the `single-icon` color and button-typography precedents). Admin-UI editability + `Baseline_Guard` coverage is a deliberate follow-up.

## Scope

Bound: text color, background color, font-family, font-size, line-height, font-weight, letter-spacing, text-transform, padding, and border color/width/radius (12 properties).

Out of scope (deliberate follow-up): margin (would strip theme heading spacing) and the compound sub-features — highlight/mark, link colors, icon color/size/padding, inline image, and text-shadow.

## Behavior on activation (intentional re-skin)

Only `color`, `font-family`, `font-size`, `line-height`, and `font-weight` are visibly affected — an unset Advanced Text block adopts the design-system defaults (neutral.900 / Inter / 2rem / 1.2 / 700), the same trade-off accepted for the button font. Background, padding, border, letter-spacing, and text-transform resolve to inert values, so they change nothing. A per-instance value still wins by specificity.

## ⚑ For reviewer confirmation

`font-size.heading` (`2rem`), `line-height.heading` (`1.2`), and `font-weight.heading` (`700`) are the only properties that force **every** Advanced Text block — including body-text uses of the block — to one value, overriding a theme's per-tag (h1/h2/p) styling. Please confirm these exact numbers, or whether `font-size` in particular should be inert instead of a fixed default.

## Test plan

- New `Css_BuilderTest` case asserts all 12 declarations emit in the grouped `.wp-block-kadence-advancedheading{...}` rule with the right var names and resolved fallbacks (including the font-family stack `Inter, system-ui, sans-serif`).
- New `Variant_ResolverTest` case asserts the variant set is registered and `resolve_default()` returns all 12 property/value pairs.
- Repointed `VariantsControllerTest`'s "block that accepts no variants" 404 case off `advancedheading` (now registered) onto `kadence/spacer`.
- Full `Resources/Design_Tokens` wpunit suite: 950 tests, 0 failures. `phpstan` clean (`phpstan-baseline.neon` untouched); `cspell` clean.

---

This PR targets `SOFT-3403/base`; the roll-up into `feature/design-tokens-module` is a separate integration-base PR.

### DIFF
--- a/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
+++ b/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
@@ -397,6 +397,10 @@
 			"image-bg": {
 				"$type": "color",
 				"$value": "{primitive.color.transparent}"
+			},
+			"heading-bg": {
+				"$type": "color",
+				"$value": "{primitive.color.transparent}"
 			}
 		},
 		"spacing": {
@@ -415,6 +419,10 @@
 			"media-padding": {
 				"$type": "dimension",
 				"$value": "{primitive.dimension.spacing.none}"
+			},
+			"heading-padding": {
+				"$type": "dimension",
+				"$value": "{primitive.dimension.spacing.none}"
 			}
 		},
 		"radius": {
@@ -431,6 +439,10 @@
 				"$value": "{primitive.dimension.radius.none}"
 			},
 			"column": {
+				"$type": "dimension",
+				"$value": "{primitive.dimension.radius.none}"
+			},
+			"heading": {
 				"$type": "dimension",
 				"$value": "{primitive.dimension.radius.none}"
 			}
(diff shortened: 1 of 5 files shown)


pauloiankoski:
Superseded by #1166 — retargeted directly onto `feature/design-tokens-module` (single-phase ticket, no base branch needed).

PR #1167, opened 2026-07-23
[DTM] SOFT-3403 [2/2]: Block_Default_Css editor parity

pauloiankoski:
🎬 https://www.loom.com/share/299881253c7e40ac93aa538cbf10e9d3

## Summary

Gives the `Block_Default_Css` projector a separate **editor** output so a block's token defaults render correctly in the block editor, matching the front end. Front-end output is unchanged.

## The problem

The projector emitted **one** CSS string used for both the front-end and editor style handles. For `kadence/advancedheading` that breaks in the editor: `useBlockProps()` puts `wp-block-kadence-advancedheading` on a **wrapper `<div>`**, while the actual heading is a descendant `<h2 class="… kadence-advancedheading-text">`. So the rule `.wp-block-kadence-advancedheading{…}` landed on the wrapper div in the editor — non-inherited props (background, padding, border) styled the wrong element, and the theme's `.editor-styles-wrapper <tag>` rules overrode the inherited typography on the real heading. On the front end the same class is on the heading, so it was always correct there.

## The fix

- **`Css_Builder`** gains an editor build. For a block that declares an `editor_selector`, the editor rule is scoped under `.editor-styles-wrapper` and retargeted at that selector; blocks without one emit editor CSS **byte-for-byte identical** to their front-end CSS (no change for `image`/`single-icon`/`rowlayout`/`column`). The front-end `css()` path is untouched.
- **`Variant_Set`** carries an optional `editor_selector`.
- **`declarations.php`** declares `editor_selector` for `kadence/advancedheading` → `.wp-block-kadence-advancedheading .kadence-advancedheading-text`, so the editor rule is `.editor-styles-wrapper .wp-block-kadence-advancedheading .kadence-advancedheading-text{…}`.
- **`Projector::enqueue_editor()`** now emits the editor build; `enqueue_front_end()` is unchanged.
- **Cache keys are context-separated** (`front:` / `editor:` memo prefixes and distinct object-cache keys) so the two builds never collide.

Specificity holds: the editor rule (three classes) beats the theme's per-tag rule, and the block's own per-instance editor styles are applied **inline** (`style=…` on the heading/RichText), so a value set on the block always wins.

## Verified in the editor

In the live block editor: the token typography (font-weight 400, text-transform, font-size) now applies to the real `<h2>`; the old front-end-selector rule is no longer present in the editor; and a per-instance color set on the block still wins over the token.

## Live re-injection parity (also in this PR)

`Projection/Editor_Css.php` aggregates the projectors for the `/projected-css` REST endpoint, which the editor uses to **live re-inject** CSS when a token/variant changes without a reload (`src/extension/design-tokens/live-css.js`). To keep that path in parity too, this PR adds `editor_css()` to the `Css_Projector` interface — `Css_Var`/`Variant` return `css()` (their selectors are context-independent), `Block_Default_Css` returns its editor-scoped build — and `Editor_Css` now aggregates `editor_css()`. So **both** editor page-load and live re-injection use the correctly-scoped editor CSS. No follow-up.

## Test plan

- `Css_BuilderTest`: asserts the editor CSS for advancedheading uses the `.editor-styles-wrapper … .kadence-advancedheading-text` selector with all 13 declarations; asserts the front-end CSS is unchanged; asserts a block without an `editor_selector` emits editor CSS byte-for-byte equal to its front-end CSS.
- `Variant_SetTest`: parses `editor_selector` / null when omitted.
- `Resources/Design_Tokens/Projection/Block_Default_Css` 13/13, `Resources/Design_Tokens/Registry` 137/137; `phpstan` clean (baseline untouched); `cspell` clean.

---

Stacked on #1166 (`[1/2]`). Base branch of this PR is `SOFT-3403/phase-1-advancedheading-token-bindings`.

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
+++ b/includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
@@ -128,6 +128,39 @@ final class Css_Builder {
 	}
 
 	/**
+	 * Cached variant of css(): memoized per request and persisted in the object cache keyed on the store
+	 * version (and plugin version), so a token write (which bumps the store version) and a plugin upgrade
+	 * both invalidate it automatically.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $version The store version the resolved set was built from.
+	 * @param string $slug    The token set slug.
+	 *
+	 * @return string
+	 */
+	public function css_for_version( string $version, string $slug ): string {
+		return $this->for_version( $version, $slug, false );
+	}
+
+	/**
+	 * Cached variant of editor_css(): same memo/object-cache mechanics as {@see self::css_for_version()}, but
+	 * keyed under a distinct `editor:` context so the editor-scoped string (which differs from the front-end
+	 * one for any block declaring an `editor_selector`) never collides with, or gets served in place of, the
+	 * front-end cache entry.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $version The store version the resolved set was built from.
+	 * @param string $slug    The token set slug.
+	 *
+	 * @return string
+	 */
+	public function editor_css_for_version( string $version, string $slug ): string {
+		return $this->for_version( $version, $slug, true );
+	}
+
+	/**
 	 * Shared build for both {@see self::css()} and {@see self::editor_css()}; only the selector base differs
 	 * per block, per the presence of an `editor_selector` declaration.
 	 *
@@ -192,39 +225,6 @@ final class Css_Builder {
 		}
 
 		return $css;
-	}
-
-	/**
-	 * Cached variant of css(): memoized per request and persisted in the object cache keyed on the store
-	 * version (and plugin version), so a token write (which bumps the store version) and a plugin upgrade
-	 * both invalidate it automatically.
-	 *
-	 * @since TBD
-	 *
-	 * @param string $version The store version the resolved set was built from.
-	 * @param string $slug    The token set slug.
-	 *
-	 * @return string
-	 */
-	public function css_for_version( string $version, string $slug ): string {
-		return $this->for_version( $version, $slug, false );
-	}
-
-	/**
-	 * Cached variant of editor_css(): same memo/object-cache mechanics as {@see self::css_for_version()}, but
-	 * keyed under a distinct `editor:` context so the editor-scoped string (which differs from the front-end
-	 * one for any block declaring an `editor_selector`) never collides with, or gets served in place of, the
-	 * front-end cache entry.
-	 *
-	 * @since TBD
-	 *
-	 * @param string $version The store version the resolved set was built from.
-	 * @param string $slug    The token set slug.
-	 *
-	 * @return string
-	 */
-	public function editor_css_for_version( string $version, string $slug ): string {
-		return $this->for_version( $version, $slug, true );
 	}
 
 	/**
--- a/includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
+++ b/includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
@@ -246,13 +246,14 @@ final class Css_Builder {
 	 */
 	private function for_version( string $version, string $slug, bool $editor ): string {
 		$context  = $editor ? 'editor' : 'front';
-		$memo_key = $context . ':' . $version . ':' . $slug;
+		$suffix   = $version . ':' . $slug;
+		$memo_key = $context . ':' . $suffix;
 
 		if ( isset( $this->memo[ $memo_key ] ) ) {
 			return $this->memo[ $memo_key ];
 		}
 
-		$cache_key = 'block_default_css_' . $context . '_' . KADENCE_BLOCKS_VERSION . '_' . $version . ':' . $slug;
+		$cache_key = 'block_default_css_' . $context . '_' . KADENCE_BLOCKS_VERSION . '_' . $suffix;
 		$cached    = wp_cache_get( $cache_key, self::CACHE_GROUP, false, $found );
 
 		if ( $found && is_string( $cached ) ) {
--- a/includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
+++ b/includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
@@ -35,8 +35,11 @@ use RuntimeException;
  *   - attribute set to any value → the block's higher-specificity rule wins, including an explicit `0`.
  *
  * And because the Projector enqueues this onto KB's editor style handle as well as the front-end one,
- * the editor canvas gets the same default for free — the reason this CSS approach is preferred over an
- * attribute/slug/control approach for these families. Nothing here is `!important`.
+ * the editor canvas gets the same token default — for most blocks the identical rule, and for a block
+ * whose editor markup renders the binding on a different element (one declaring an `editor_selector`,
+ * currently Advanced Heading) a rule re-scoped to that element so the default still lands. This
+ * editor-canvas parity is the reason this CSS approach is preferred over an attribute/slug/control
+ * approach for these families. Nothing here is `!important`.
  *
  * Only a binding that declares a `css_prop` and references a token contributes, and only the block's
  * `$default` variant is read (named variants are the selectable-variant projector's job). Pure: no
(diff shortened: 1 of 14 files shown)


includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
```diff
@@ -107,6 +107,38 @@ public function __construct( Token_Registry $registry, Variant_Resolver $variant
...
+	 * per block, per the presence of an `editor_selector` declaration.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $slug   The token set whose resolved values the `$default` aliases resolve against.
+	 * @param bool   $editor Whether to build the editor-scoped variant.
+	 *
+	 * @return string The CSS, or an empty string when there is nothing to project.
+	 */
+	private function build( string $slug, bool $editor ): string {
```
d4mation, line 178: This new `private` method sits above other `public` methods which violates the order defined in `AGENTS.md`.

includes/resources/Design_Tokens/Projection/Css_Var/Projector.php
```diff
@@ -204,6 +204,19 @@ public function css(): string {
...
+	 * custom properties, which carry no block-scoped selector, so nothing about the editor's markup shape
+	 * (e.g. a wrapper div) requires a different build.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public function editor_css(): string {
+		return $this->css();
+	}
```
d4mation: Maybe we should have an abstract class which has this? That way we only override it when necessary, like in `Block_Default_Css\Projector`.

includes/resources/Design_Tokens/Registry/declarations.php
```diff
@@ -421,8 +421,17 @@ static function ( string $slug ): array {
...
+			//
+			// In the editor, useBlockProps() puts `.wp-block-kadence-advancedheading` on a wrapper <div>, not
+			// on the heading element the bindings above are meant to style — the real heading is a descendant
+			// carrying the stable `kadence-advancedheading-text` class. So the editor build of this CSS (see
+			// Css_Builder::editor_css()) targets that descendant instead, scoped under `.editor-styles-wrapper`
+			// so it still outranks the theme's element styles there too. Per-instance color/font-size render as
+			// INLINE styles on that same element (and font-weight inline on its child), so they keep winning
+			// regardless of this rule's specificity.
+			'block'           => 'kadence/advancedheading',
+			'editor_selector' => '.wp-block-kadence-advancedheading .kadence-advancedheading-text',
```
d4mation, line 433: Man, I hate this, but I get it 😢

tests/wpunit/Resources/Design_Tokens/Projection/Block_Default_Css/Css_BuilderTest.php
```diff
@@ -202,6 +202,64 @@ public function testTheShippedDeclarationsEmitTheAdvancedHeadingSurfaceRule(): v
...
+	/**
+	 * A block whose Variant_Set declares no `editor_selector` (e.g. Image) must emit an editor build
+	 * byte-for-byte identical to its front-end build — no `.editor-styles-wrapper` prefix and no
+	 * re-targeting — so blocks without the wrapper-div problem see zero regression from this mechanism. Uses
+	 * an isolated registry (rather than the shipped one) so the assertion is not diluted by the shipped
+	 * Advanced Heading declaration, which DOES declare an `editor_selector` and legitimately diverges.
+	 *
+	 * @return void
+	 */
+	public function testABlockWithNoEditorSelectorEmitsIdenticalFrontEndAndEditorCss(): void {
```
d4mation, line 255: Thank you for this test in particular!

tests/wpunit/Resources/Design_Tokens/Projection/Css_Var/ProjectorTest.php
```diff
@@ -144,6 +144,19 @@ public function testEnqueueEditorAppendsInlineStyleToEditorHandle(): void {
...
+	// ---- Editor CSS -----------------------------------------------------------------------------------
+
+	/**
+	 * The token-var projector is context-independent — it emits `:root { --kb-token--*: … }` custom
+	 * properties with no block-scoped selector — so its editor build is byte-for-byte identical to its
+	 * front-end build.
+	 *
+	 * @return void
+	 */
+	public function testEditorCssEqualsCss(): void {
```
d4mation, line 156: This is also a super helpful test!

includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
d4mation, line 1: We have a stale docblock for `$memo` that should be updated to match this change.

The classdoc also states the editor canvas gets the same default as the front-end for free, which isn't entirely true anymore.

includes/resources/Design_Tokens/Projection/Contracts/Css_Projector.php
d4mation, line 1: The classdoc calls out how the Editor CSS projector works but with this change it isn't fully accurate anymore. It should be updated.

includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
```diff
@@ -173,20 +207,55 @@ public function css( string $slug = 'default' ): string {
...
+	 *
+	 * @param string $version The store version the resolved set was built from.
+	 * @param string $slug    The token set slug.
+	 * @param bool   $editor  Whether to build the editor-scoped variant.
+	 *
+	 * @return string
+	 */
+	private function for_version( string $version, string $slug, bool $editor ): string {
+		$context  = $editor ? 'editor' : 'front';
+		$memo_key = $context . ':' . $version . ':' . $slug;
```
d4mation: We end up rebuilding very similar strings for `$memo_key` and `$cache_key`. 

Since this is still unreleased, we should probably clean this up. How about this?

```php
$suffix    = $version . ':' . $slug;
$memo_key  = $context . ':' . $suffix;
$cache_key = 'block_default_css_' . $context . '_' . KADENCE_BLOCKS_VERSION . '_' . $suffix;
```

includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
pauloiankoski: Moved `build()` down into the private section so it sits below the public methods. Fixed in a6ec1b8.

includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
pauloiankoski: Updated the `$memo` docblock (the key now folds in the front/editor context) and reworded the classdoc so it no longer claims the editor gets the identical default for free: a block declaring an `editor_selector` gets a re-scoped rule instead. e57b56e.

includes/resources/Design_Tokens/Projection/Block_Default_Css/Css_Builder.php
pauloiankoski: Extracted the shared `version:slug` suffix exactly as you suggested. 120f03b.

includes/resources/Design_Tokens/Projection/Contracts/Css_Projector.php
pauloiankoski: Updated the classdoc: it now describes the `css()` / `editor_css()` split, and corrects that `Editor_Css` aggregates `editor_css()` (not `css()`). 7a7e411.

includes/resources/Design_Tokens/Projection/Css_Var/Projector.php
pauloiankoski: Added `Abstract_Css_Projector` with the default `editor_css()` returning `css()`. Css_Var and Variant (which both had the same trivial override) now inherit it; Block_Default_Css keeps its own override. Placed under `Projection\Contracts` to match the existing `Adapter\Contracts\Abstract_Adapter`. 7006b2f.

includes/resources/Design_Tokens/Registry/declarations.php
pauloiankoski: Yeah, I was upset that some of the props weren’t being applied to the Text block in the editor. The reason is that it uses a different wrapper and has a different structure than the block on the frontend. This is a "optional" solution in case we wanna make them match 100%.

d4mation (approved):

PR #1173, opened 2026-07-24
[DTM] SOFT-3906/SOFT-3995: Blocks-side token seam registration + singlebtn proving ground

pauloiankoski:
🎥 https://www.loom.com/share/1a90e2fd351f4b289486507e5213e655

## Summary

Lands the Kadence Blocks side of the design-token picker: a **field-as-trigger** UI injected into the token-agnostic `@kadence/components` control seams (SOFT-3995), with the singlebtn Border Radius as the proving ground (SOFT-3906).

Each token-mapped control slot renders as an input-styled trigger (with its corner icon) that opens a two-tab popover — **Style Library** (pick a size token, or Reset to the inherited default) and **Custom** (a number input, unit switch, and slider on one row). All token knowledge lives in Kadence Blocks; the `@kadence/components` package stays token-agnostic and only exposes the neutral `kadence.components.control.editor` / `.actions` seams.

## What's here

- `src/extension/design-tokens/component-token-ui.js` — the token UI. `TokenFieldControl` is the field-as-trigger (corner icon + trigger + Style Library/Custom popover, with tab icons and labels); `isTokenAlias`/`findTokenEntry` are the shared helpers; `TokenChip`/`TokenPickerButton` remain for the whole-value box-shadow control, which has no per-slot field.
- `src/extension/design-tokens/component-token-ui.scss` — trigger + popover styling to the sidebar design (13px type, `#949494` borders / muted values, 2px field radius, 4px 260px popover). Wraps the four corners into a 2×2 grid, hides the control's own unit select for token-mapped controls, and gives the measure control's linked toggle the Border control's pressed fill — all scoped via `:has()`.
- `src/extension/design-tokens/register-component-filters.js` — registers the seam listeners. The `editor` seam replaces each token-mapped slot with a `TokenFieldControl`; a per-slot pick writes through the adapter's `writeLeaf`, so a linked slot writes every side and an individual slot writes only its own (per-corner tokens fall out for free). The `actions` seam serves only the box-shadow chip-or-picker.
- `src/extension/token-picker/index.js` — `pickableTokensForControl` resolves the control's role (from the bound token, else inferred from the control attribute) and **prefers the primitive size scale**, hiding the component-specific semantic radii.
- `src/extension/token-picker/parse-css-length.js` — resolves a token length (e.g. `"0.5rem"`) to its numeric size, seeding the Custom tab when leaving a token.
- `includes/resources/Design_Tokens/Registry/declarations.php` — registers the primitive radius size scale (`None` / `Small` / `Medium` / `Large` / `Round`) as pickable dimension tokens. The values already ship in the baseline; the semantic radii keep aliasing them, so nothing changes until a step is overridden.
- `includes/class-kadence-blocks-editor-assets.php` — enqueues `dist/early-filters.css`, so the injected control styles actually load in the editor.
- `src/early-filters.js` — bootstraps `registerComponentTokenFilters()` at editor load, so the picker lights up for any control that passes a `context`.
- `src/blocks/singlebtn/edit.js` — the Border Radius proving ground. Its linked/individual mode is **derived from the corner values** (all four equal reads as linked, otherwise individual), so no new attribute is stored and existing buttons open in the right mode. Switching to linked collapses the corners to the top-left value (so the button reflects the link immediately) via `onChangeControl`, with an ephemeral session-only override so unlinking already-equal corners still works. It passes a neutral `context` (`blockName`, `attribute`, `unit`/`units`/`onUnit`, `min`/`max`/`step`, and `defaultValue` from the preset binding) so the field shows sizes, a unit switch, and the inherited default's size + value rather than a bare "Default".
- `src/blocks/singlebtn/components/backend-styles/index.js` — the button preview now emits each border-radius corner through `css.render_size()`, so a bound token resolves to its `var(--kb-token--…)` reference (a complete length) while a numeric value still gets its unit. Without this, picking a token wrote a `{dot.alias}` that concatenated with the unit into invalid CSS, so the preview only reflected custom numbers.
- `includes/class-kadence-blocks-css.php` — the front-end twin of the above. `render_measure_output()` (the PHP renderer singlebtn's Border Radius uses) now resolves a design-token alias corner to its `var(--kb-token--…)` reference, so a picked token renders on the published page instead of falling back to the default. Its 12 duplicated per-corner branches are collapsed into one `render_measure_side()` helper that carries the numeric, `position`-keyword, design-token-alias, and legacy-spacing-variable cases; the three breakpoints are unchanged.

## Behavior

- The popover anchors to the **left** of the field; its two tabs carry icons (globe / sliders) beside their labels. It opens on the **Custom** tab when the slot already holds a literal value, otherwise on **Style Library**.
- **Style Library** lists the radius sizes; `Reset` clears the slot's override, enabled only when one is set.
- On its inherited default the field reads the effective size (e.g. `Medium 0.5rem`), and that row is tagged `Default` and marked active.
- **Custom** edits a literal value + unit with a slider on a single row; switching to a custom value drops the alias and leaves a usable number.
- The linked/individual mode is **derived from the corners** — all four equal reads as linked, any difference reads as individual — so it persists across reselecting the block with no extra attribute. Switching to linked collapses the four corners to the top-left value (the button reflects it immediately); switching to individual keeps the current values for the session.
- The linked/individual toggle matches the Border control's pressed (theme-color) fill.
- A picked token — linked (all corners) or an individual corner — is reflected live on the button preview, and renders identically on the published front end.

## Notes

- **Border-radius corner normalization is expected CSS, not a bug.** When a corner's radius is larger than the space on its side of the button (for example a large radius on a short button), CSS scales *all four* corners down by one shared factor so the curves can't overlap: `f = min(sideLength / sum-of-the-two-radii-on-that-side)`, applied to every corner when `f < 1`. `getComputedStyle` still reports the specified value while the painted (used) value is the scaled one, so a corner can look smaller than its stored value. This is identical in the editor and on the front end and needs no code change. See the spec: [CSS Backgrounds and Borders — Corner Overlap](https://www.w3.org/TR/css-backgrounds-3/#corner-overlap).

## Stacked on

- #1172 — the `@kadence/components` pin. Review/merge that first.

## Test plan

- [x] `bun run test src/extension/design-tokens src/extension/token-picker` green (seam dispatch, per-corner/linked pick + clear + custom, role inference + primitive-size preference, `parse-css-length`)
- [x] `bun run lint-js` clean on the touched files
- [x] PHPStan + PHPCS clean on the touched PHP
- [x] cspell clean
- [x] Verified in the editor against the Figma: field-as-trigger with corner icons, 2×2 corners, left popover, tab icons + labels, Style Library sizes, single-row Custom tab, inherited-default display, matching linked toggle, and live preview for both linked and per-corner tokens
- [x] Verified linked/individual in the editor: individual corners persist across reselecting the block, the popover opens on the Custom tab for a literal, and switching to linked collapses the corners so the button updates immediately
- [x] `render_measure_output` token resolution covered by wpunit (`KadenceBlocksCssTokenEmissionTest`): alias → `var()`, alias mixed with numeric siblings rotated through all four corners, and fail-open for malformed braces; `KadenceBlocksCssBcSnapshotTest` stays byte-identical after the per-corner refactor
- [x] Verified editor == front end by saving a mixed config (`None` / `Large` / custom `20px` / `Round`) and comparing computed radii on the published page — all four corners match


src/extension/design-tokens/component-token-ui.js
```diff
@@ -0,0 +1,127 @@
...
+					{tokens.map((entry) => (
+						<MenuItem
+							key={entry.id}
+							onClick={() => {
+								onClose();
+								onSelect(entry.alias);
+							}}
+						>
+							{entry.label}
+							<span className="kadence-token-picker__preview">{entry.value}</span>
```
d4mation, line 411: These end up rendering with no space between them. Maybe they should have `justify-content: space-between;` applied?

image

src/extension/design-tokens/register-component-filters.js
```diff
@@ -0,0 +1,225 @@
...
+	const onUnlink = write
+		? () => {
+				const entry = findTokenEntry(tokens, leaf);
+				const parsed = entry ? parseCssLength(entry.value) : null;
+				write(adapter.writeLeaf(ctx.value, ctx.index, parsed ? parsed.size : ''));
+			}
+		: undefined;
+
+	return <TokenChip value={leaf} tokens={tokens} onUnlink={onUnlink} />;
+}
```
d4mation, line 142: When you start to only unlink certain corners, this ends up looking really weird. Is there a way that we could instead make it look like the original inputs with the Token Name/Label but disabled?

image

src/extension/design-tokens/register-component-filters.js
```diff
@@ -0,0 +1,225 @@
...
+	border: {
+		writeAll: (value, alias) => {
+			const next = { ...(value || {}) };
+			['top', 'right', 'bottom', 'left'].forEach((side) => {
+				const arr = Array.isArray(next[side]) ? [...next[side]] : ['', '', ''];
+				arr[2] = alias;
+				next[side] = arr;
+			});
+			return next;
+		},
```
d4mation: Having a way to write all Corners I do think is a good idea, but maybe we should also allow setting a per-Corner Token? Right now the only way you can do that is by unlinking specific Corners. But then the unlinked corners still can only have a Static value rather than a Token value.

If a picker could surface for each Corner that would be very nice.

Another thing to consider: It overwrites all of the Corners without confirmation. If the user had per-Corner Border Radii configured, they could lose those configurations without realizing it.

src/extension/design-tokens/register-component-filters.js
```diff
@@ -0,0 +1,225 @@
...
+		return defaultEditor;
+	}
+
+	const tokens = tokensFor(ctx);
+	const write = writerFor(ctx);
+	const onUnlink = write
+		? () => {
+				const entry = findTokenEntry(tokens, leaf);
+				const parsed = entry ? parseCssLength(entry.value) : null;
+				write(adapter.writeLeaf(ctx.value, ctx.index, parsed ? parsed.size : ''));
```
d4mation: When unlinking, we end up throwing away the unit here. I figure that is intentional because of how the unit selector works, but we probably should try to calculate the equivalent based on the unit selector. Because otherwise if your Token was `2rem` but the unit selector was set to `px`, on unlink you end up having `2px` effectively saved. 

image

d4mation (commented):
Could you record a Loom for this as well, please? Tickets interacting with UI especially are very helpful to have a Loom for.

You could also share it in the channel as a Demo and ask for some feedback from Donaldo about some of the things I surfaced here. Maybe he'll have some different opinions on what we could maybe do to handle these issues (or maybe that they're not important for the initial release).

src/extension/design-tokens/component-token-ui.js
pauloiankoski: Fixed in the redesign. The rows and the Custom row now use flex with `space-between` and a gap. Let me know if one spot still looks too tight.

src/extension/design-tokens/register-component-filters.js
pauloiankoski: Done. Each corner is now an input-style field that shows its token name and value, so it looks like the old inputs.

src/extension/design-tokens/register-component-filters.js
pauloiankoski: Per-corner tokens work now. Each corner has its own picker and can hold a token or a custom value.

The overwrite point is still true: clicking Link sets all four corners to the top-left value with no confirm. I kept it this way because the mode is derived (no new attribute). Maybe Donaldo can say if we need a confirm or undo.

src/extension/design-tokens/register-component-filters.js
pauloiankoski: Fixed for tokens. A token corner keeps its alias, so `2rem` stays `2rem` no matter what the unit selector says. Custom numbers still share one unit, but that is old behavior and a separate thing.

src/blocks/singlebtn/edit.js
```diff
@@ -1191,6 +1196,60 @@ export default function KadenceButtonEdit(props) {
 																		showReset
 																	/>
 																}
+																context={{
+																	blockName: 'kadence/singlebtn',
+																	attribute: 'borderRadius',
+																	defaultValue:
+																		tokenBinding.borderRadius?.presetValue,
```
d4mation, line 1203: It looks like this is pointing at an incorrect default. If you look at the pre-existing CSS, the default value should be `3px`. When set to this, it uses the Medium Token which is `0.5rem`.

https://github.com/stellarwp/kadence-blocks/blob/SOFT-3906/singlebtn-proving-wire/src/blocks/advancedbtn/style.scss#L50

This could result in sudden changes for existing users who haven't customized their Border Radius, right?

We'll want to make sure that this is set appropriately in `baseline.json`.

src/extension/design-tokens/component-token-ui.js
```diff
@@ -0,0 +1,418 @@
...
+					value={number === '' ? undefined : Number(number)}
+					initialPosition={typeof min === 'number' ? min : 0}
+					min={typeof min === 'number' ? min : 0}
+					max={max}
+					step={step}
+					withInputField={false}
+					onChange={(next) => writeNumber(next)}
+				/>
+			)}
+		</div>
```
d4mation, line 265: These currently don't align nicely vertically. Can we correct that?

image

I also wonder if it is very clear what the range slider is used for with how it is positioned. 🤔

d4mation (commented):
Two small things. Once addressed, I think we're good here :)

d4mation (commented):
Turns out changes here also break saving Preset changes.

```json
{
    "data": {
        "status": 422,
        "block": "kadence\/singlebtn",
        "errors": [
            {
                "path": "$extensions.com.kadence.designTokens.presets.kadence\/singlebtn.large-border-radius.tokens.button-radius",
                "code": "alias_malformed",
                "message": "Value looks like an alias but is not a whole-string \"{dot.path}\" reference."
            }
        ]
    }
}
```

Issue is in `src/extension/preset-picker/capture.js#L26-L31` and something like this could possibly fix it:

```js
if (kind === 'dimension') {
	const dimension = normalizeDimension(value, unit);
	if (dimension.value === '') {
		return '';
	}
	// A token alias is a whole-string reference; the preset stores it as-is, no unit.
	if (isTokenAlias(dimension.value)) {
		return dimension.value;
	}
	return `${dimension.value}${dimension.unit}`;
}
```

But there's a separate limitation as well according to the comment at Line 27 where it will flatten a 4 corner array to the first side. So we will need to update that to properly account for 4 separate corner values as well.

src/blocks/singlebtn/edit.js
pauloiankoski: CleanShot 2026-07-31 at 13 04 33

@d4mation Do you think it would be a good solution to add that as another baseline option so we can make it the default?

src/extension/design-tokens/component-token-ui.js
pauloiankoski: It currently looks fine on my end. I wasn’t able to reproduce the misalignment.

CleanShot 2026-07-31 at 17 41 50

Anyway, I might rework this custom tab in a follow-up by making the unit selector look like the native inputs and removing the range slider, which doesn’t make much sense here, in my opinion.

src/blocks/singlebtn/edit.js
pauloiankoski: Added on https://github.com/stellarwp/kadence-blocks/pull/1173/commits/dba21bfc112888d4a4822d6c99d80e4efd512c9e

pauloiankoski:
@d4mation I have create two new PRs addressing the presets topic. See: https://github.com/stellarwp/kadence-blocks/pull/1202 and https://github.com/stellarwp/kadence-blocks/pull/1203

src/blocks/singlebtn/edit.js
d4mation: This introduces a small possible problem:

If you manually choose `Small`, it shows it as "unsynced" from the Preset because the Default is _actually_ empty rather than what we're showing as the Default value in the picker.

Since `Small` is now `3px`/`0.1875rem` in the baseline, I think we could probably set this default more explicitly rather than making it default to empty. That would solve this small UX quirk while also ensuring that if that Token is updated, it applies to both Blocks _without_ any saved Border Radius configuration and those with `Small` populated, correct?

We were avoiding setting the default in the Block Attributes specifically before, but now that it is Token-backed with one that matches the existing CSS I think we should be fine to do this 🤔 

As long as we can ensure that Button Blocks _without_ a Border Radius and ones with the default of `Small` explicitly defined render as `3px`/`0.1875rem` (matching the pre-Design Tokens styling) and that when the Token Value is updated it is reflected in both of them, I think that injecting the default when it isn't set should be safe. 

image

d4mation (approved):
Approving this as it is largely fine, but we will want to address https://github.com/stellarwp/kadence-blocks/pull/1173#discussion_r3751831126 separately.

src/blocks/singlebtn/edit.js
d4mation: However: We do need to be careful about eagerly populating the default as that breaks assumptions with the Tablet/Mobile settings (where empty means "inherit from Desktop"). This could be complex. All the more reason to fix with a targeted PR :)

pauloiankoski:
@d4mation Created [SOFT-4162](https://linear.app/nexcess/issue/SOFT-4162/make-the-token-backed-default-explicit-instead-of-empty-for-block) for the issue to be addressed and merging this.

PR #1182, opened 2026-07-27
Color Palette Switching [8/9] - Palette selector UI (Style Book + editor)

d4mation:
🎫 https://linear.app/nexcess/issue/SOFT-3990/b4-palette-selector-ui-style-book-editor-color-picker
:video_camera: https://www.loom.com/share/ad3754b723174400a976b86f7e6312c7

The palette selector UI, wiring the B1–B3 engine to the editor and the Style Book.

> **Note: these UI surfaces are temporary.** The editor palette picker and the Style Book Palettes page are interim scaffolding to exercise and validate the palette engine end to end; they are slated to be replaced/removed once the final surfaces land. Please don't critique their design/UX too harshly — review the engine wiring rather than the presentation.

- **Editor** — a per-block **Color Palette** selector in the block inspector (`src/extension/palette-picker`) reads the active set's palettes from the localized `window.kadenceDesignTokensPalettes` catalog (new `Editor\Palette_Catalog`) and writes the `kbPalette` attribute (B2), so a block renders against a chosen palette while its siblings follow the set `$current`. Surfaced in the Design Tokens panel in `early-filters.js`.
- **Style Book** — a **Palettes** page: a palette dropdown, a "Set as Current" action that switches the set's `$current`, and the current palette's color groups with per-swatch color editing. Backed by a `use-palettes` hook and the palette REST client (`api/paths` + `api/client`, wired to B3), plus a Palettes nav section.

Swatch edits and the current-palette switch write through the palette REST surface.

Stacked on B3 (#1181). Completes the SOFT-3982 tree.

### Checklist


includes/resources/Design_Tokens/Editor/Palette_Catalog.php
pauloiankoski, line 1: Every other catalog class here (`Variant_Catalog`, `Block_Preset_Catalog`, and `Pickable_Tokens_Catalog`) has its own test file, but `Palette_Catalog` doesn’t. Can we add a `Palette_CatalogTest.php` that covers the `all()` output shape and the label fallback behavior (when a palette node doesn’t have a string label)?

pauloiankoski (approved):

includes/resources/Design_Tokens/Editor/Palette_Catalog.php
d4mation: [2d17d52](https://github.com/stellarwp/kadence-blocks/pull/1182/commits/2d17d5287bc83eca6a90177f78ba03006370e736) :)

PR #1184, opened 2026-07-27
[DTM] SOFT-3992 [2/3]: rename Variant to Preset

pauloiankoski:
## Summary

Renames the per-block skin concept from **Variant** to **Preset** across every layer — PHP classes/dirs, the REST contract, the JS editor extension, and the persisted formats. The editor UI already said "Preset"; this brings the code in line. Second of three stacked phases for SOFT-3992. Stacked on Phase 1.

## What changed

- **PHP**: `Projection/Variant/` → `Projection/Preset/`; `Variant_Set` → `Preset_Bindings` (the per-block binding *definition*, one per block — deliberately not `Preset_Set`); `Variant_Resolver` → `Preset_Resolver`, `Effective_Variants` → `Effective_Presets`, `Variant_Value_Normalizer` → `Preset_Value_Normalizer`, `Unknown_Variant_Exception` → `Unknown_Preset_Exception`, `Variants_Controller` → `Presets_Controller`, `Variant_Catalog` → `Preset_Catalog`, `Admin/Feed/Variants` → `Presets`. The unrelated `Editor/Block_Preset_Catalog` → `Attribute_Default_Catalog` (so "preset" is not overloaded a third way).
- **REST** (namespace `kb-design-tokens/v1` unchanged): base `variants` → `presets`, routes/params/schema title, and the `description` schema strings.
- **JS**: `src/extension/variants/` → `presets/`, `variant-picker/` → `preset-picker/`, the components, the `kadenceDesignTokensVariants` global, filter handles/fns, and API path builders.
- **Persisted formats**: the `variants` JSON key → `presets` (token document + `baseline.json`, `$extensions.com.kadence.designTokens.presets.<block>`); block attribute `kbVariant` → `kbPreset`; saved CSS class `kb-variant--<slug>` → `kb-preset--<slug>`.

## Scope note for reviewers

The internal multi-set switch/namespace machinery inside the projection CSS builders (and the `data-kb-token-set` selectors) is **intentionally left un-renamed** — SOFT-3982 A2 (SOFT-3985) is deleting it. This PR renames the `Projection/Variant/` → `Projection/Preset/` directory/namespace and the preset concept, but leaves those doomed internals with their current names; the file-move-vs-rewrite overlap with SOFT-3985 is reconciled separately.

## Test plan

- PHPStan (level max, baseline untouched), PHPCS, cspell, ESLint green.
- Full `Resources/Design_Tokens` wpunit suite (973 tests) + jest (123 tests) green, including the renamed test files and the `baseline.json` snapshot.
- Manual editor smoke recommended before merge: the Preset picker renders, selecting a preset writes `kbPreset` and the `kb-preset--<slug>` class, and Save-as-Preset round-trips through the renamed REST routes.


includes/resources/Design_Tokens/Registry/Preset_Bindings.php
d4mation, line 1: A lot of the text in this class still refers to them as "Preset Sets". If we want to rename this to _Bindings_, we should update those as well and further clarify the class doc.

Additionally, it looks like we're using `preset_set` in places like `Token_Registry` which should be updated to match.

d4mation (approved):

PR #1185, opened 2026-07-27
[DTM] SOFT-3992 [3/3]: rename token set to Token Library

pauloiankoski:
## Summary

Renames the whole-token-document concept from **token set / active set** to **Token Library** in code (**"Library"** in UI copy). Applied to the surviving single-active-library foundation only. Third of three stacked phases for SOFT-3992. Stacked on Phase 2.

Naming follows a prefix-aware convention: `Token_Library` in class/component names that lack a token-conveying prefix; bare `library` where a `design-tokens` prefix already means "token" (so the JS global is `kadenceDesignTokensLibraries`, the option `..._active_library`, the REST route `active-library`); visible UI copy says "Library" (never "token").

## What changed

- **PHP**: `Active_Set_Store` → `Active_Token_Library_Store`, `Set_Catalog` → `Token_Library_Catalog`, `Active_Set_Controller` → `Active_Token_Library_Controller`; `register_active_set_store()` → `register_active_token_library_store()`; `Token_Store`/`Token_Table` docblock wording "token set" → "token library".
- **REST** (namespace unchanged): base `active-set` → `active-library`, routes, schema title `design-token-active-set` → `design-token-active-library`, and the `description` strings.
- **Persisted / wire**: option `kadence_blocks_design_tokens_active_set` → `..._active_library`; action hook `..._active_changed` → `..._active_library_changed`.
- **JS**: global `kadenceDesignTokensSets` → `kadenceDesignTokensLibraries`; `activeSet()` → `activeLibrary()`; `DEFAULT_TOKEN_SET_SLUG` → `DEFAULT_LIBRARY_SLUG`.
- **UI copy**: the surviving "The token set changed…" dialog strings → "The library changed…" (in the Add/Rename/Delete primitive dialogs).

## Scope note for reviewers

Per the single-active-library direction (SOFT-3982), the per-block library override surface — `kbTokenSet`, `data-kb-token-set`, the `token-set-picker` extension, and the `SWITCH_ATTR`/switch/namespace machinery — is being **deleted** by SOFT-3982 (A1/A3), so this PR intentionally does **not** rename it. Those identifiers remaining in the tree is expected; only the surviving foundation is renamed. References to renamed symbols inside the doomed `token-set-picker` were updated just enough to keep the build green.

## Test plan

- PHPStan (level max, baseline untouched), PHPCS, cspell, ESLint green.
- Full `Resources/Design_Tokens` wpunit (973 tests) + jest (123 tests) green, including the renamed store/catalog/controller test files.
- Manual: switching the active library re-emits the `:root` token block; the surviving foundation resolves under the renamed option/REST/global.


d4mation (commented):
There's a handful of files that still reference "Token Set" in some way:

- Documents_Controller
- Token_History_Store
- Block_Default_Css/Css_Builder
- Document_Write_Pipeline
- Token_Resolver
- User_Primitives_Controller
- Presets_Controller

Sometimes the reference is just `set` (like the `DELETED_ACTION` doc in `Token_Store`) so we should be careful to catch those too.

Those references wouldn't be removed by SOFT-3982, so they should be updated accordingly.

Also, `SET_PARAM` should be updated to be `LIBRARY_PARAM` most likely for cohesion as well.

d4mation (approved):

PR #1186, opened 2026-07-27
Color Palette Switching [9/9] - Per-block palette re-skins semantics and variants (full color-graph switch layer)

d4mation:
Linear: [SOFT-4019](https://linear.app/nexcess/issue/SOFT-4019/b5-per-block-palette-re-skins-semantics-and-variants-full-color-graph)
:video_camera: https://www.loom.com/share/ad3754b723174400a976b86f7e6312c7

Extends the palette feature (B1-B4) so a color palette fully re-skins a block: full color coverage, sparse **delta** storage over the default palette, and a per-block switch layer that re-skins direct-token colors, semantics, shadow composites, and variant/button colors.

## Coverage + partial-delta model

- **Coverage** — `brand.button`, `brand.button-hover`, and `neutral.200` (border) are swatches in the **default** palette at their current values. The primary button keeps its own primitive so the shipped default look is unchanged, but a palette can now override the button/border. `sunset`/`forest` carry a `brand.button(-hover)` delta so a Primary Button pinned to them re-skins. (Transparent block backgrounds stay off-palette on purpose.)
- **Partial storage + full-field UI** — `GET /palettes/{id}` returns the effective view (the full default field set, each swatch flagged `overridden` or inherited); writes persist only the **deltas** (swatches that differ from the default); the default palette keeps every swatch. The Style Book page shows all fields with an "inherited from default" hint.

## Per-block re-skin (semantics + variants)

A block pinned to a palette previously re-tinted only the palette's **primitive** colors. Semantics that alias them, variant/button colors, and shadow composites are declared at `:root`, and CSS custom-property late-binding resolves them where they are declared, so re-declaring a primitive on a subtree did not re-resolve them.

- **Resolve each palette's full color graph** — `Token_Resolver::resolve_palette($slug, $paletteId)` resolves the whole token graph as if a given palette were current; `Effective_Palettes::overlay_for_palette()` supplies that palette's color overlay.
- **Emit the resolved graph per palette** — `[data-kb-palette="<id>"]` re-declares every color var any palette re-tints (primitives, the semantics that alias them, and shadow composites) as flat literals, so direct-token consumers (a heading reading `semantic--color--text`) re-skin. Only vars a palette actually changes are emitted; a color no palette touches is left to inherit.
- **Re-resolve variants under a shared rule** — a shared attribute-presence `[data-kb-palette]` rule re-emits the canonical `--kb-token--variant--*` declarations so a variant var re-resolves against the subtree's re-tinted semantics. A variant/button whose color aliases a palette-changed token follows the chosen palette, respecting variant selection (a secondary variant re-tints to the palette's secondary color, not the primary).
- **Version-key the variant collect memo** — the variant projection's `collect()` memo is keyed on the store version alongside the slug, so a write (or a second reader such as the palette layer) builds against the current document rather than a pre-write structure.

Accepted v1 limitation: a block that reads a numbered `--global-paletteN` bridge directly (not through a token or a variant) still resolves it at `:root`. Buttons and variants route through the variant vars, so they re-skin.

Stacked on B4 (#1182). Full `Resources/Design_Tokens` suite green (992 tests), phpstan/phpcs/cspell/eslint clean, assets rebuilt.

### Checklist


includes/resources/Design_Tokens/Registry/declarations.php
```diff
@@ -147,6 +147,47 @@ static function ( string $slug ): array {
...
+		'label' => __( 'Neutral 300', 'kadence-blocks' ),
+		'group' => __( 'Palette', 'kadence-blocks' ),
+	],
+	[
+		'id'    => 'primitive.color.neutral.200',
+		'type'  => 'color',
+		'label' => __( 'Neutral 200', 'kadence-blocks' ),
+		'group' => __( 'Palette', 'kadence-blocks' ),
+	],
+];
```
d4mation, line 189: As I mentioned in my Loom, we may need to update this list. Things get weird particularly with Button Variations and how they're being pointed at the Neutral Palette right now due to them not having an actual color source in the current Kadence Blocks release.

pauloiankoski (approved):

pauloiankoski (approved):

PR #1188, opened 2026-07-28
[DTM] SOFT-3904 [1/3]: Shared PHP-JS conformance fixture parity

pauloiankoski:
## Summary

The anti-drift keystone for design-token aliases. A block control can store a token reference as a whole-string alias (`{dot.alias}`), and that value is resolved twice: in PHP for the front end (`Alias::path_of` + `Css_Var::from_id`) and in JS for the editor preview (`resolveTokenAlias`). If the two ever disagree by one character, an authored token renders one way in the editor and another on the front end.

This phase makes one JSON fixture the executable contract for both languages. Tests only — no shipped code changes.

## What this adds

- **`tests/wpunit/Resources/Design_Tokens/Schema/Vocabulary/AliasConformanceTest.php`** — a new wpunit test that loads the *same* fixture the jest suite loads (`src/extension/design-tokens/__tests__/fixtures/token-alias-conformance.json`, no second copy) and asserts, per entry:
  - each alias: `Alias::is_alias()` is true, and `'var(' . Css_Var::from_id( Alias::path_of( $alias ) ) . ')'` is byte-identical to the fixture's expected `cssVar`;
  - each non-alias: `Alias::is_alias()` is false;
  - the loose-vs-strict boundary: braced-but-malformed values (`{bad path}`, `{unclosed`, `unopened}`, `{}`) are `looks_like_alias()` true but `is_alias()` false, while plain literals (`palette1`, `16px`) are false for both;
  - a guard that both fixture sections are non-empty, so a truncated fixture fails loudly instead of passing vacuously.
- **Fixture edge case** — added `{semantic.icon-size.default}` → `var(--kb-token--semantic--icon-size--default)` to exercise the hyphen+dot mixed-id sanitization edge. Both suites pick it up automatically.
- **JS gap-fill** — added the `KadenceColorOutput` / `colorValue` filter-seam pass-through cases to `alias-css-output.test.js` (the existing file covered `dimensionValue` but never the color seam). The loose-vs-strict boundary was already covered by the fixture-driven cases in `alias.test.js`.

## Acceptance property

Verified by hand: a one-character edit to any fixture `cssVar` fails **both** `bun run test` and the wpunit conformance test. One fixture, two suites, cannot drift.

## Test plan

- `slic run wpunit Resources/Design_Tokens/Schema/Vocabulary/AliasConformanceTest` — green (28 tests).
- `bun run test` — green.
- cspell clean on changed files; `phpstan-baseline.neon` untouched.

## Stacking

Part of the SOFT-3904 stack, based on `SOFT-3992/phase-3-token-set-to-library` (this work merges after the SOFT-3992 rename stack). Followed by `[2/3]` (BC snapshots) and `[3/3]` (emission + fail-open).

### DIFF
--- a/src/extension/design-tokens/__tests__/alias-css-output.test.js
+++ b/src/extension/design-tokens/__tests__/alias-css-output.test.js
@@ -19,6 +19,7 @@ import { registerTokenAliasFilters } from '../register-filters';
 
 const helpersRoot = path.dirname(require.resolve('@kadence/helpers/package.json'));
 const KadenceBlocksCSS = require(path.join(helpersRoot, 'dist/cjs/css/index.js')).default;
+const KadenceColorOutput = require(path.join(helpersRoot, 'dist/cjs/kadence-color-output/index.js')).default;
 
 const ALIAS = '{semantic.radius.media}';
 const ALIAS_VAR = 'var(--kb-token--semantic--radius--media)';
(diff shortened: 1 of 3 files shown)


d4mation (approved):
Nice!

PR #1190, opened 2026-07-28
[DTM] SOFT-3904 [3/3]: Alias emission + fail-open matrix

pauloiankoski:
## Summary

The positive and negative halves of the render-site contract, at the real render sites (not just the predicate level `[1/3]` pins). Tests only — no shipped code changes.

## What this adds

**`tests/wpunit/KadenceBlocksCssTokenEmissionTest.php`** — 71 tests across four blocks:

1. **Per-method bare `var()` emission** — a strict alias fed through each representative `render_*` method (`render_number`, `render_color`, `render_range`, `render_measure_range`, `render_measure`, `render_border_radius`, `render_responsive_range`, `render_shadow`, typography line-height/letter-spacing, `get_border_value` width) emits the exact `var(--kb-token--<id>)` on the expected property.
2. **Fail-open matrix** — a `Generator` provider crosses the malformed inputs (`{bad path}`, `{unclosed`, `unopened}`, `{}`) against 11 methods. Per case: the output carries no `var(` / no `--kb-token--`; the call never fatals (a `is_numeric`-adjacent TypeError on a brace string is exactly the regression this guards); and the method returns its own documented normal empty/false/no-declaration result. The per-method normal results were confirmed by reading each method, including the `render_typography` asymmetry (line-height passes a malformed value through under `!empty()`, letter-spacing rejects it under `is_numeric()`).
3. **Per-side dimension mixing** — `render_measure`, `render_measure_range`, `render_border_radius` with exactly one alias entry and three numeric siblings, the alias rotated through every side/corner so a per-side loop ordering bug cannot hide.
4. **Inline box-shadow** — `render_shadow` with the alias in each of hOffset / vOffset / blur / spread individually, asserting `var()` lands in the correct shorthand position.

## Test plan (full ticket)

- `slic run wpunit` — all three ticket classes green: `AliasConformanceTest` (28), `KadenceBlocksCssBcSnapshotTest` (45), `KadenceBlocksCssTokenEmissionTest` (71).
- `bun run test` — green (128 tests).
- cspell clean; single shared fixture (no second copy); no edits to `includes/` or non-test `src/`.

## Stacking

Final phase of the SOFT-3904 stack. Based on `[2/3]` (BC snapshots).

### DIFF
--- a/tests/wpunit/KadenceBlocksCssTokenEmissionTest.php
+++ b/tests/wpunit/KadenceBlocksCssTokenEmissionTest.php
@@ -1,0 +1,203 @@
+<?php declare( strict_types=1 );
+
+namespace Tests\wpunit;
+
+use Generator;
+use Kadence_Blocks_CSS;
+use Tests\Support\Classes\TestCase;
+
+/**
+ * Render-site contract for the alias recognizer's positive and negative halves.
+ *
+ * For every relaxed `render_*` method in {@see Kadence_Blocks_CSS}, a strict `{dot.alias}`
+ * value must come out as a bare `var(--kb-token--<id>)` in the built CSS (emission), a
+ * malformed brace string must never mint a `var()` or reach numeric handling as if it were
+ * an alias (fail-open), a 4-side/4-corner value with exactly one alias side must mix the
+ * `var()` with the sibling numeric sides correctly (per-side mixing), and an alias placed in
+ * any single box-shadow part must land in the right position of the shadow shorthand
+ * (inline box-shadow).
+ */
+final class KadenceBlocksCssTokenEmissionTest extends TestCase {
+
+	protected $css;
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->css = new Kadence_Blocks_CSS();
+	}
+
+	protected function tearDown(): void {
+		parent::tearDown();
+	}
+
+	/**
+	 * render_number returns the bare var() reference for a strict alias, ignoring the unit.
+	 *
+	 * @return void
+	 */
+	public function testRenderNumberEmitsBareVarForAlias(): void {
+		$this->assertSame( 'var(--kb-token--semantic--radius--media)',
+			$this->css->render_number( '{semantic.radius.media}', 'px' ),
+			'render_number must emit the bare var() for a strict alias' );
+	}
+
+	/**
+	 * render_color returns the bare var() reference for a strict alias.
+	 *
+	 * @return void
+	 */
+	public function testRenderColorEmitsBareVarForAlias(): void {
+		$this->assertSame( 'var(--kb-token--semantic--radius--media)',
+			$this->css->render_color( '{semantic.radius.media}' ),
+			'render_color must emit the bare var() for a strict alias' );
+	}
+
+	/**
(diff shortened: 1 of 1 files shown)


tests/wpunit/KadenceBlocksCssTokenEmissionTest.php
```diff
@@ -0,0 +1,748 @@
...
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->css = new Kadence_Blocks_CSS();
+	}
+
+	protected function tearDown(): void {
+		parent::tearDown();
+	}
```
d4mation, line 32: We shouldn't need this `tearDown()` call.

tests/wpunit/KadenceBlocksCssTokenEmissionTest.php
```diff
@@ -0,0 +1,748 @@
...
+
+	/**
+	 * render_number returns the bare var() reference for a strict alias, ignoring the unit.
+	 *
+	 * @return void
+	 */
+	public function testRenderNumberEmitsBareVarForAlias(): void {
+		$this->assertSame( 'var(--kb-token--semantic--radius--media)',
+			$this->css->render_number( '{semantic.radius.media}', 'px' ),
+			'render_number must emit the bare var() for a strict alias' );
```
d4mation, line 42: The line breaks on your assertions are a little strange. This would make it a little more readable.

```suggestion
		$this->assertSame(
			'var(--kb-token--semantic--radius--media)',
			$this->css->render_number( '{semantic.radius.media}', 'px' ),
			'render_number must emit the bare var() for a strict alias'
		);
```

d4mation (approved):
Minor issues, but the tests are great! Thanks!

tests/wpunit/KadenceBlocksCssTokenEmissionTest.php
pauloiankoski: Resolved in https://github.com/stellarwp/kadence-blocks/pull/1190/commits/0663d688a3e3ebce0f0653a0608222f2e3c53f1f

tests/wpunit/KadenceBlocksCssTokenEmissionTest.php
pauloiankoski: Resolved in https://github.com/stellarwp/kadence-blocks/pull/1190/commits/0663d688a3e3ebce0f0653a0608222f2e3c53f1f

d4mation: I'm not sure exactly what happened, but that commit is empty...? 😅 https://github.com/stellarwp/kadence-blocks/commit/0663d688a3e3ebce0f0653a0608222f2e3c53f1f

I saw you closed and opened the PR though, so I figure you're working on it :)

pauloiankoski: Oh, I ran into an issue while committing the changes. I have a pre-commit hook that runs the CI tests locally, and while it was running, I tried to amend the commit with some additional changes. That’s what caused the mess.

I’m pushing everything to the base branch now that I’ve merged all the PRs. 800769115928527d064d54b525eea19c4ec17dee

PR #1192, opened 2026-07-28
SOFT-3992: rename proposal for the SOFT-3982 palette work

pauloiankoski:
## Summary

Applies the SOFT-3992 terminology rename to the palette work on this stack, so the palette branches stay compatible once SOFT-3992 lands in the feature base. This is a **reconciliation proposal** — it renames only this stack's own additions to match SOFT-3992's new vocabulary:

- **Style Book → Style Library** (admin screen: dirs, entry, handles, DOM ids, CSS/BEM prefix, UI copy)
- **Variant → Preset** (`Projection/Variant/` → `Projection/Preset/`, `Variant_Resolver` → `Preset_Resolver`, `variant-picker/` → `preset-picker/`, `kbVariant` → `kbPreset`, the `variants` document key → `presets`, etc.)
- **token set / active set → Token Library / Library** (code says `library`; UI says "Library"): `Set_Catalog` → `Token_Library_Catalog`, the JS globals, option/action keys, REST `active-set` → `active-library`, and the presets REST param `set` → `library`.

Branched off `soft-3982-palette-coverage-and-delta-model` (the palette tip), so the diff is rename-only.

## Heads-up: CI here will be red until SOFT-3992 merges

This branch references the renamed class/symbol **definitions** that SOFT-3992 introduces. Those definitions do not exist on this pre-SOFT-3992 base yet, so a standalone run (PHPStan / wpunit / build) fails by design. It is meant to be applied **after** SOFT-3992 (phases 1-3) reach the feature base.

## How it was verified

A throwaway integration branch was assembled — `feature/design-tokens-module` + SOFT-3992 (phases 1-3) + this branch — and checked:

- **Zero rename leftovers** (broad grep across `includes/` and `src/`, doomed surfaces excluded).
- **PHPStan**: no errors.
- **jest**: 127 tests pass. **ESLint**: no import/reference errors.
- **Merge**: no conflicts beyond the documented collapse/deletion resolutions (the doomed multi-set surfaces this stack already deletes win; renamed-but-deleted files like the old `Set_Catalog` are dropped, not resurrected). No soft-3982 functionality was changed — only names.

## What is intentionally left alone

The doomed per-block multi-set surfaces this stack removes (`kbTokenSet`, `data-kb-token-set`, `token-set-picker`, the switch/namespace machinery in the `Css_Var`/`Preset` projection builders) are **not** renamed — renaming code that is being deleted would be wasted churn.

## Also surfaced (fixed on the SOFT-3992 PRs, not here)

The integration check found a real defect in SOFT-3992 itself: the presets REST client sent `?set=` while the controller now reads `library`, plus some leftover "token set" JS docblocks and a `binding set` / `binding library` over-rename. Those were fixed on the SOFT-3992 phase PRs.


d4mation:
@pauloiankoski I'm not sure if we need this anymore as I fixed the renames in these files while addressing the merge conflicts. I think it may be best to close this one and if we notice anything that was missed, fix it directly. Thanks for this, though! It would have helped _a ton_ if the SOFT-3982 branches got in before the rename :)

pauloiankoski:
Closing this in favor of the fixes that were applied directly to the PRs.

PR #1194, opened 2026-07-29
SOFT-3870: keep token-backed palette colors in color controls under "Use only Custom Colors"

d4mation:
:ticket: https://linear.app/nexcess/issue/SOFT-3870/token-backed-palette-colors-hidden-in-kadence-color-controls-when-use
:video_camera: https://www.loom.com/share/3054b09c52ac4e719efb43ff4dd127ea

With "Use only Custom Colors" enabled, the shared Kadence color controls hid the token-backed global-palette colors (`palette1`..`palette9`), while `core/button` still showed them. This restores them and makes selecting one palette-aware.

- **Retain + alias the palette swatches.** A listener on the `kadence.components.popColorControl.colors` seam (added in `@kadence/components` stellarwp/kadence-components#18) re-adds the global-palette slots under the override and sets each swatch value to the token alias it is bound to. Selecting one stores `{dot.alias}`, which resolves to `var(--kb-token--<id>)` and follows the active color palette instead of freezing to a fixed hex. `src/early-filters.js` registers it at editor load.
- **Localized slot -> alias map (single source of truth).** `Palette_Catalog` emits a `slots` map (`slug -> {alias}`) on `window.kadenceDesignTokensPalettes`, resolved through `Kadence_Palette_Slot` so the slot set can never drift from the projection declared in `declarations.php`. The listener reads that map and is a no-op when it is empty (design-tokens registry inactive), so a site with no active tokens keeps its literal palette colors.
- **Pin.** `@kadence/components` is pinned to the commit carrying the seam and the shared helpers.

Covered by a `Palette_Catalog` `slots` wpunit test and the listener unit test.

### Scope

The controls preview the active color palette. Making the control accurately preview a per-block palette override is out of scope here and is handled by the later, global color-control rework.

### Checklist


package.json
```diff
@@ -29,7 +29,7 @@
 		"@dnd-kit/sortable": "^10.0.0",
 		"@floating-ui/react": "^0.27.12",
 		"@hcaptcha/react-hcaptcha": "^1.8.1",
-		"@kadence/components": "github:stellarwp/kadence-components#1.1.9",
+		"@kadence/components": "github:stellarwp/kadence-components#SOFT-3870/color-control-override-seam",
```
pauloiankoski: Let’s consider setting this to use the `feature/design-tokens-module` branch, as that dependency is about to be merged.

pauloiankoski (approved):

package.json
d4mation: Done in 6898b85 :)

Leaving this PR open until all the other stacked PRs of yours are merged to avoid making those merges more difficult.

PR #1200, opened 2026-07-30
fix: Improve type handling and return values

rahulacaleffi:
🎫  

This PR fixes the PHPStan errors on `release/M26.lanturn` that were introduced on latest releases.

Detailed breakdown:
- Correct the parameter and return docblocks on `field_name()` and `field_id()` in the advanced form input block, and cast their return value to a string. The false `@return void` was surfacing as errors in the blocks that consume them.
- Cast the `sanitize_unique_id()` return value to a string in the abstract block, and fix a few typos in its docblocks.
- Narrow the `library_url` request value to a string before passing it into typed methods in the prebuilt library REST API. No behaviour change.
- Check the stored option shape before reading it in `API_Url_Trait::get_saved_library_urls()`.
- Correct the array check in `Image_Trait::sanitize_image_sizes_array()` so non-array input is skipped instead of looped over.
- Adjust docblocks to the correct response types.

### Checklist

### DIFF
--- a/includes/resources/Traits/API_Url_Trait.php
+++ b/includes/resources/Traits/API_Url_Trait.php
@@ -143,14 +143,17 @@ trait API_Url_Trait {
 	 * @return string[]
 	 */
 	protected function get_saved_library_urls(): array {
-		$settings = json_decode( (string) get_option( 'kadence_blocks_cloud' ), true );
+		$stored   = get_option( 'kadence_blocks_cloud' );
+		$settings = is_string( $stored ) ? json_decode( $stored, true ) : null;
 		$urls     = [];
 
-		if ( ! empty( $settings['connections'] ) && is_array( $settings['connections'] ) ) {
-			foreach ( $settings['connections'] as $connection ) {
-				if ( ! empty( $connection['url'] ) ) {
-					$urls[] = rtrim( (string) $connection['url'], '/' );
-				}
+		if ( ! is_array( $settings ) || empty( $settings['connections'] ) || ! is_array( $settings['connections'] ) ) {
+			return $urls;
+		}
+
+		foreach ( $settings['connections'] as $connection ) {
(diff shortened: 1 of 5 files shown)


includes/resources/Traits/API_Url_Trait.php
```diff
@@ -143,10 +143,11 @@ protected function is_kadence_api_url( string $url ): bool {
...
-		$settings = json_decode( (string) get_option( 'kadence_blocks_cloud' ), true );
+		$stored   = get_option( 'kadence_blocks_cloud' );
+		$settings = is_string( $stored ) ? json_decode( $stored, true ) : null;
 		$urls     = [];
 
-		if ( ! empty( $settings['connections'] ) && is_array( $settings['connections'] ) ) {
+		if ( is_array( $settings ) && ! empty( $settings['connections'] ) && is_array( $settings['connections'] ) ) {
 			foreach ( $settings['connections'] as $connection ) {
 				if ( ! empty( $connection['url'] ) ) {
 					$urls[] = rtrim( (string) $connection['url'], '/' );
```
gilbert-hernandez, line 153: @rahulacaleffi 

I'm a little confused here. The check in the foreach loop is now different. I see in commit 214b4077599559a446b8ecdd0f7dca2997802bdf that there was a check for:

`if ( is_array( $connection ) && ! empty( $connection['url'] ) && is_scalar( $connection['url'] ) )`

Was this intentional or is it a regression?

rahulacaleffi: @gilbert-hernandez sharp eyes! It was intentional. I wanted to clear what wasn't really necessary to the PHPStan fixes, however, it might be good to have it to avoid any warnings in case it's not an array. I brought back the `is_array( $connection )` piece.

gilbert-hernandez (approved):
@rahulacaleffi Thanks for adding the guard back! LGTM.

PR #1202, opened 2026-07-31
[DTM] SOFT-3906: per-corner preset values + the save-as-preset 422

pauloiankoski:
## Summary

Addresses the review on #1173: saving a preset returned a 422 when a corner held a design token, and the capture flattened four corners to one.

- A picked token alias is now stored verbatim. The capture used to append the control's unit, producing `{primitive.dimension.radius.sm}px`, which the server rejects as `alias_malformed` — `Alias::looks_like_alias()` fires on a brace anywhere in the value.
- A preset property can hold a **per-corner slot list**: exactly four values, each an alias or a literal, so a button with different corners saves as a preset instead of narrowing to its first side. Four identical corners still collapse to a single scalar, so every existing preset is byte-identical and there is no migration.
- The Border Radius control's linked/individual mode now follows the corners the user actually sees, and is tracked per breakpoint.

## Shape

A preset property stays a bare value unless the corners differ:

```json
"button-radius": "{semantic.radius.control}"
"button-radius": ["{radius.none}", "{radius.sm}", "{radius.full}", "{radius.lg}"]
```

Exactly four slots — "every corner" is already expressed by a scalar, so a shorter list would be a second spelling of the same thing.

## What changed

- `Dtcg_Validator` — accepts a four-slot list, each slot validated by the same alias-or-literal rule as a scalar, so "alias anywhere" stays one rule applied once.
- `Presets_Controller` — new `guard_slot_arrays()` rejects a slot list on a non-dimension property. The schema validator cannot make that call: the key it walks is a property name like `button-radius`, not a token id, so it never resolves a `$type`. The registry does know, so the check sits with the other registry-aware guards. `guard_aliases_resolve()` now walks every corner.
- `Preset_Resolver` — `flatten()` returns the slots **unjoined** so the editor reads each corner; `project()` joins them into a `var()` shorthand for CSS. Joining in `flatten()` would hand the editor one opaque string its per-corner display and compare cannot parse.
- `Preset_Value_Normalizer` — matches each corner on its own, so a captured per-corner literal re-joins the theming cascade like a scalar does.
- `matchesPreset` — compares corner by corner against a per-corner preset value. Without this a block matching its preset exactly would read as "overridden".
- The component seam narrows a per-corner inherited default to the slot being rendered.

## Two fixes to the linked/individual mode

Both were only reachable once per-corner presets existed.

1. The mode was derived from the block's own attribute, which is **empty** when a preset supplies the values — so it trivially read as "all equal" and opened linked, hiding a preset's four different corners. It now derives from the effective corners: the stored corner where set, otherwise what that corner inherits.
2. The responsive control keeps one mode but writes three attributes, so linking while on Tablet rewrote Desktop's corners. Mode and the session override are now per device.

## Also here

`semantic.radius.control` now resolves to the button's long-standing 3px rather than 8px, and the radius scale gains the steps that made that expressible (`xs` 2px, `sm` 3px, `md` 6px, `lg` 8px, `xl` 16px). Reviewed and approved separately — without it, every uncustomized button would have silently changed size.

## Test plan

- [x] PHPStan clean at level max, no baseline entries added
- [x] Full wpunit suite, snapshot suite byte-identical (the BC guard that matters — every existing preset is a scalar)
- [x] jest, phpcs, cspell clean
- [x] Verified in the editor: saved presets covering every capture case — four different corner tokens, mixed alias + literal, uniform alias collapsing to a scalar, and mixed literals — then confirmed each projects the expected CSS

### DIFF
--- a/src/blocks/singlebtn/edit.js
+++ b/src/blocks/singlebtn/edit.js
@@ -85,7 +85,11 @@ import { addFilter, applyFilters, doAction } from '@wordpress/hooks';
 import BackendStyles from './components/backend-styles';
 import { PresetButton } from '../../extension/preset-picker/PresetButton';
 import { usePresetBinding, resetAttr } from '../../extension/token-indicators';
-import { deriveMeasureMode, measureAttrsForDevice } from '../../extension/token-indicators/normalize';
+import {
+	deriveMeasureMode,
+	inheritedMeasureSlots,
+	measureAttrsForDevice,
+} from '../../extension/token-indicators/normalize';
 import { TokenLabel } from '../../extension/token-indicators/components/TokenLabel';
 import { TokenControlRow } from '../../extension/token-indicators/components/TokenControlRow';
 
@@ -277,6 +281,15 @@ export default function KadenceButtonEdit(props) {
 	// reset that clears the mapped attribute back to the preset value (served by the existing scoped CSS).
 	const tokenBinding = usePresetBinding('kadence/singlebtn', attributes);
 	const resetToken = (attr) => resetAttr(attr, setAttributes, tokenBinding[attr]?.kind);
+
+	// What an unset Border Radius corner falls back to on the active device: another breakpoint's corner
+	// before the preset's, matching the cascade the button actually renders through. The corners stay
+	// stored-empty — this only tells the field's popover which size is in effect and where it came from.
+	const inheritedBorderRadius = inheritedMeasureSlots(
+		previewDevice,
+		{ desktop: borderRadius, tablet: tabletBorderRadius },
+		tokenBinding.borderRadius?.presetValue
+	);
 
 	useEffect(() => {
 		setAttributes({ inQueryBlock: getInQueryBlock(context, inQueryBlock) });
@@ -1215,8 +1228,8 @@ export default function KadenceButtonEdit(props) {
 																context={{
 																	blockName: 'kadence/singlebtn',
 																	attribute: 'borderRadius',
-																	defaultValue:
-																		tokenBinding.borderRadius?.presetValue,
+																	defaultValue: inheritedBorderRadius.values,
+																	inheritedDefault: inheritedBorderRadius.inherited,
 																	unit: borderRadiusUnit,
 																	units: ['px', 'em', 'rem', '%'],
 																	onUnit: (value) =>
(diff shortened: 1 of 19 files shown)


src/blocks/singlebtn/edit.js
```diff
@@ -1218,36 +1234,43 @@ export default function KadenceButtonEdit(props) {
...
-																		// (now all-equal) reads as linked.
-																		const corner = borderRadius?.[0] ?? '';
+																		// Collapse to the top-left corner so the button
+																		// reflects the link immediately, and drop the
+																		// override so the derived mode (now all-equal)
+																		// reads as linked. Only the ACTIVE device's
+																		// attribute is written, so the other breakpoints
+																		// keep their own corners.
+																		const corner =
+																			borderRadiusForDevice.value?.[0] ?? '';
```
d4mation: There's a weird scenario here for Tablet/Mobile:

1. Set customized corners on Desktop
2. Switch to Tablet or Mobile
3. It'll default to "Linked" at the Small Token size (because it is empty) instead of defaulting to Desktop's settings.

This could be potentially confusing, right? 🤔 

I think that having it default to "Empty" to inherit from Desktop by default is the right call, but we will need to have it show what is _actually_ set for Desktop instead of showing the default value and we will need to display to the user whether we're inheriting from Desktop or whether it is explicitly set for Tablet/Mobile.

pauloiankoski: Refactored to be empty when no selection is made, inheriting values from the previous device breakpoint.

src/blocks/singlebtn/edit.js
```diff
@@ -1217,37 +1246,68 @@ export default function KadenceButtonEdit(props) {
...
+																			// Tablet/Mobile to Desktop's current radius.
+																			// Remember the choice instead, so the control
+																			// links without the corners leaving inherit.
+																			setBorderRadiusModeOverride((current) => ({
+																				...current,
+																				[previewDevice]: 'linked',
+																			}));
+
+																			return;
+																		}
```
d4mation: Weird inconsistency in behavior:

1. Create a button
2. Set the Corners like this:
    - Blank
    - Any Token
    - Any Token
    - Any Token
3. Link the corners
4. The "linked" field is blank, but the other corners are still set to the Token

This _only_ happens if the Top-Left corner is blank. If Top-Left has a Token assigned, then linking the corners results in them all matching the Top-Left corner.

I think that "collapsing" to what is set for Top-Left makes sense (we have to standardize it to one of them, and the first one is good), but it should be able to clear out the others if Top-Left itself is cleared out.

pauloiankoski: Fixed in 2e5cbb1. The early return fired whenever the top-left was blank, so the other corners kept their tokens. It now fires only when all four are empty, so a blank top-left clears the rest.

d4mation (approved):

PR #1203, opened 2026-07-31
[DTM] SOFT-4085: responsive preset values (per-breakpoint overrides)

pauloiankoski:
## Summary

A block already stores a different border radius per breakpoint, and each of those can hold four per-corner values. A preset could not. This adds that, following the pattern responsive tokens already use.

Two concrete gaps closed:

1. A preset could not carry its own per-breakpoint values — the preset resolver and preset CSS builder had no breakpoint concept at all.
2. **"Save as a new preset" silently dropped tablet and mobile.** The capture read only the desktop attribute, so a user who set per-breakpoint corners lost them with no warning — the same class of bug as the corner flattening in #1202.

## What already worked, and needed no code

A preset pointing at a **token** was already responsive, per corner: the token layer emits `@media` redeclarations of `--kb-token--*` at `:root`, and each `var()` in a per-corner shorthand resolves independently at use time. Verified in the browser with a probe — redeclaring one token inside a media query changes only that corner.

So this PR is only about a preset carrying values the tokens do not.

## Shape

Mirrors the responsive token leaf exactly — `semantic.font-size.control` in the shipped baseline — so the override map sits next to the value it overrides:

```json
"button-radius": {
  "$value": ["{radius.lg}", "{radius.none}", "{radius.lg}", "{radius.none}"],
  "$extensions": {
    "com.kadence.designTokens": {
      "responsive": {
        "tablet": ["{radius.md}", "{radius.none}", "{radius.md}", "{radius.none}"],
        "mobile": ["4px", "0px", "4px", "0px"]
      }
    }
  }
}
```

Mirroring the leaf rather than inventing a preset-level node buys two things: `Responsive::has_responsive()` and `responsive_of()` work verbatim, and `validate_extension_tokens()` needs no change — the envelope hangs off the property, which that walk already visits. A preset-level `responsive` sibling would have needed one, since that loop returns as soon as it finds a `tokens` key and never descends to siblings.

A property with no breakpoints stays a bare value, so every existing preset is untouched and there is no migration.

## What changed

- `Extensions::preset_value_of()` / `preset_responsive_of()` — one reader, so the unwrap never spreads as `isset($x['$value'])` across the resolver, guards, catalog and capture.
- `Dtcg_Validator` — envelope branch. Each override is just another preset value, so it goes through `validate_extension_value()` unchanged; only the breakpoint-key check is new.
- `Preset_Resolver` — unwraps the base; adds `resolve_responsive()` and `resolve_responsive_literal()`. An override whose alias does not resolve drops that breakpoint only, leaving the base intact.
- `Projection/Preset/Css_Builder` — redeclares the preset var inside each breakpoint's `@media`, mirroring the token builder. Only the canonical var is redeclared: the scoped `--global-*` / `--kb-*` bridges already point at it, so every corner follows.
- Both REST write guards walk base plus every override. This also fixed `guard_slot_arrays()` wrongly rejecting *any* envelope on a non-dimension property, since it tested `is_array()` on the raw entry.
- `Preset_Value_Normalizer` aliases per breakpoint; `Preset_Catalog` exposes a per-breakpoint map for the editor.
- New `responsive_attrs` binding key on `button-radius`. The block names its per-device attributes by a prefix convention (`borderRadius` -> `tabletBorderRadius`), which is a naming rule rather than something safely derivable, so it is declared instead of string-built in JS.

## Test plan

- [x] PHPStan clean at level max, no baseline entries added
- [x] Full wpunit suite, snapshot suite byte-identical
- [x] jest, phpcs, cspell clean
- [x] Verified in the editor: saved a preset from a button with a different per-corner radius on desktop, tablet and mobile; confirmed it stores the envelope and projects the base plus both `@media` blocks

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Preset/Css_Builder.php
+++ b/includes/resources/Design_Tokens/Projection/Preset/Css_Builder.php
@@ -520,19 +520,6 @@ final class Css_Builder {
 	}
 
 	/**
-	 * The canonical preset var name for a (block, preset, property):
-	 * "--kb-token--preset--<block>--<preset>--<property>", e.g.
-	 * --kb-token--preset--kadence-singlebtn--secondary--button-bg.
-	 *
-	 * @since TBD
-	 *
-	 * @param string $block    The block name.
-	 * @param string $preset   The preset slug.
-	 * @param string $property The block property.
-	 *
-	 * @return string
-	 */
-	/**
 	 * Redeclare each preset var that varies by breakpoint inside that breakpoint's `@media` block.
 	 *
 	 * Only the canonical var is redeclared — the scoped `--global-*` / `--kb-*` bridges are untouched,
@@ -611,6 +598,19 @@ final class Css_Builder {
 		return md5( (string) wp_json_encode( $breakpoints ) );
 	}
 
+	/**
+	 * The canonical preset var name for a (block, preset, property):
+	 * "--kb-token--preset--<block>--<preset>--<property>", e.g.
+	 * --kb-token--preset--kadence-singlebtn--secondary--button-bg.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $block    The block name.
+	 * @param string $preset   The preset slug.
+	 * @param string $property The block property.
+	 *
+	 * @return string
+	 */
 	private function preset_var( string $block, string $preset, string $property ): string {
 		return Css_Var::get_prefix()
 			. self::PRESET_SEGMENT
--- a/includes/resources/Design_Tokens/Projection/Preset/Css_Builder.php
+++ b/includes/resources/Design_Tokens/Projection/Preset/Css_Builder.php
@@ -148,12 +148,15 @@ final class Css_Builder {
 	 *
 	 * @since TBD
 	 *
-	 * @param string $active_slug The active library's slug.
+	 * @param string                $active_slug The active library's slug.
+	 * @param array<string, string> $breakpoints Breakpoint => media-query string (e.g.
+	 *                                           "tablet" => "(max-width: 1024px)"), for the per-breakpoint
+	 *                                           redeclarations. Empty emits none.
 	 *
 	 * @return string The CSS, or an empty string when no block contributes a slot-targeted value.
 	 */
-	public function css( string $active_slug ): string {
-		return $this->build( $active_slug );
+	public function css( string $active_slug, array $breakpoints = [] ): string {
+		return $this->build( $active_slug, $breakpoints );
 	}
 
 	/**
@@ -166,13 +169,15 @@ final class Css_Builder {
 	 *
 	 * @since TBD
 	 *
-	 * @param string $active_slug The active library's slug.
-	 * @param string $version     The store version the active library was built from.
+	 * @param string                $active_slug The active library's slug.
+	 * @param string                $version     The store version the active library was built from.
+	 * @param array<string, string> $breakpoints Breakpoint => media-query string, for the per-breakpoint
+	 *                                           redeclarations.
 	 *
 	 * @return string
 	 */
-	public function css_for_version( string $active_slug, string $version ): string {
-		$cache_key = 'preset_css_root_' . KADENCE_BLOCKS_VERSION . '_' . $active_slug . '_' . $version;
+	public function css_for_version( string $active_slug, string $version, array $breakpoints = [] ): string {
+		$cache_key = 'preset_css_root_' . KADENCE_BLOCKS_VERSION . '_' . $active_slug . '_' . $version . '_' . $this->breakpoint_signature( $breakpoints );
 
 		if ( isset( $this->memo[ $cache_key ] ) ) {
 			return $this->memo[ $cache_key ];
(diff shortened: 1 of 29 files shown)


includes/resources/Design_Tokens/Projection/Preset/Css_Builder.php
```diff
@@ -525,6 +532,85 @@ private function block_selector( string $block ): string {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string                $active_slug The active library's slug.
+	 * @param array<string, array{selector:string, default:string, presets:array<string, array<string, array{target:string, value:string}>>}> $collected The collected preset structure, for the block/preset list.
+	 * @param array<string, string> $breakpoints Breakpoint => media-query string.
+	 *
+	 * @return string
+	 */
+	private function responsive_blocks( string $active_slug, array $collected, array $breakpoints ): string {
```
d4mation, line 539: This definition was placed below the docblock for `preset_var()` which broke that association 😅 Could you fix that, please?

pauloiankoski: Fixed on https://github.com/stellarwp/kadence-blocks/pull/1203/commits/6e3a1e584a7adee00e3363308df42b113bb843bd

src/extension/preset-picker/capture.js
d4mation, line 1: Related issue, but technically not introduced in this PR:

1. Create a Button
2. Modify the Desktop and Tablet Border Radii
3. Save as a new Preset

The Border Radii will be unset because `appendPreset` adds it to the catalog's presets list but doesn't add the _values_ for that preset. Once `onSaved` runs, it clears out the changes in-memory. Additionally, when I tried to modify the Border Radii again, it claims we haven't diverged from the Preset (likely because it cannot find the Preset data properly) and won't let me create a new one until I switch to a baked in one like "Primary Button" again.

src/extension/preset-picker/capture.js
```diff
@@ -80,6 +80,51 @@ function attrToLiteral(kind, value, unit, presetValue) {
...
+	const responsive = {};
+
+	Object.entries(property.responsive_attrs || {}).forEach(([breakpoint, attr]) => {
+		const raw = get(attributes, attr, '');
+
+		if (isEmptyValue(property.kind, raw)) {
+			return;
+		}
+
+		const value = attrToLiteral(property.kind, raw, unit, presetValue);
```
d4mation: 1. Create a button
2. Set Desktop to "Linked" and choose a Token
3. Set only the Top-Left for Tablet to a Token
4. Save the Preset

This will result in the following being saved:

```json
"button-radius": {
  "$value": "{primitive.dimension.radius.xl}",
  "tablet": ["{primitive.dimension.radius.full}", "{semantic.radius.control}", "{semantic.radius.control}", "{semantic.radius.control}"]
}
```

Tablet thinks it should be saving the default Token for the other corners instead of inheriting from Desktop.

pauloiankoski: Fixed in 387ae6a. The unset corners were filled from the preset instead of the cascade. Tablet now falls back to Desktop, Mobile to Tablet then Desktop, so your steps save `[full, xl, xl, xl]`.

src/extension/preset-picker/capture.js
pauloiankoski: Fixed in 387ae6a. That was the cause: with no values in the catalog, every control unbinds, which explains all three symptoms. The save now seeds them (aliases resolved to literals), and delete drops them.

d4mation: I think this is still slightly broken.

The issue now is that Tablet seems to still aggressively inherit from Desktop even when a Tablet Border Radius was set for the Preset.

When following the same steps as above, it says it applied the Preset correctly, but only Desktop's seems to get applied properly in the editor.

I ended up recording a Loom because I think it may be easier in this case: https://www.loom.com/share/1c0ba5472e1b4dd59a653ad2a31c0afa

src/extension/preset-picker/capture.js
d4mation: Hmm, is it possible to have it save nothing for the other corners on Tablet? Because this ends up hardcoding in what Desktop was at the time of saving.

It likely doesn't matter because you can't edit a Preset from this page anyway, but we'll want to be sure we end up handling the save in the same way between this page and the Style Library page.

src/extension/preset-picker/capture.js
pauloiankoski: Reproduced from the Loom. The save and the projected CSS were right — the button did render `9999px` on Tablet — but the controls never read the preset's breakpoint values, so every device fell back to Desktop. Fixed in e68fabd.

Also from the Loom: an unset field now prints the value in effect in muted text instead of blank, with nothing written to the block (d1ed3ee). Per your Slack note it stays visually empty only when the fallback is the baseline rather than a preset.

tablet-preset-value-in-field

src/extension/preset-picker/capture.js
pauloiankoski: Not without a projection change: a per-corner value projects as one custom property holding a four-value shorthand, and the breakpoint redeclares that same var — one property, four corners, all or nothing. Expressing a gap means changing the value model, the resolver and the projection together (~1.5-2 days), and it touches contracts the stacked Style Library PRs build on.

Tracked as SOFT-4167 so both writers land on the same corner model. Note the fill only happens when SOME corners are set — linking all four (your Loom steps) saves exactly what you picked.

d4mation (approved):

PR #1209, opened 2026-08-04
Support custom design library locations

rahulacaleffi:
🎫  [SMTNC-1912](https://linear.app/nexcess/issue/SMTNC-1912/pattern-hub-fails-to-load-when-using-a-child-theme-after-kadence)

The library hardening in 3.7.8.1 only accounted for the Kadence hosts and for libraries saved through Cloud Connect. Design libraries registered in PHP with the existing `kadence_blocks_custom_prebuilt_libraries` filter were left out, so their tab still opened in the editor but nothing inside it loaded: categories, patterns and previews all came back as `Invalid Request, Unknown Library`.

This PR widens the list of locations the editor can request. It brings `kadence_blocks_custom_prebuilt_libraries` back into play and adds a new `kadence_blocks_allowed_library_urls` filter for adding or removing locations, while keeping the list built server-side so it stays under the site owner's control. A site owner can also drop URLs straight into the array in the plugin, if necessary, even though the filter is the one always recommended.

Detailed breakdown:
- `API_Url_Trait::get_allowed_library_urls()` merges the Kadence hosts, saved Cloud Connect connections and libraries registered in PHP, and has a commented spot for a site owner to add their own.
- New `kadence_blocks_allowed_library_urls` filter for adding or removing a location without editing the plugin.
- A location the editor could not load is recorded, and an admin notice lets a site manager approve or dismiss it. Approved locations are stored and merged into the list above. Recording and approving are both limited to users who can `manage options`.
- Matching is by host, subdomains included, so a connection saved with a trailing slash and a starter pack served from a subdomain both work again.
- Fixed the design library sidebar erroring when the remembered tab no longer matches an available library, which left the block unable to render.
- Existing integrations need no changes on their side.
- Tests cover both filters, host and subdomain matching, saved connections, and malformed input.

### Checklist


bordoni (approved):

rahulacaleffi:
@bordoni , requesting your review again since a couple changes have been pushed. Thanks in advance.

1. Added the option to approve/disapprove URLs, if they were added via 'enqueue_block_editor_assets' (JS).
2. A small fix for a bug found during the tests, that was preventing Design Library modal to open, in case a Connection Cloud added doesn't exist anymore.

Artifact:
Screenshot 2026-08-05 at 10 01 35 AM

bordoni (approved):

gilbert-hernandez (approved):
LGTM. I remember hoping that the allowlist was correct in 3.7.8.1.

PR #1210, opened 2026-08-05
[DTM] SOFT-4075 [1/6]: App shell, routing, and the design-token layer

pauloiankoski:
Removes the superseded Style Library prototype and puts the new app frame in its place: the shell every screen renders into, query-string routing, the wp-admin layout, and the design-token layer the rest of the stack builds on.

## What this adds

- **App shell** — `AppShell` with header, sidebar, content and settings-panel slots; the content column shrinks when the panel opens. `AppHeader` renders the title with slots for the library selector and header actions.
- **Routing** — pure `parseRoute`/`serializeRoute` helpers plus `useStyleLibraryRoute`, syncing the active screen and the open settings target to the URL via the History API. The args are namespaced `kb-screen` and `kb-item`: wp-admin is shared URL space, and an unprefixed arg can collide with core or another plugin. Unknown query args are always preserved.
- **Design-token layer** — split into `_primitives.scss` and `_semantic.scss` with a hard contract: a primitive holds a raw value and references nothing; a semantic token holds only a `var()` reference and is named for its role. Values mirror the design system's tables (gray ramp, grid-unit spacing, radii, font sizes, line heights, weights) and are expressed in `rem`, except border widths and the radius sentinel.
- **wp-admin integration** — the app fills the viewport below the admin bar with the sidebar and content scrolling independently; the admin footer is hidden and the four core admin-notice hooks are emptied, both scoped to this screen only.

## Notes for review

- Tokens are declared on the **page body class**, not the app root, because `@wordpress/components` portals popovers and modals outside the root where root-scoped custom properties do not resolve.
- The shipped screen loses the old prototype's token editors until the per-screen tickets land. That is the agreed trade for rebuilding the frame properly.

## Screenshot

phase-1-app-shell-and-routing

## Test plan

- Load **Kadence → Style Library**: the frame fills the viewport, the page body does not scroll, no admin footer or notices appear inside the app.
- `?kb-screen=` and `?kb-item=` round-trip through back/forward; `page=kadence-blocks-style-library` is never clobbered.
- `bun run test`, `bun run lint-js`, `slic run wpunit Resources/Design_Tokens`, `slic run snapshot`.


src/style-library/hooks/use-style-library-route.js
```diff
@@ -0,0 +1,49 @@
...
+
+	const apply = useCallback((partial, method) => {
+		setRoute((current) => {
+			const next = { ...current, ...partial };
+
+			window.history[method](null, '', serializeRoute(next, window.location.href));
+
+			return next;
+		});
+	}, []);
```
d4mation, line 54: This updates `window.history` in a state updater, which React StrictMode could double-invoke.

pauloiankoski: Refactored on https://github.com/stellarwp/kadence-blocks/pull/1210/commits/d53b1fa75c0ec7475f320fa81d41a664a4ef410d

d4mation (approved):

PR #1212, opened 2026-08-05
[DTM] SOFT-4075 [3/6]: Library selector and live switching

pauloiankoski:
Wires the header's library selector to the existing REST surface, adds a shared select component, and replaces the page reload on library switch with an in-place refresh.

## What this adds

- **Library selector** — lists libraries, switches the active one, creates new ones, and deletes or resets. Deleting the `default` library resets it rather than removing it, and the confirmation copy says so.
- **`SelectDropdown`** — a generic select (value, options, change handler, one optional trailing action) knowing nothing about libraries. The Color Palette screen's palette selector and the screen header's inline control reuse it; `LibrarySelector` is a thin composition over it.
- **`title` on the documents REST payload** — the store already held a library's title but the controller never echoed it, so the UI could only show slugs. The collection reuses `list_stores()`'s single query, so surfacing it costs no extra queries.
- **Live switching** — a new `GET /kb-design-tokens/v1/feed/{slug}` returns the admin feed for a library, and the app refreshes in place instead of reloading. First paint still reads the server-printed global, so startup is unchanged.

## Notes for review

- `Admin\Feed\Feed_Assembler::for_slug()` is the **single** path both the page-load localizer and the new endpoint use. If those ever diverged, a switched library would render differently from a freshly loaded one — `Feed_ControllerTest` asserts the two payloads are byte-for-byte identical for the same library.
- An untitled default library displays as "Your Library". That is a display fallback; nothing is written to the database and the REST contract still returns an empty title.

## Screenshot

phase-3-library-selector

## Test plan

- Switch libraries: the page does not reload and every screen reflects the new library.
- Create a library; a name colliding with an existing one is reported on the field rather than silently disabling the button.
- `slic run wpunit Resources/Design_Tokens/Rest`, `slic run snapshot`, `bun run test`.


src/style-library/hooks/use-libraries.js
```diff
@@ -0,0 +1,129 @@
...
+ *                  switchLibrary, createLibrary, deleteLibrary }`.
+ */
+export function useLibraries(feed, refreshFeed) {
+	const [libraries, setLibraries] = useState([]);
+	const [isLoading, setIsLoading] = useState(true);
+	const [isBusy, setIsBusy] = useState(false);
+	const [switchError, setSwitchError] = useState(null);
+	const [createError, setCreateError] = useState(null);
+	const [deleteError, setDeleteError] = useState(null);
+	const activeSlug = feed?.slug;
```
d4mation: This should fallback to `DEFAULT_LIBRARY_SLUG` to be safe. It shouldn't ever be undefined, but since we have that Constant anyway we should use it.

src/style-library/components/organisms/DeleteLibraryModal.js
```diff
@@ -0,0 +1,125 @@
...
+						<Notice status="error" isDismissible={false}>
+							{error.message}
+						</Notice>
+					)}
+					<p>
+						{isDefault
+							? sprintf(
+									// translators: %s: the default library's display title.
+									__(
+										"Reset '%s'? Its token values return to the shipped baseline.",
```
d4mation: While accurate, I don't think users would understand what the "baseline" is.
```suggestion
										"Reset '%s' to its default values?",
```

src/style-library/helpers/libraries.js
```diff
@@ -0,0 +1,121 @@
...
+ * @return {string} The display name.
+ */
+export function libraryDisplayTitle(library) {
+	if (library?.title) {
+		return library.title;
+	}
+
+	if (isDefaultLibrary(library?.slug)) {
+		return __('Your Library', 'kadence-blocks');
+	}
```
d4mation: Maybe this default should be derived from PHP rather than within the JS here? This way `library.title` is defined normally.

pauloiankoski: Addressed on https://github.com/stellarwp/kadence-blocks/pull/1212/commits/15e40c6ad0aec837c11f2f10bff5c07fc2cc9a6b, https://github.com/stellarwp/kadence-blocks/pull/1212/commits/263716723f96035fdcc0f429621f9b0f70a68ec2 and https://github.com/stellarwp/kadence-blocks/pull/1212/commits/3a4d04ae18e3739d84d8f1ec1cf8fd9069b907a8

src/style-library/hooks/use-libraries.js
pauloiankoski: Addressed on https://github.com/stellarwp/kadence-blocks/pull/1212/commits/e51f91f7ac42c33a2139a6e80c6fd3fdd1191aa2

d4mation (approved):

PR #1214, opened 2026-08-05
[DTM] SOFT-4075 [5/6]: Settings panel and field library

pauloiankoski:
Adds the settings panel and a schema-driven field library, so the per-screen tickets describe their settings as JSON instead of assembling JSX.

## What this adds

- **`SettingsPanel`** — header with close, optional state tabs, a scrolling field area, and a sticky footer. Delete renders only for user-created items, gated on `isDeletable`; baseline items never show it.
- **`SettingsForm`** — walks a schema's panels and fields, resolving each `type` through the registry. Titled panels use core's collapsible `PanelBody`, keeping its borders and the `-1px` collapse that stops adjacent dividers doubling.
- **Twelve field types** — `text`, `number-unit`, `range-number`, `select`, `stepper`, `unit`, `toggle`, `color`, `color-list`, `token-select`, `box-sides`, `shadow`, each wrapping a native `@wordpress/components` control where one fits.
- **Schema helpers** — normalization, `fieldComponentFor`, dot-path get/set, dirty tracking. An unregistered type is dropped and warned once in development.
- **Per-field responsiveness** — eligible fields carry their own breakpoint switcher in the label row, reading and writing the backend's base-plus-`tablet`/`mobile` shape. Fields that are not responsive-capable are unchanged.

## Notes for review

**The fields are a first implementation and are expected to improve as the screens that use them are built.** What is settled, and expensive to change later, is the *contract* the screen tickets author against:

- the `{field, value, onChange}` shape and `readOnly`
- the schema's panels-and-fields structure and how a type resolves
- **`box-sides`' value shape: a 4-element positional array, `[top, right, bottom, left]`**, matching the per-corner storage work on the related ticket. The same order serves radius and side properties; only the decorating icons differ.

Renaming edits a display **label** — the token id is immutable and shown read-only. The panel does not call the id-rewriting rename endpoint.

The pickable-token payload is now attached to the Style Library bundle as well as the editor's, so `token-select` has a catalogue to offer. The app reads it through its own accessor and stays independent of `src/extension/`.

## Screenshot

phase-5-settings-panel-and-field-library

## Test plan

- Dev build, then `?kb-gallery=1&kb-item=demo`: the panel opens with every field type rendering, the content column shrinks, and the footer stays pinned while the fields scroll.
- Toggle a breakpoint on one responsive field; the other fields keep their own.
- `bun run test`, `slic run wpunit Resources/Design_Tokens`, `slic run snapshot`.


d4mation:
Easier to comment like this for this PR I think

---

z-index issues with the color picker:

image

Weird icon next to numeric inputs (that I guess reset to defaults?)

Also, the unit gets clipped a bit.

image

Unlinked radius spills off the side

image

pauloiankoski:
@d4mation The fields in the settings included in this PR have not been very much detailed. They look better in the PRs where they are used.

d4mation (approved):

PR #1216, opened 2026-08-05
[DTM] SOFT-4075 [6/6]: Explicit library activation, rename, and safer delete

pauloiankoski:
## Summary

Separates **the library the site renders with** from **the library you are editing**, and puts changing the former behind an explicit confirmation.

Before this, those were the same variable. The only way to look at another library was to make it live, so picking an option in the header dropdown immediately restyled the whole front end — no confirmation, no warning, no undo. Creating a library did the same thing implicitly, because the create flow chained a switch onto its own success.

Now the dropdown only changes what you are editing. Nothing a visitor sees moves until someone reads the confirmation and agrees.

## What changed

**Two slugs instead of one.** `useLibraries` returns `activeSlug` (the site pointer) and `editingSlug` (what the app shows). They start equal and diverge as soon as you open another library. `setActiveLibrary` now has exactly two callers: the activation flow, and the successor step of delete.

**`switchLibraryFlow` split in two** — `openLibraryFlow` (refreshes the feed, never touches the pointer) and `activateLibraryFlow` (writes the pointer, and deliberately does *not* refresh the feed, since activating the library already on screen changes nothing inside it).

**Deleting the active library requires naming its successor.** The confirm button stays disabled until one is chosen, and the successor is activated *before* the delete — the reverse order would let the server fall the pointer back to the default in between, briefly serving a site-wide look nobody picked. When only one library could take over, the picker is skipped and the copy names that library instead: the site can land in exactly one place, so asking would be a question with a single answer. The choosing is dropped, never the telling.

**Rename**, via a dedicated `PUT /documents/{slug}/title` route. This is the one PHP change and it is worth a look: sending a title alongside an empty document to a bulk write route merges and then re-validates the entire stored document, so a rename fails for any library whose document does not currently validate — for reasons that have nothing to do with the new name. This is not hypothetical; it is how the default library behaved. The new route writes the label column alone, bumps no version, and fires no change action.

**Selector indicators.** A muted `Default` badge marks the library that can only be reset, never deleted — answering "why can't I delete this one?" before anyone tries. A theme-colored `Active` badge marks the one the site is serving.

**A whole-app scrim while a library loads**, replacing the small spinner beside the selector. Opening a library replaces the values behind every screen at once, so a spinner tucked next to one control understated it and left stale content looking current and clickable. Driven by its own flag, not `isBusy` — rename and activate change nothing on screen, and blanking the app for those would be theatre.

**The header names the library on first paint.** The feed payload now carries `title`, so the toggle no longer shows a slug-derived guess and then corrects itself when the libraries request lands.

## Notable decisions

- **Title only, never slug.** The slug is the library's identity — the active pointer stores it, the table has a `UNIQUE KEY` on it, every route addresses by it — and it is invisible once a title exists.
- **Rename needs its own duplicate check.** `slugifyLibraryTitle` derives the slug at creation time only, so slug and title diverge permanently after a rename. Reusing the slug-based check would reject renaming to a name nothing on screen displays. Creation keeps the slug check, because that is where the slug is minted.
- **The successor dropdown has no preselection**, but is skipped entirely when there is only one candidate. Preselecting is the worst of both: a field the user can skim past while the destination quietly defaults. Naming the library in the sentence puts it in the text they have to read to reach the button. Worth knowing this is the common path, not a corner — the default library can never be deleted, so any site with two libraries meets it the first time it deletes the active one.
- **Modal z-index is scoped to this admin page.** Core stacks popovers above modals on purpose, so that a popover opened *inside* a modal works; overriding that globally would break those.
- **Activation's partial-failure case is documented, not designed away.** If activation succeeds and the delete then fails, the site sits on a library the user explicitly chose and the target survives — a retry finishes the job. The alternative ordering trades that for default styles nobody picked.

## Screenshots

**Selector while editing the active library** — the muted `Default` badge marks the library that can only be reset, never deleted; the themed `Active` badge marks the one the site is serving.

phase-6-selector-badges-editing-active

**Selector while editing a non-active library** — the check sits on the library being edited, the `Active` badge stays on the one the site serves, and `Set as active` appears in the header. This is the split the whole PR is about, in one frame.

phase-6-selector-editing-non-active-library

**Set as active** — names both libraries and states the site-wide consequence before the confirm. Cancel is focused on open.

phase-6-set-as-active-confirmation

**Rename** — `Save` is disabled while the name is unchanged, so a rename that would spend a request to change nothing cannot be submitted.

phase-6-rename-library

**Deleting the active library** — the successor picker, unselected, with the confirm button disabled until it is answered. On a site where only one library could take over, this picker is replaced by copy naming that library outright.

phase-6-delete-active-library-successor

## Test plan

- `npx jest src/style-library` — 162 pass (25 new across the flows and helpers)
- `slic run wpunit Resources/Design_Tokens/Rest/V1/DocumentsControllerTest` — 62 pass (9 new)
- `slic run wpunit Resources/Design_Tokens/Rest/V1/Feed_ControllerTest` — 8 pass (2 new)
- `slic run snapshot` — 3 pass (Builder snapshot gained one key)
- `vendor/bin/phpstan analyse`, `composer phpcs`, `npx eslint`, `cspell`, `npm run build` — all clean

Regression tests worth knowing about, because each one fails if an implicit behavior comes back:

- opening a library never calls `setActiveLibrary`
- creating a library never calls `setActiveLibrary`
- deleting the active library without a successor issues no request at all
- the successor is activated *before* the delete (asserted through a shared call log, since the ordering is the correctness property)
- renaming works against a stored document that is not even valid JSON, and leaves it byte-identical

## Manual verification

Not yet done end to end. Worth confirming on a real site: renaming the default library now succeeds, and a rename leaves token values intact (the PATCH-vs-PUT hazard is invisible in the modal itself).

### DIFF
--- a/includes/resources/Design_Tokens/Database/Token_Store.php
+++ b/includes/resources/Design_Tokens/Database/Token_Store.php
@@ -333,6 +333,8 @@ final class Token_Store extends Query {
 	 * @throws DatabaseQueryException If the update fails.
 	 */
 	public function save_title( string $slug, string $title ): bool {
+		$title = sanitize_text_field( $title );
+
 		if ( $title === '' ) {
 			return false;
 		}
@@ -576,6 +578,8 @@ final class Token_Store extends Query {
 		if ( $slug !== null ) {
 			$data = [ 'slug' => $slug ] + $data;
 		}
+
+		$title = sanitize_text_field( $title );
 
 		if ( $title !== '' ) {
 			$data['title'] = $title;
--- a/includes/resources/Design_Tokens/Database/Token_Store.php
+++ b/includes/resources/Design_Tokens/Database/Token_Store.php
@@ -290,6 +290,70 @@ final class Token_Store extends Query {
 	}
 
 	/**
+	 * Write a token library's human-readable label, touching nothing else.
+	 *
+	 * A title is metadata: it appears in the library picker and nowhere in any projection, so
+	 * renaming a library must not travel through the document write path. Doing so would re-validate
+	 * and re-resolve the entire stored document to change a label — turning a rename into a failure
+	 * for any library whose stored document does not currently validate, for reasons that have
+	 * nothing to do with the new name.
+	 *
+	 * Neither the version hash nor the change action moves here, for the same reason: no projector
+	 * output depends on a title, so bumping the version would invalidate the projected-CSS cache for
+	 * a change no visitor can see.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $slug  The token library slug.
+	 * @param string $title The new label. An empty string is rejected rather than treated as
+	 *                      "leave the stored title alone", which is what the document write path
+	 *                      does with an empty title — here, clearing is simply not an operation.
+	 *
+	 * @return bool True when the title was written.
+	 *
+	 * @throws DatabaseQueryException If the update fails.
+	 */
+	public function save_title( string $slug, string $title ): bool {
+		if ( $title === '' ) {
+			return false;
+		}
+
+		$existing = $this->qb()
+						->where( 'slug', $slug )
+						->get( ARRAY_A );
+
+		if ( ! is_array( $existing ) ) {
+			// The default library has no row until something is written to it, so naming it is that
+			// first write. The empty document stored alongside the title is not a change of content —
+			// it is exactly what the library already resolved to, since no row means "render from
+			// baseline". Every other slug is rejected by the caller before reaching this point.
+			try {
+				$this->qb()->insert( $this->build_document_data( '', $title, $slug ) );
+			} catch ( DatabaseQueryException $e ) {
+				// Duplicate-key on a concurrent first write → the row now exists and carries someone
+				// else's title; report the miss rather than a fatal.
+				return false;
+			}
+
+			return true;
+		}
+
+		$this->qb()
+			->where( 'slug', $slug )
+			->update(
+				[
+					'title'      => $title,
+					'updated_at' => current_time( 'mysql', true ),
+				]
+			);
+
+		// Deliberately not gated on the affected-row count: MySQL reports zero rows affected when an
+		// UPDATE sets a column to the value it already holds, which is a no-op, not a failure. A real
+		// write failure throws instead.
+		return true;
+	}
+
+	/**
 	 * Delete a named token library only when the stored version matches expected_version.
 	 *
 	 * The default library is never row-deleted; use save_document_conditional with an empty
--- a/includes/resources/Design_Tokens/Database/Token_Store.php
+++ b/includes/resources/Design_Tokens/Database/Token_Store.php
@@ -309,6 +309,70 @@ final class Token_Store extends Query {
 	}
 
 	/**
+	 * Write a token library's human-readable label, touching nothing else.
+	 *
+	 * A title is metadata: it appears in the library picker and nowhere in any projection, so
+	 * renaming a library must not travel through the document write path. Doing so would re-validate
+	 * and re-resolve the entire stored document to change a label — turning a rename into a failure
+	 * for any library whose stored document does not currently validate, for reasons that have
+	 * nothing to do with the new name.
+	 *
+	 * Neither the version hash nor the change action moves here, for the same reason: no projector
+	 * output depends on a title, so bumping the version would invalidate the projected-CSS cache for
+	 * a change no visitor can see.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $slug  The token library slug.
+	 * @param string $title The new label. An empty string is rejected rather than treated as
+	 *                      "leave the stored title alone", which is what the document write path
+	 *                      does with an empty title — here, clearing is simply not an operation.
+	 *
+	 * @return bool True when the title was written.
+	 *
+	 * @throws DatabaseQueryException If the update fails.
+	 */
+	public function save_title( string $slug, string $title ): bool {
+		if ( $title === '' ) {
+			return false;
+		}
+
+		$existing = $this->qb()
+						->where( 'slug', $slug )
+						->get( ARRAY_A );
+
+		if ( ! is_array( $existing ) ) {
+			// The default library has no row until something is written to it, so naming it is that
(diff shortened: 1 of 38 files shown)


includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
```diff
@@ -351,6 +360,23 @@ public function register_routes(): void {
...
+					// rename is a metadata edit, and routing it through a document write would
+					// re-validate and re-resolve the whole stored document to change a label.
+					'methods'             => 'PUT',
+					'callback'            => [ $this, 'update_title' ],
+					'permission_callback' => [ $this, 'update_item_permissions_check' ],
+					'args'                => $this->get_title_params(),
+				],
+				'schema' => [ $this, 'get_item_schema' ],
+			]
+		);
```
d4mation, line 391: Any reason for a "Get Title" route to not be added as well?

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
```diff
@@ -351,6 +360,23 @@ public function register_routes(): void {
...
 
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/' . self::SLUG_ROUTE . '/' . self::TITLE_ROUTE,
+			[
+				[
+					// Its own route rather than the title parameter the bulk write routes accept: a
+					// rename is a metadata edit, and routing it through a document write would
+					// re-validate and re-resolve the whole stored document to change a label.
+					'methods'             => 'PUT',
```
d4mation: Any reason to exclude `POST`/`PATCH`? 

`PUT` is "more correct", but WP is WP and people likely would expect all three defined in some way 🫠 

We use all three for "Set Current Palette" for instance.

src/style-library/components/organisms/ActivateLibraryModal.js
```diff
@@ -0,0 +1,91 @@
...
+			<p>
+				{sprintf(
+					// translators: %s: the library the site currently uses.
+					__(
+						'You can switch back at any time, but any custom values in "%s" stay in that library — they are not copied over.',
+						'kadence-blocks'
+					),
+					currentTitle
+				)}
+			</p>
```
d4mation, line 77: This isn't explicitly true when switching from Default -> Another Library.

If a Token isn't set in a new library, then it inherits from Default.

image

src/style-library/components/organisms/LibrarySelector.js
```diff
@@ -40,46 +50,70 @@ import { CreateLibraryModal } from './CreateLibraryModal';
...
+		if (library.slug === activeSlug) {
+			badges.push({ text: __('Active', 'kadence-blocks'), variant: 'state' });
+		}
+
+		return {
+			value: library.slug,
+			label: libraryDisplayTitle(library),
+			badges,
+		};
+	});
```
d4mation, line 86: I wonder if "Default" vs "Active" could be confusing here?

image

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
pauloiankoski: https://github.com/stellarwp/kadence-blocks/pull/1216/commits/771e8d089e93605bf905add2726596569eefafe8

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
pauloiankoski: https://github.com/stellarwp/kadence-blocks/pull/1216/commits/120105b60004b88811fc13505dc1bbe9591e87d3

src/style-library/components/organisms/ActivateLibraryModal.js
pauloiankoski: I think the second library is currently resolving from the baseline instead of the default library. Anyway, I didn't get which part of the copy should be changed. Could you help me with that?

src/style-library/components/organisms/LibrarySelector.js
pauloiankoski: Open to ideas. My goal here was to have the Default tag just like we did for the border radius dropdown, and I added the Active because I got confused when trying to use the checkmark for that. I tried to use the checkmark to show the active one, but when opening the dropdown and seeing the checkmark not in the current selected option was confusing. I introduced that alongside the "Set as active" option.

includes/resources/Design_Tokens/Database/Token_Store.php
```diff
@@ -289,6 +289,70 @@ public function save_document_conditional(
...
+	 * @param string $slug  The token library slug.
+	 * @param string $title The new label. An empty string is rejected rather than treated as
+	 *                      "leave the stored title alone", which is what the document write path
+	 *                      does with an empty title — here, clearing is simply not an operation.
+	 *
+	 * @return bool True when the title was written.
+	 *
+	 * @throws DatabaseQueryException If the update fails.
+	 */
+	public function save_title( string $slug, string $title ): bool {
```
d4mation, line 335: We should sanitize this `$title` to be safe.

It is being sanitized when accepted by the REST API already, but this will protect it at this layer too.

src/style-library/components/organisms/ActivateLibraryModal.js
d4mation: I am misremembering based on reverted PRs. This is how it worked for like a week and it doesn't anymore 🫠 I'm remembering now that the whole Color Palette system was to supersede that idea.

Don't take a week off work, apparently.

src/style-library/components/organisms/LibrarySelector.js
d4mation: I'm not sure how to rename it :/ Let's leave it for now. We can easily change it later.

includes/resources/Design_Tokens/Database/Token_Store.php
pauloiankoski: https://github.com/stellarwp/kadence-blocks/pull/1216/commits/a0eb16b0c532e96294ba278803d89fa69de3f9e4

d4mation (approved):

PR #1218, opened 2026-08-05
[DTM] SOFT-4076 [1/5]: Palette data layer

pauloiankoski:
The data layer for the Color Palette screen. Pure helpers and flows only — no UI.

## What's here

- `helpers/palettes.js` — the palette-view-to-`SwatchGrid` mapper, immutable node edits (add / remove / rename / reorder a swatch), and the custom-color naming helpers.
- `helpers/palette-flows.js` — open, activate, create, delete, the settings-panel save, swatch removal, add color, add color group, and reorder. Mirrors the `helpers/library-flows.js` + `hooks/use-libraries.js` split: pure orchestration here, thin state binding in the hook.
- `hooks/use-palettes.js` — the state binding.
- `api/client.js` — adds the one missing verb, `deletePalette()`.

No PHP. The palette REST surface already supports everything this screen needs.

## The rule the whole ticket rests on

A palette's **structure** — which swatches exist, their labels, their grouping, their order — lives only on the library's `$default` palette. `Palettes_Controller::effective_view()` builds every palette's groups from that node; a non-default palette stores only its value deltas.

So every structural edit writes the **default** palette, built from a fresh read of the default's own view, no matter which palette is being edited. Only a color value writes to the edited palette, through the granular per-swatch endpoint. `writeDefaultPaletteFlow` is the single place that encodes this, and its jest suite asserts the saved payload carries the default view's values rather than the edited palette's — the failure mode it guards against is silent data loss.

## Test plan

- `bun run test` — the flows and helpers are covered directly, including the routing guard above, the row-first ordering of swatch deletion, and that creating a palette never activates it.
- No UI to exercise yet; the screen lands in the next PR.

## Screenshots

None — this PR is the data layer and adds no UI. The screen itself lands in [2/5].

### DIFF
--- a/src/style-library/helpers/palette-flows.js
+++ b/src/style-library/helpers/palette-flows.js
@@ -1,0 +1,687 @@
+/**
+ * Pure orchestration for the Color Palette screen's write flows: open, activate, create, delete,
+ * the settings-panel save (rename + recolor), swatch removal, add color, add color group, and
+ * within-group reorder. Extracted out of `hooks/use-palettes` so each flow can be exercised
+ * directly in tests without rendering a component — a flow takes the REST calls it needs
+ * (imported here, so a test mocks `api/client`) plus a small set of injected callbacks for the
+ * state a caller reacts to (busy, a scoped error slot, and a `reload`/`refreshFeed` pair the hook
+ * provides). Every flow settles pessimistically and re-throws on failure so its caller (a modal,
+ * or the settings panel) can tell success from failure; none reloads the page.
+ *
+ * `writeDefaultPaletteFlow` is the single place the default-vs-edited write routing (a palette's
+ * structure lives only on its `$default` node — see the module's own docblock below) is encoded.
+ * Every structural flow in this file funnels through it; only `saveSwatchEditsFlow`'s color half
+ * ever writes the palette being edited.
+ *
+ * The central distinction here mirrors `library-flows.js`: the palette a site *renders with*
+ * (`$current`) versus the palette the app is *editing*. Opening is free and reversible — it only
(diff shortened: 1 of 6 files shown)


src/style-library/helpers/palette-flows.js
```diff
@@ -0,0 +1,702 @@
...
+		})
+		.then(() => reload())
+		.then(() => refreshFeed(slug))
+		.then(() => onBusy(false))
+		.catch((err) => {
+			onError({ message: errorMessage(err) });
+			onBusy(false);
+			throw err;
+		});
+}
```
d4mation, line 116: This doesn't happen until #1221, but it is easiest to highlight here:

When called from `reorderSwatchesFlow()`, it goes against what the comment in `reorderSwatches()` claims. 

> // Applied locally at drop time — a drop that snaps back behind a spinner is worse UX than
> // a rare rollback. `reorderSwatchesFlow`'s `reload()` re-reads the effective view on both
> // success (a no-op visually) and failure (restores server order), so this optimistic step
> // never needs its own undo path.

`reload()` happens _before_ the `catch()`, so if the `savePalette()` is rejected then the `reload()` is skipped.

d4mation (commented):
I'm not done with this stack. But leaving this comment for now.

src/style-library/helpers/palette-flows.js
pauloiankoski: Addressed on https://github.com/stellarwp/kadence-blocks/pull/1218/commits/94c9c0309874ebdcdb69b40cf6dc403a20940022

d4mation (approved):

PR #1220, opened 2026-08-05
[DTM] SOFT-4076 [3/5]: Settings panel contract and swatch panel

pauloiankoski:
Establishes the per-screen settings-panel contract, and ships the swatch panel that proves it. **This is the PR the remaining six screens will copy.**

## The contract

A screen exposes its panel as a static: `MyScreen.SettingsPanel = MyScreenSettings`. `StyleLibraryApp` reads it off the resolved component and mounts it into `AppShell`'s slot when the static exists and `route.item` is non-empty, passing the identical `{ route, navigate, library }` the screen received. No context provider, no new plumbing.

The screen and its panel are **siblings**, not parent and child. Everything they must agree on is therefore either server state or route state.

## `kb-scope`, and why it exists

The route gains a third, generic arg alongside `kb-screen` and `kb-item`:

```
?kb-screen=<id>   which screen
?kb-scope=<id>    the screen's sub-selection
?kb-item=<id>     the item open in the panel
```

`helpers/route.js` stays screen-agnostic — Color Palette reads `scope` as the palette being edited; another screen can read it as whatever its own sub-selection is.

This is load-bearing, not convenience. Because the screen and the panel are separate mounts of `usePalettes`, a per-instance `editingId` could not stay in sync: a freshly mounted panel had no way to learn which palette the screen had open and fell back to the active one, so editing a non-active palette showed the wrong color and a save would have written to the wrong palette. Deriving `editingId` from the route makes the two instances agree by construction. `hooks/use-libraries.js` documents the same principle for its own `editingSlug` — "a second copy... could drift".

## Also here

- The dev-only field-library demo moves onto `PlaceholderScreen` as its own static, still gated on `NODE_ENV` plus `?kb-gallery=1`, so `StyleLibraryApp` carries no screen-specific knowledge.
- `useSettingsPanel` gains a readiness contract: pass `null` while the caller's data is loading. It seeds on the item changing, or on the first arrival of real values for an already-open item, but never re-seeds afterward — an in-flight edit is never clobbered by an unrelated refresh.
- `ColorGradientPicker` gains an additive `colorOnly` option, defaulting off. The Color Palette panel uses it because a gradient the picker accepted *would* persist (a function-shaped string passes color validation) and then project as a color. Gradients are SOFT-4114.

## Test plan

- `bun run test`.
- In wp-admin: open a swatch, edit its name and color, save. Deep-link a `?kb-item=` URL and confirm the panel seeds. Open a non-active palette, select a swatch, and confirm the panel shows that palette's value.

## Screenshots

**Swatch settings panel** (note the picker has no Gradient tab — that is the `colorOnly` option)
pr3-settings-panel

**The color picker** — saturation area, hue and alpha sliders, and the format-cycling field row. The single Color tab is the `colorOnly` option; the Gradient tab returns with SOFT-4114 — replace this line with `pr3-color-picker.png`
pr3-color-picker


src/style-library/helpers/color-picker.js
```diff
@@ -0,0 +1,218 @@
...
+ */
+function percentOr(value, fallback) {
+	if (value === undefined) {
+		return fallback;
+	}
+
+	const number = Number(String(value).replace('%', ''));
+
+	return Number.isNaN(number) ? fallback : number / 100;
+}
```
d4mation, line 151: `react-color`'s `EditableInput` fires `onChange` with `''` when the field is cleared.

This means that the `if (value === undefined)` check won't return true and `Number( '' )` results in `0` rather than `NaN` so that additional attempt to return the fallback also never happens.

I figure that instead we should want this to reset to the fallback/default, right?

src/style-library/hooks/use-settings-panel.js
```diff
@@ -3,25 +3,85 @@
...
+ * @param {Object}  draft         The panel's current draft values.
+ * @param {?Object} initialValues The caller's persisted values, or null while still loading.
+ *
+ * @since TBD
+ *
+ * @return {boolean} True when `draft` differs from `initialValues`.
+ */
+export function computeIsDirty(draft, initialValues) {
+	return !isEqual(draft, initialValues || {});
+}
```
d4mation, line 74: What happens if the Hex `#rgb(a)` value is uppercase instead of the expected lowercase? What if `draft` gets converted to `rgba()` format and `initialValues` is still Hex? This could result in some weird issues.

src/style-library/helpers/color-picker.js
pauloiankoski: Addressed on 1691eb0372cc71e5bf50cffee6c4a7a1462942c6

src/style-library/hooks/use-settings-panel.js
pauloiankoski: Addressed on a247202d4 with `colord` package support.

d4mation (approved):

PR #1221, opened 2026-08-05
[DTM] SOFT-4076 [4/5]: Palette and swatch mutations

pauloiankoski:
Every write the Color Palette screen offers, bound to the already-tested flows.

## What's here

- **Add color** — mints a user color primitive, then appends it as a swatch.
- **Add Color Group** — the group is created *with* its first swatch, because the server drops an empty group even on the default palette.
- **Create Color Palette** — from the dropdown's trailing action. Creates and **opens**; never activates.
- **Rename palette** — the id stays fixed and only the label changes. The id is referenced by the `$current` pointer and by `data-kb-palette` attributes already rendered onto blocks, so re-deriving it from the new label would orphan them.
- **Set as active** — `ActivatePaletteButton` + `ActivatePaletteModal`, plus an Active badge in the dropdown. The only action that writes `$current`, which re-tints resolved colors across the whole site. The button renders nothing when the palette being edited is already active.
- **Delete palette** — targets the palette being **edited**, not the active one. Hidden for `$default`, which the server refuses to delete. Afterwards the listing is re-read so the fallback is whatever the server resolved.
- **Swatch delete** — removes the row from the default palette first, then best-effort cleanup of a user-created token, so a cleanup failure never reverses the removal.
- **Persisted reordering** — within a group.
- `ScreenHeader` gains a `secondaryAction` slot, and its lead row now holds the action slots at their intrinsic width while the title truncates, since Color Palette can fill it with five items at once.

## Note on reordering

Order is part of a palette's structure, so it is shared across every palette, exactly like the set of swatches, their labels, and their grouping. Only color values are per-palette. See `.local/palette-shared-structure-note.md`.

## Test plan

- `bun run test`.
- In wp-admin, verified: create does not activate; activate moves `$current` and the badge; a color added while editing one palette appears on all of them; a new group survives with its first swatch; deleting the active palette falls back to the server-resolved `$current`; Delete is absent (not disabled) on the default palette.
- Still worth a manual pass: drag-reorder persistence across a reload.

## Screenshots

**Header on a non-active palette** — all five lead items at once, which is also what the lead-row hardening is for
pr4-header-actions

**Activate confirmation**
pr4-activate-modal

### DIFF
--- a/src/style-library/components/organisms/AddColorGroupModal.js
+++ b/src/style-library/components/organisms/AddColorGroupModal.js
@@ -1,0 +1,86 @@
+/**
+ * The "Add Color Group" modal: a name field and Cancel / Add actions. Same skeleton as
+ * `CreatePaletteModal` — they do not merge, because their validation, copy, and flow
+ * (`addGroupFlow`) differ: this one mints the group's first swatch in the same write, since the
+ * server drops a color group with zero swatches even on the default palette.
+ */
+
+/**
+ * WordPress dependencies
+ */
+import { Button, Modal, Notice, TextControl } from '@wordpress/components';
+import { useState } from '@wordpress/element';
+import { __, sprintf } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { slugifyPaletteLabel } from '../../helpers/palettes';
+import './AddColorGroupModal.scss';
+
+/**
+ * Render the add-color-group modal.
+ *
+ * @param {Object}             props         The component props.
+ * @param {Object}             props.palette The palette being edited's effective view, for the duplicate-group check.
+ * @param {boolean}            props.isBusy  Whether the add request is in flight.
+ * @param {?{message: string}} props.error   The current structure error, if any.
+ * @param {Function}           props.onClose Called to dismiss the modal.
+ * @param {Function}           props.onAdd   Called with the typed label to create the group.
+ *
+ * @since TBD
+ *
+ * @return {JSX.Element} The modal.
+ */
+export function AddColorGroupModal({ palette, isBusy, error, onClose, onAdd }) {
+	const [label, setLabel] = useState('');
+	const groupId = slugifyPaletteLabel(label);
+	const isDuplicate = groupId !== '' && (palette?.groups ?? []).some((group) => group.id === groupId);
+
+	return (
+		<Modal
+			title={__('Add Color Group', 'kadence-blocks')}
+			className="kadence-blocks-style-library__add-color-group-modal"
+			onRequestClose={onClose}
+			// Locked while pending: no close icon, Escape does nothing, clicking outside does
+			// nothing. An add request in flight cannot be walked away from mid-request.
+			isDismissible={!isBusy}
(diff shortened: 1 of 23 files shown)


src/style-library/components/organisms/DeletePaletteModal.js
```diff
@@ -0,0 +1,81 @@
...
+			)}
+			<p>{__('Its color overrides are removed permanently.', 'kadence-blocks')}</p>
+			{isActive && (
+				<p>
+					{__(
+						'This is also your active palette, so your site will fall back to another one automatically.',
+						'kadence-blocks'
+					)}
+				</p>
+			)}
```
d4mation: This should be more specific, right? "another one" doesn't say which Palette would be used.

src/style-library/components/organisms/CreatePaletteModal.js
d4mation, line 1: After creating a Palette, the page gets stuck in a "loading" state.

image

d4mation, line 1: Also, after creating a Palette, the following modal shows for a split second. It may be related to the stuck loading state issue.

image

src/style-library/components/organisms/AddColorGroupModal.js
d4mation, line 1: This modal shows briefly after creating a Color Group. It doesn't seem to lock you into a "loading" state though.

image

src/style-library/components/organisms/CreatePaletteModal.js
pauloiankoski: 728ac0c4d

pauloiankoski: 14c358531

src/style-library/components/organisms/AddColorGroupModal.js
pauloiankoski: 14c358531

src/style-library/components/organisms/DeletePaletteModal.js
pauloiankoski: faad3e16b Implemented a solution just like the one I did for the Library selector.

src/style-library/components/organisms/AddColorGroupModal.js
d4mation: There's another issue now:

After creating a Palette, without refreshing the page, you get this issue if you try to immediately add a new Color:

```json
{
    "code": "rest_design_tokens_conflict",
    "message": "The token library was modified since you last read it. Reload and try again.",
    "data": {
        "status": 409,
        "slug": "default"
    }
}
```

pauloiankoski: Reproduced, and rename hits it too. `createPaletteFlow` and `renamePaletteFlow` were the only two writes that never called `refreshFeed`, so the page kept the version token it read before the write while the server had already bumped it — the next write 409s. Both refresh now (bbe0ac0).

The rename docblock argued a refresh was pointless because a label cannot reach a resolved value. True for the CSS, but the feed also carries the version token, which is what made it wrong; corrected there too.

d4mation (approved):

PR #1224, opened 2026-08-06
Fix: Navigation scroll spy throws TypeError on pages without matching anchors

Contributolo:
## Problem

When a Navigation block has scroll spy enabled and its menu links point to anchors on *other* pages (e.g. `/services/#design`, `/about/#team`), every page that does not contain those anchor targets throws:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'content')
    at l (gumshoe.min.js:7:1417)
    at m (gumshoe.min.js:7:1474)
    at g.detect (gumshoe.min.js:7:2490)
    requestAnimationFrame
    n @ gumshoe.min.js:7
```

It repeats on every animation frame for as long as the visitor stays at the bottom of the page.

This is easy to hit with a site-wide header: the menu is shared across all pages, but the anchor sections only exist on a few of them. Archive pages, blog posts and search results then all throw.

## Steps to reproduce

1. Add a Navigation block with **scroll spy enabled** and use it as a site-wide header.
2. Give the menu at least one cross-page anchor link, e.g. `/about/#team`.
3. Open a page that does **not** contain any of those anchor targets.
4. Scroll to the very bottom.

## Root cause

`initScrollSpy()` in `src/assets/js/kb-navigation-block.js` instantiates Gumshoe for every `[data-scroll-spy="1"]` block without checking whether any link resolves to an element on the current page.

Gumshoe's `setup()` only collects links whose target exists:

```js
const b = document.getElementById(decodeURIComponent(a.hash.substr(1)));
b && i.push({ nav: a, content: b });
```

So on such a page `contents` stays empty, and `getActive()` (bundled Gumshoe 5.1.2) has no guard for that:

```js
var getActive = function (contents, settings) {
	var last = contents[contents.length-1];      // undefined when contents is empty
	if (useLastItem(last, settings)) return last; // reads last.content -> TypeError
```

The reason it looks intermittent is `useLastItem`:

```js
var useLastItem = function (item, settings) {
	return !!(isAtDocumentBottom() && isInView(item.content, settings, true));
};
```

`item.content` is only evaluated once `isAtDocumentBottom()` returns `true`, so the error surfaces at the bottom of the page rather than on load.

Gumshoe 5.1.2 is the latest release and the upstream project is no longer maintained, so the guard has to live on the Kadence side.

## Fix

Resolve the selector once, and skip instantiation when none of the links point to an element present in the document. Pages that do have anchor targets are unaffected.

## Testing

Reproduced and verified with the real Navigation block markup in a standalone page, headless Chromium:

| Build | At page bottom, no anchor targets present |
|---|---|
| current `master` | `TypeError: Cannot read properties of undefined (reading 'content')` |
| this branch | no errors |

Regression check on a page that *does* contain the anchor targets, using the patched build: scrolling through the sections still applies `current-menu-item` to the matching entries (verified for three sections in sequence), no errors.

Both `src/assets/js/kb-navigation-block.js` and the built `includes/assets/js/kb-navigation-block.min.js` are updated, since the minified file is what the plugin enqueues.


Contributolo:
@rahulacaleffi one more small navigation fix, if you have a moment.

`initScrollSpy()` instantiates Gumshoe for every scroll-spy block regardless of whether any menu link resolves to an element on the current page. With cross-page anchors in a site-wide menu (`/about/#team` and the like), pages without those targets leave Gumshoe's `contents` array empty, and `getActive()` then reads `contents[-1]` and throws. It only surfaces at the bottom of the page, because `useLastItem` short-circuits on the at-bottom check first — which makes it look intermittent.

The change skips instantiation when nothing resolves; pages that do contain the anchors behave exactly as before. Gumshoe 5.1.2 is the final upstream release, so the guard has to live on this side.

PR #1225, opened 2026-08-06
[DTM] SOFT-4071: Token label overrides

pauloiankoski:
Lets a user rename any token — baseline or user-created — to a custom display label, stored per library as authoring metadata.

## What's here

- `Schema/Vocabulary/Extensions.php` — a `tokenLabels` section under `com.kadence.designTokens`.
- `Document/Token_Label_Index.php` — the pure read/write helper, mirroring `User_Primitive_Index`.
- `Schema/Validation/Dtcg_Validator.php` — an explicit validation branch for the new section.
- `Rest/V1/Documents_Controller.php` — `PUT`/`DELETE /documents/{slug}/labels/{id}`, version-conditional (409).
- `Admin/Feed/Feed_Assembler.php` + `Admin/Feed/Builder.php` — the effective label applied onto the admin feed, and a `labelOverridden` flag per entry.

`Resolver/Effective_Document::build()` strips `$extensions`, so an override can never reach projected CSS.

## Three decisions worth knowing

**The apply lives in `Feed_Assembler`, not in either emitter.** `Feed_Assembler::for_slug()` is the one pipeline both the page-load localizer and `Feed_Controller` call, and `Feed_ControllerTest` asserts those two payloads are identical. Applying in an emitter would give one of them overrides and not the other. It also can't live in `Token_Registry::to_ui_schema()` — the registry has no slug or store access and is library-independent by construction, so it cannot apply per-library data at all.

**The new section is deliberately absent from `Extensions::get_sections()`.** That list drives a preset-shaped tokens-map walk in the validator, and this section isn't preset-shaped. Anything outside the list passes through *unvalidated*, which is why `colorPalettes` already carries its own branch and why this one does too.

**The write path is narrowed on purpose** — the full DTCG validator and the resolver dry-run are skipped for a label write. Running them would make renaming impossible in any library whose stored document doesn't already validate, which is the trap that broke library rename before. Accepted consequence: a rename succeeds in an already-invalid library, since the write can't make it worse and refusing would hold authoring metadata hostage to unrelated corruption. The section is still validated whenever the full document is.

## Behavior notes

- An empty or whitespace-only label on `PUT` is a clear, exactly equivalent to `DELETE`, so a client needs no separate reset path. Storing an empty label is impossible by construction.
- A user-created primitive's label is written into the existing `userPrimitives` section rather than layered as a second store over the same id — one id never carries two competing labels.
- Clearing is safe for user-created ids: `Token_Definition::from_user_primitive()` falls back to a humanized last segment, so nothing is left nameless.
- The block-editor picker keeps declared labels. Its catalog emits one shared token list with only *values* per library, so per-library labels there would mean reshaping that payload per library.

## Test plan

`slic run wpunit Resources/Design_Tokens` — 1135 tests green, including a REST-level test proving an override actually surfaces through `Feed_Controller`, not merely that the two feed payloads agree with each other.

The Builder snapshot is regenerated in the final commit; the only change is each schema row gaining `labelOverridden`.

### DIFF
--- a/includes/resources/Design_Tokens/Document/Token_Label_Index.php
+++ b/includes/resources/Design_Tokens/Document/Token_Label_Index.php
@@ -19,6 +19,8 @@ final class Token_Label_Index {
 	 * All label overrides, keyed by canonical dot-path id. Read-side fail-soft: entries whose
 	 * key or value is not a non-empty string are dropped rather than surfaced, so a
 	 * hand-corrupted section degrades to "no override" instead of a type error downstream.
+	 * A whitespace-only label counts as empty — it would render as a nameless token, which is
+	 * the state an override exists to prevent.
 	 *
 	 * @since TBD
 	 *
@@ -36,7 +38,7 @@ final class Token_Label_Index {
 		$labels = [];
 
 		foreach ( $map as $id => $label ) {
-			if ( is_string( $id ) && $id !== '' && is_string( $label ) && $label !== '' ) {
+			if ( is_string( $id ) && $id !== '' && is_string( $label ) && trim( $label ) !== '' ) {
 				$labels[ $id ] = $label;
 			}
 		}
@@ -72,7 +74,8 @@ final class Token_Label_Index {
 	 * Store an override. An empty label is refused at the API level — storing an empty label
 	 * is impossible by construction, not merely validated against: clearing is remove(), never
 	 * set( '' ). The controller trims and routes '' to the clear path before this is ever
-	 * called; the throw guards future callers.
+	 * called; the throw guards future callers. Whitespace-only is empty here too, matching the
+	 * read side — an entry all() drops must not be storable in the first place.
 	 *
 	 * @since TBD
 	 *
@@ -85,7 +88,7 @@ final class Token_Label_Index {
 	 * @return array<string, mixed>
 	 */
 	public function set( array $document, string $id, string $label ): array {
-		if ( $label === '' ) {
+		if ( trim( $label ) === '' ) {
 			throw new InvalidArgumentException( 'A token label override cannot be empty; use remove() to clear it.' );
 		}
 
--- a/includes/resources/Design_Tokens/Document/Token_Label_Index.php
+++ b/includes/resources/Design_Tokens/Document/Token_Label_Index.php
@@ -1,0 +1,206 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Document;
+
+use InvalidArgumentException;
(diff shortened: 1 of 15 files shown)


includes/resources/Design_Tokens/Document/Token_Label_Index.php
```diff
@@ -0,0 +1,206 @@
...
+		$map = $this->read_map( $document );
+
+		if ( ! is_array( $map ) ) {
+			return [];
+		}
+
+		$labels = [];
+
+		foreach ( $map as $id => $label ) {
+			if ( is_string( $id ) && $id !== '' && is_string( $label ) && $label !== '' ) {
```
d4mation: We should probably use `trim( $label ) !== ''`. Even though the REST API Controller wouldn't allow for it, it will prevent future issues if new code were to allow saving something like `' '`.

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
```diff
@@ -123,6 +125,52 @@ final class Documents_Controller extends Controller {
...
+	private const ID_PARAM = 'id';
+
+	/**
+	 * The id path segment for the labels sub-route. Same grammar as PATH_ROUTE — a token id is a dot-path.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const ID_ROUTE = '(?P<' . self::ID_PARAM . '>[\w.-]+)';
```
d4mation: If this is specific to the Labels sub-route, it should probably be named as such. `LABELS_ID_PARAM` or similar.

includes/resources/Design_Tokens/Document/Token_Label_Index.php
pauloiankoski: Fixed in 5cd2521. Not reachable today since the controller trims and routes empty to the clear path, but worth hardening. `all()` now drops a whitespace-only label, and `set()` throws on one too: otherwise a future caller could store `'  '`, `has()` would say yes and `all()` would drop it — an override that silently does not exist. Covered in both tests.

includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
pauloiankoski: Fixed in 5cd2521. Renamed to `LABELS_ID_PARAM` / `LABELS_ID_ROUTE` — both are only used by the labels sub-route (route registration, `set_label`, `delete_label`, the DELETE args), and every neighbor is already scoped that way (`PATH_*`, `TITLE_*`, `SLUG_*`). The value stays `'id'`, so the capture name, route and args are unchanged. The other controllers keep their `ID_PARAM`, where the id is the resource itself.

d4mation (approved):

PR #1226, opened 2026-08-06
[DTM] SOFT-4073: Token sort order

pauloiankoski:
User-defined ordering of registered tokens, stored per library and applied where the Style Library lists them.

Builds directly on the label-override work in the PR below it — same extensions namespace, same index-helper shape, same validator pattern, same version-conditional route style, and the same apply seam.

## What's here

- `Schema/Vocabulary/Extensions.php` — a `tokenOrder` section under `com.kadence.designTokens`.
- `Document/Token_Order_Index.php` — the pure read/write helper.
- `Schema/Validation/Dtcg_Validator.php` — an explicit validation branch.
- `Rest/V1/Documents_Controller.php` — `PUT`/`DELETE /documents/{slug}/order/{group}`, reusing the existing version guard and metadata-persist path rather than duplicating them.
- `Admin/Feed/Feed_Assembler.php` + `Admin/Feed/Builder.php` — the order merged onto the admin feed.

## Storage shape

Group-keyed ordered id arrays, not a flat `{ id => int }` map. The UI schema already buckets tokens by group and a drag never crosses a group boundary, so ordered arrays make ties and out-of-range integers impossible by construction — a flat integer map would have to answer "what happens on a tie" and "what does 7 mean in a group of 4".

## The merge rules — the real risk in this change

Stored order is partial and advisory, never authoritative membership:

- Ids the map doesn't mention still appear, appended after the ordered ids in declaration order.
- Ids it mentions that no longer resolve are ignored on read and pruned on save.
- **Reordering can never hide a token.** The merged result is always the full registered set, only permuted.
- Removing the stored order restores declaration order.

Each has direct coverage. The never-hides-a-token rule is asserted across four fixtures — partial, stale, duplicate and empty — comparing id sets with `assertEqualsCanonicalizing`.

## The Builder snapshot is deliberately unchanged

An empty-order merge is the identity, so the snapshot fixture — which exercises `build()` with no stored order — comes out byte-identical. **A snapshot failure here is a bug in the merge, never a prompt to regenerate.** That property is what guarantees this feature can't disturb a library nobody has reordered.

## Scope

Feed layer only. `Editor/Pickable_Tokens_Catalog` is untouched, so the block-editor picker and the size/spacing dropdowns keep declaration order — a user-arranged scale reading 8px, 4px, 16px in a dropdown other people pick from would be a bad outcome, and that catalog emits one shared token list anyway, so per-library ordering there would multiply its payload by the number of libraries.

Palette order and preset order are out of scope — both are already ordered document data with their own endpoints.

## The order route: multi-word group labels

The route's `{group}` segment carries a `__()`-translated UI-schema label (`Font Size`, `Border Radius`, `Media Padding`), not a slug. Two things about that had to be fixed here.

The route regex was `[\w-]+`, which cannot match a space, so every multi-word group 404'd at WordPress's routing layer before reaching the handler. That is now `[^/]+` — only the path separator has to be excluded, since the group is a single path segment.

Widening the regex made the route *match*, but the value arriving at the handler was still percent-encoded: a browser must send `Border%20Radius`, and `WP_REST_Server` does not decode path segments before populating URL params, so the group lookup compared `"Border%20Radius"` against the real labels and missed. Every multi-word group returned `rest_design_tokens_unknown_group`; single-word groups worked, which is what made it look like the route was fine. The group param now goes through a `sanitize_callback` that applies `rawurldecode` once, shared by the PUT and DELETE routes. `rawurldecode` rather than `urldecode`, because a `+` in a path segment is a literal plus, not a space.

Both were caught only by driving the real UI — the original dispatch test built its route with a literal, unencoded space, which is a request no client can send, so it passed against broken code. It now encodes the segment with `rawurlencode`, and was confirmed to fail (404 against an expected 200) before the decode was applied.

## Test plan

`slic run wpunit Resources/Design_Tokens` — 1182 tests green. Snapshot suite green **without regeneration**.

The order route is additionally exercised through real `WP_REST_Server` dispatch with a percent-encoded multi-word group, which is the case every unit-level test missed.

### DIFF
--- a/includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
@@ -193,11 +193,20 @@ final class Documents_Controller extends Controller {
 	/**
 	 * The group path segment for the order sub-route.
 	 *
+	 * Unlike the token dot-path routes, this segment carries a human-readable, `__()`-translated
+	 * UI-schema group label (e.g. "Font Size"), not a slug — so it cannot be restricted to
+	 * `\w`/`-`. WordPress decodes the request path before matching route regexes (see
+	 * `WP::parse_request()`), so by match time the segment is the literal, fully-decoded label:
+	 * spaces, punctuation, and non-ASCII characters all appear as themselves, not percent-escapes.
+	 * The only character that must stay excluded is the path separator, since the group is a
+	 * single URL path segment; anything else is accepted here and validated against the real
+	 * declared groups downstream in `unknown_group()`.
+	 *
 	 * @since TBD
 	 *
 	 * @var string
 	 */
-	private const GROUP_ROUTE = '(?P<' . self::GROUP_PARAM . '>[\w-]+)';
+	private const GROUP_ROUTE = '(?P<' . self::GROUP_PARAM . '>[^/]+)';
 
 	/**
 	 * The request parameter that carries the ordered list of token ids on an order write.
@@ -2027,7 +2036,9 @@ final class Documents_Controller extends Controller {
 					'description' => __( 'The UI-schema group name whose token order is being set.', 'kadence-blocks' ),
 					'type'        => 'string',
 					'required'    => true,
-					'pattern'     => '^[\w-]+$',
+					// Matches self::GROUP_ROUTE: a translated group label, not a slug — only the
+					// path separator is excluded. See the GROUP_ROUTE docblock for why.
+					'pattern'     => '^[^/]+$',
 				],
 				self::VERSION_PARAM => [
 					'description'       => __( 'The version the client last read; empty for a first write. A mismatch is rejected with HTTP 409.', 'kadence-blocks' ),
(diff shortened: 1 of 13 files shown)


includes/resources/Design_Tokens/Rest/V1/Documents_Controller.php
```diff
@@ -1806,6 +2021,62 @@ private function get_label_delete_params(): array {
...
+	private function get_order_params(): array {
+		return array_merge(
+			$this->get_order_delete_params(),
+			[
+				self::ORDER_PARAM => [
+					'description'          => __( 'The token ids in their new order for this group.', 'kadence-blocks' ),
+					'type'                 => 'array',
+					'required'             => true,
+					'items'                => [ 'type' => 'string' ],
+					'additionalProperties' => false,
```
d4mation: `additionalProperties` won't do anything here I don't think?

pauloiankoski: It does nothing here. WP only reads `additionalProperties` in the object validate/sanitize path; for an array it looks at `items`, `minItems`, `maxItems` and `uniqueItems`. It came in with the route itself (aedc9b0), copy-pasted from the object-typed args above it. Removed in caf7146 — no `uniqueItems` needed either, since the handler already runs `array_unique` + `array_intersect` against the group ids.

d4mation (approved):

PR #1231, opened 2026-08-06
[DTM] SOFT-4080: Spacing screen

pauloiankoski:
Builds the **Spacing** screen of the Style Library. Targets the Border Width branch; the four scale screens were implemented as one linear chain.

The thinnest of the four — the spacing scale is the only one already declared, so this is a config, a preview renderer, a screen registration, and a single line of PHP.

## What it adds

- `components/pages/SpacingScreen.js` / `.scss` — the config and a filled preview square sized to the token's value.
- `declarations.php` — adds `group_key: 'spacing'` to the **existing** spacing declaration block. No new declarations, no `baseline.json` edit, and no change to any declared value or slug.

That last point is deliberate: each declared spacing step drives a `--global-kb-spacing-<slug>` custom property that existing block content already consumes, so this screen edits values real content depends on. Nothing about the shipped scale changes here — only its editability.

## The preview

Rendered at true CSS size, with the row growing to contain it, so the scale reads as a scale. It is clamped at 96px, which is where the mock's largest render sits: 3XL, 4XL and 5XL pin at the same size rather than stretching the list, and a pathological custom value cannot blow up the page.

`ListRow` frames its preview slot at a fixed 5rem and documents that slot as per-screen overridable. The shared `ScaleScreen` root carries a `scale-screen--<config.id>` modifier class for exactly this, and the override is scoped on it rather than leaking into the shared row chrome.

## Verification

`phpstan` clean with no baseline changes; `phpcs`, `cspell` and `eslint` clean; the Design_Tokens suite green (1258 tests); the CSS BC snapshot suite green with no diffs, which is the assurance that the declaration change did not disturb existing output; 193 jest tests; build succeeds.

Checked in the browser: previews render at true size with rows growing, and the clamp holds — 3XL, 4XL and 5XL pin identically while XXL at 5rem is visibly smaller.

## Screenshots

**The screen**
screen

### DIFF
--- a/src/style-library/components/pages/SpacingScreen.scss
+++ b/src/style-library/components/pages/SpacingScreen.scss
@@ -1,0 +1,25 @@
+// The value-sized square preview's own box. Unlike the other scale screens, this preview does not
+// fill the shared `.list-row-preview` slot — it is sized to the token's own value (set inline, per
+// row) so the mock's "true CSS size" behavior reads correctly, hence the slot override just below.
+.kadence-blocks-style-library__spacing-preview {
+	display: block;
+	background: var(--kb-sl-color-gray-100);
+	border: var(--kb-sl-border-width-input) solid var(--kb-sl-color-gray-700);
+	// The largest declared steps (3XL/4XL/5XL, 104px-160px) would otherwise blow the row past the
+	// mock's own tallest row (96px); the preview is an at-a-glance visual, not a measurement
+	// instrument, so it pins here and the value column disambiguates the top three rows.
+	max-width: 6rem;
+	max-height: 6rem;
+}
+
+// Scoped to this screen via `ScaleScreen`'s per-screen root modifier class (see
+// `.local/style-library-reference.md`) — Border Radius and Border Width keep the default framed
+// 80px slot from `ListRow.scss`; only Spacing needs a bare, value-sized square with the row
+// growing to contain it.
+.kadence-blocks-style-library__scale-screen--spacing .kadence-blocks-style-library__list-row-preview {
+	width: auto;
+	height: auto;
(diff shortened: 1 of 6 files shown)


src/style-library/components/pages/SpacingScreen.scss
```diff
@@ -0,0 +1,27 @@
...
+	// `border-box`, not the `<div>` default `content-box` — otherwise the border below adds on top
+	// of the inline width/height instead of drawing inside it, so the rendered box is 2px larger
+	// than the token's value on every step. Same fix as `.border-width-preview`.
+	box-sizing: border-box;
+	display: block;
+	background: var(--kb-sl-color-gray-100);
+	border: var(--kb-sl-border-width-input) solid var(--kb-sl-color-gray-700);
+	// Unclamped on purpose: the preview must show a token's true CSS size, not a normalized
+	// approximation, so every step (including 3XL/4XL/5XL) renders at its real value and the row
+	// grows to fit it. Accepted consequence: a very large custom value produces a very tall row.
```
d4mation, line 14: The docblock in `SpacingScreen.js` for `renderSpacingPreview` claims it is clamped. We should ensure that things match up properly to avoid confusion.

pauloiankoski: The docblock was the stale one. The clamp was dropped in 3eb5bb8 ("Render Spacing scale-screen previews at true CSS size"), which updated the SCSS comment but not the JS. Confirmed on the Spacing screen that 3XL (6.5rem) and 4XL (8rem) render past the old 6rem and the rows grow. Docblock corrected in 462ad53.

d4mation (approved):

PR #1236, opened 2026-08-07
[DTM] SOFT-4082: Shadow screen

pauloiankoski:
Builds the **Shadow** screen of the Style Library — the last of the BASE STYLES scale screens, and the first consumer of the `shadow` composite backend.

## The design was not missing

This ticket previously said no Shadow board could be found and told whoever picked it up to confirm with design first. That was wrong. The board is on the canvas at node [`481:22041`](https://www.figma.com/design/IEheoiM2vqbyTevD64ifXa/Kadence-Blocks-Sidebar?node-id=481-22041) — but it is *named* "Border Radius" in Figma, which is why a search by board name missed it. Worth renaming it there so nobody repeats the search.

## What it adds

- `components/pages/ShadowScreen.js` / `.scss` — the config, the preview renderer, and the screen/panel wrappers.
- `helpers/shadow.js` — pure parse/serialize/render helpers, jest-covered.
- `declarations.php` + `baseline.json` — `primitive.shadow.{xs,sm,md}` under a `Shadow` group with `group_key: 'shadow'`.

`ShadowField` is reused **unchanged**: the scaffold work had already built the Color row, the X/Y/Blur/Spread inputs and the Inset toggle, and it was already registered as a `shadow` field type, so a schema can declare it directly.

## It rides the shared contract after all

The Border Radius plan explicitly excluded Shadow from the `ScaleScreen` contract, on the grounds that it has six fields and no value column. On inspection both assumptions were softer than they looked — `SettingsForm` already renders a `shadow` field type, and `ListRow` already suppresses a falsy value column. So this reverses that exclusion rather than forking the screen.

The contract gains exactly two optional, identity-defaulted seams: `parseValue` (resolved value → draft) and `buildLeaf` (draft → token leaf). The four existing screens pass neither and are behaviorally unchanged. A bespoke screen would have duplicated the write flows, the serialized reorder, the stale-item self-heal and the draft-channel wiring to avoid two function keys.

## The number-versus-string boundary

`ShadowField` emits bare numbers for X/Y/Blur/Spread; the composite stores dimension strings (`"0px"`, `"2px"`). The conversion lives at the screen boundary in `helpers/shadow.js`, not inside `ShadowField` — changing the field would affect every other consumer of it. Values are px-only, matching the board, which shows no unit selector.

Confirmed against the running app: a save writes `{ offsetX: "0px", offsetY: "1px", blur: "2px", spread: "0px", inset: true }` — strings, not numbers — and `inset` is omitted entirely when false.

## Two decisions worth reviewing

**The seeded colors are literal hex, not palette aliases.** The panel seeds by parsing the *resolved* CSS value, so any alias in a shadow primitive would be silently rewritten to a literal the first time someone saves that row. Literal seeds make that a non-event. The same reasoning is why **`semantic.shadow.card` and `semantic.shadow.media` are left untouched** — `card` aliases `primitive.color.neutral.900`, and re-pointing it at the new scale would detach it from the palette.

The consequence, which design should know about: a shadow color set through this screen can only ever be a literal. Pointing a shadow at a palette token is not possible here.

**The preview slot needs `overflow: visible`.** A shadow is drawn outside its box, and the shared row slot clips with `overflow: hidden` — left alone, every shadow on the screen would be invisible and the previews would look like empty squares. The override is scoped to this screen through the per-screen root modifier.

## Verification

`phpstan` clean with no baseline changes; `phpcs`, `cspell` and `eslint` clean; the full `Resources/Design_Tokens` suite green (1264 tests); snapshot suites green with no diffs; 224 jest tests including 11 new cases for the shadow helpers.

Driven in the browser: the three rows render with visibly increasing shadows and no clipping; the panel matches the board field for field and seeds correctly from the stored value; **live preview works on the composite** (editing Blur moved that row to `0px 2px 28px 0px` while its siblings kept their stored values); the unsaved-changes guard and Discard behave as they do on the scalar screens; Inset renders live and round-trips through a save.

Not exercised by hand: drag-reorder on this screen, the "+ Add Shadow" mint flow, and the Color popover's portal styling.

## Not included

Stacked shadows — a DTCG `$value` that is an array of shadow objects stays unsupported in v1, and a shadow token remains a single object. If the design ever wants layered shadows that is a separate backend change.

## Screenshots

**The screen with MD selected**
screen-with-panel

### DIFF
--- a/src/style-library/__tests__/scale-flows.test.js
+++ b/src/style-library/__tests__/scale-flows.test.js
@@ -221,6 +221,71 @@ describe('saveScaleTokenFlow', () => {
 		expect(onError).toHaveBeenCalledWith({ message: failure.message });
 		expect(onBusy.mock.calls).toEqual([[true], [false]]);
 	});
+
+	it('uses the supplied buildLeaf for the value write', async () => {
+		client.saveTokenLeaf.mockResolvedValue({});
+		const refreshFeed = jest.fn().mockResolvedValue({});
+		const buildLeaf = jest.fn().mockReturnValue({ $type: 'shadow', $value: { color: '#171717' } });
+		const draftValue = { color: '#171717', offsetX: 0, offsetY: 2, blur: 8, spread: 0, inset: false };
+
+		await saveScaleTokenFlow({
+			...baseArgs,
+			tokenType: 'shadow',
+			draft: { label: 'MD', value: draftValue },
+			initial: { label: 'MD', value: { ...draftValue, offsetY: 4 } },
+			refreshFeed,
+			onBusy: jest.fn(),
+			onError: jest.fn(),
+			buildLeaf,
+		});
+
+		expect(buildLeaf).toHaveBeenCalledWith('shadow', draftValue);
+		expect(client.saveTokenLeaf).toHaveBeenCalledWith(
+			'kb-design-tokens/v1',
+			baseArgs.tokenId,
+			{ $type: 'shadow', $value: { color: '#171717' } },
+			'default'
+		);
+	});
+
+	it('defaults to buildTokenLeaf when no buildLeaf is supplied', async () => {
+		client.saveTokenLeaf.mockResolvedValue({});
+		const refreshFeed = jest.fn().mockResolvedValue({});
+
+		await saveScaleTokenFlow({
+			...baseArgs,
+			draft: { label: 'SM', value: '0.25rem' },
+			initial: { label: 'SM', value: '0.125rem' },
+			refreshFeed,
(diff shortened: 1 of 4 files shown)


pauloiankoski:
This PR was **not** intentionally merged. When SOFT-4082 was reordered to sit below SOFT-4150, its head became an ancestor of its base, so the diff went empty and GitHub marked it merged. No code landed anywhere as a result.

The Shadow screen still needs review: it now lives in **#1249**, with the same commits, targeting `SOFT-4081/icon-sizes-screen`. Please review there.

PR #1237, opened 2026-08-07
[DTM] SOFT-4077: Typography screen

pauloiankoski:
Builds the **Typography** screen of the Style Library — the size scale, rendered as cards with live sample text in the selected family.

This is the first of three stacked PRs for SOFT-4077. It is a complete, working screen on its own; the font **catalog** and **Add/Delete Font** arrive in the two that follow, because they need a backend change that deserves its own review.

## A rename that had to happen first

`primitive.fontFamily.{sans,serif,mono}` existed in `baseline.json` but **could never have been registered**. `Token_Definition::from_array()` — the declaration path — enforces the same `^[a-z0-9]+([.-][a-z0-9]+)*$` id charset as the user-primitive path and throws on a camelCase segment, because the id feeds `Css_Var::from_id()`. So those tokens could never reach the UI schema, and the font selector would have had nothing to list.

This was a pre-existing violation of the repo's own kebab-case rule, invisible only because nothing had declared them yet. Renamed to `primitive.font-family.*`, atomically with the `semantic.font-family.control` alias that points at it and the wpunit fixtures that reference it. The semantic layer was already correctly kebab-case, so the inconsistency was confined to the primitive layer.

Note the DTCG `$type` stays `fontFamily` — that vocabulary is separate from token *names*, exactly as the conventions doc says.

## What the screen does

- Font **tabs** list the families available in the design system and select which one the previews render in. This is a preview selector.
- Cards show a label chip, a size chip, and the sample text at that size — the sizes are the already-declared `primitive.dimension.font-size.*` group.
- Reorder, rename, live preview and the unsaved-changes guard all come from the shared scaffold.

## It rides the shared contract

One new optional seam: `renderToolbar`, which renders the font row between the header and the list and takes over the add action, because the board puts "+ Add Size" there rather than in the header. Identity-defaulted, so the other five screens are untouched — verified in the browser that Border Radius still renders its add button inside the header.

## Font sizes are fluid

Every `primitive.dimension.font-size.*` entry carries a `clamp` extension, so the resolved value is a full `clamp(...)` string and the steps project to `--global-kb-font-size-<slug>` site-wide. The card chips show the clamp **max** via a pure helper. An explicit edit converts that step to a fixed value — preserving a stale clamp would swallow the edit.

## Verification

`phpstan` clean; `phpcs`, `cspell` and `eslint` clean; `Resources/Design_Tokens` green (1269 tests); **all three snapshot suites green with zero diffs** — those fixtures carry no font tokens, so any diff here would have been a real regression rather than an expected change; 231 jest tests.

Checked in the browser: the screen renders, the tabs switch, and every sample re-renders in the selected family.

## Not included

The font **catalog** dropdown, **Add Font** and **Delete Font** — they need the `$type`→id-segment mapping, which is the next PR in this stack.

`primitive.fontSize.*` is left alone: a second, undeclared, camelCase font-size scale referenced only from preset maps inside `$extensions`. It is unregistered and inert. Recorded here as known tech debt rather than widened into this change.

## Screenshots

**The screen with the font tabs and size cards**
a-typography-screen

### DIFF
--- a/src/style-library/components/pages/TypographyScreen.js
+++ b/src/style-library/components/pages/TypographyScreen.js
@@ -1,7 +1,7 @@
 /**
  * The Typography screen: the scale config, the FONT preview toolbar, the sample-text preview
  * renderer, and the two thin wrappers that plug into the shared `ScaleScreen`/`ScaleSettings`
- * contract (see `.local/style-library-reference.md`). Two things make this screen genuinely
+ * contract (see `ScaleScreen.js`'s module docblock). Two things make this screen genuinely
  * different from its siblings: a toolbar between the header and the list (the FONT dropdown, font
  * tabs, and the "+ Add Size" action) built on `ScaleScreen`'s optional `renderToolbar` seam, and
  * screen-level preview state (the currently previewed font) that `renderPreview` closes over —
@@ -132,8 +132,8 @@ function typographyToolbarRenderer(fonts, selectedFontId, onSelectFont) {
 }
 
 /**
- * The Typography screen's config — see `ScaleScreen`'s module docblock and
- * `.local/style-library-reference.md` for the full per-screen config contract. `formatValue`/
+ * The Typography screen's config — see `ScaleScreen`'s module docblock for the full per-screen
+ * config contract. `formatValue`/
  * `parseValue` both go through `fontSizeDisplayValue()`: the size chip and the SIZE field must both
  * seed from the authored scalar (a clamp's `max`), never the resolved `clamp(...)` string a
  * fluid step actually carries.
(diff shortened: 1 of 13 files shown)


src/style-library/components/pages/TypographyScreen.js
```diff
@@ -0,0 +1,227 @@
...
+		if (fonts.length > 0 && !fonts.some((font) => font.id === selectedFontId)) {
+			setSelectedFontId(fonts[0].id);
+		}
+	}, [fonts, selectedFontId]);
+
+	const selectedFont = fonts.find((font) => font.id === selectedFontId) ?? null;
+
+	// `useScaleScreen`'s memo deps read scalar config fields, not the config object's identity (see
+	// `.local/style-library-reference.md`), so rebuilding this object every render is safe as long as
+	// the field values it carries stay stable across renders that shouldn't change anything.
```
d4mation: Leaving comments to your `.local/` files which don't exist in the Repo won't be helpful in the future 😅 

But also: are we sure about what is stated in this comment?

This seems to be true for `baseRows` and `initialValuesFor` as they read only `config.group` and `config.parseValue`, but other usages like `addToken`, `saveToken`, and `reorderTokens` (defined in other PRs, but the comment itself stays mostly the same) take in the whole `config` object.

If this comment is taken at face-value, another dev may try to add `useMemo` later and then be confused that it ends up firing every render.

src/style-library/components/pages/TypographyScreen.js
d4mation, line 1: There are a handful of `.local/` file paths in this file. Make sure all are removed and we instead include whatever context is necessary directly in the docblock please :)

src/style-library/components/pages/TypographyScreen.js
pauloiankoski: The comment was wrong. `baseRows` and `initialValuesFor` read scalar fields, but `addToken`, `saveToken` and `reorderTokens` list the whole `config` object in their deps, so an inline object would rebuild those three every render. Rewritten in b3ea964 to say that, and to say why it is only wasted work (nothing reads them from an effect) rather than a bug.

src/style-library/components/pages/TypographyScreen.js
pauloiankoski: Removed here and across the stack — 21 links in 12 files (`ScaleScreen`, `ScaleSettings`, each screen and its SCSS, `use-draft-channel`), each fixed on the PR that introduces the file. The per-screen config contract is now documented in `ScaleScreen.js`'s module docblock (PR #1229), which is what the other screens point at; everything else was inlined where it was needed.

d4mation (approved):

PR #1239, opened 2026-08-07
[DTM] SOFT-4077: Font catalog with Add and Delete

pauloiankoski:
Adds the font **catalog** to the Typography screen: a searchable dropdown over every available family, with a contextual **Add Font** / **Delete** button. Third of three stacked PRs for SOFT-4077, and the one that makes the previous two useful.

## The model

Two controls, doing different jobs — worth stating because an earlier reading had them duplicating each other:

- **Tabs** list the families in the design system and choose which one the size previews render in.
- **The dropdown is a catalog** of everything available, not a mirror of the tabs. Its button is contextual: pick a family that is not in the design system → **+ Add Font**; pick a custom one that is → **Delete**; pick a baseline family → the button is disabled, because baseline tokens are never deletable anywhere in this app.

## What it adds

- `Admin/Feed/Font_Catalog` + a second page-load-only inline global. It is **not** part of `Feed_Assembler`'s payload, which re-fetches on every write — a ~29KB list must not ride every save.
- `SearchableSelectDropdown` — the app had no searchable select, and the catalog is **1,916 families**, so a plain select was unusable. Built new per the app's independence rule.
- `use-google-font-loader` — nothing loaded font files on this page before, so a preview of "Inter" silently rendered as `system-ui`. Families in the Google catalog now get a stylesheet injected once each; system and custom families are skipped.
- The Add/Delete flows over the user-primitive endpoints.

## A minted family is a single-family stack

`$value` is `[ "Abril Fatface" ]` — the family alone, with **no invented generic fallback**. `gfonts-array.php` carries no category field (verified across all 1,916 entries), so there is no reliable way to know whether a family is serif, sans or display. Guessing `sans-serif` for a display face would be worse than omitting it.

## Worth a reviewer's attention

**External requests from wp-admin.** The loader injects `<link>` tags to `fonts.googleapis.com` so previews render truthfully. That is a real privacy/egress decision — some sites forbid it. There is no local-hosting toggle here; the plan flags it as an open question rather than assuming.

## Verification

`phpstan` clean; `phpcs`, `cspell`, `eslint` clean; `Resources/Design_Tokens` green (1305 tests); snapshot suites green with zero diffs; 258 jest tests.

Driven end to end in the browser: the catalog loads 1,916 families; search narrows correctly; picking a family not in the system enables **+ Add Font**; adding mints `primitive.font-family.custom.abril-fatface` with `$type: fontFamily` and `$value: ["Abril Fatface"]` — which is the `$type`→segment mapping from the previous PR working in anger; the new tab appears and activates; the previews render in the real downloaded face (`document.fonts.check` confirms it, not just eyeballing); the button flips to **Delete**; deleting removes the token and the tab and restores the previous family.

## Not included

A local-hosting or opt-out path for Google Fonts. Font **weights** and styles — a family is added at its default weight only. Also unchanged: the `Body` step and the px-based scale the boards show, which are open design questions on the ticket.

## Screenshots

**The catalog dropdown, searching**
c-catalog-search

**After Add Font — new tab active, button flipped to Delete, previews in the real face**
c-font-added

### DIFF
--- a/includes/resources/Design_Tokens/Admin/Feed/Font_Catalog.php
+++ b/includes/resources/Design_Tokens/Admin/Feed/Font_Catalog.php
@@ -84,6 +84,18 @@ final class Font_Catalog {
 	 * @return string[]
 	 */
 	private function custom_names(): array {
+		/**
+		 * Filters the site-registered custom fonts, the same list the block editor localizes as
+		 * c_fonts. Each entry is keyed by the font's family name or by a whole font-stack
+		 * expression (`"My Font", sans-serif`, the shape kadence_blocks_convert_custom_fonts()
+		 * builds when a fallback is registered); an integer-keyed string entry is a plain name.
+		 *
+		 * @since TBD
+		 *
+		 * @param array<string|int, mixed> $fonts The registered custom fonts.
+		 *
+		 * @return array<string|int, mixed> The registered custom fonts.
+		 */
 		$fonts = apply_filters( self::CUSTOM_FONTS_FILTER, [] );
 
 		if ( ! is_array( $fonts ) ) {
@@ -93,13 +105,37 @@ final class Font_Catalog {
 		$names = [];
 
 		foreach ( $fonts as $key => $value ) {
-			if ( is_string( $key ) ) {
-				$names[] = $key;
-			} elseif ( is_string( $value ) ) {
-				$names[] = $value;
+			$name = is_string( $key ) ? $this->family_of( $key ) : ( is_string( $value ) ? $this->family_of( $value ) : '' );
+
+			if ( $name !== '' ) {
+				$names[] = $name;
 			}
 		}
 
 		return $names;
+	}
+
+	/**
+	 * The family name a catalog entry names, taken from a font-stack expression when it carries
+	 * one: the first family, with its surrounding quotes removed.
+	 *
+	 * A registered fallback puts the whole stack in the key (`"My Font", sans-serif`). Catalogued
+	 * verbatim, that string becomes the picked font's `$value`, which Css_Renderer quotes as one
+	 * family — `font-family: "\"My Font\", sans-serif"`, invalid — and never matches its own
+	 * Google entry when the two lists are deduplicated. The family alone is what both need. The
+	 * fallback is deliberately not carried into the token: a font primitive is minted as a
+	 * single-family stack, the same rule that keeps a generic fallback from being invented for a
+	 * Google family.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $entry The catalog entry, a family name or a font-stack expression.
+	 *
+	 * @return string The family name, or "" when the entry names none.
+	 */
+	private function family_of( string $entry ): string {
+		$first = explode( ',', $entry )[0];
+
+		return trim( $first, " \t\n\r\0\x0B\"'" );
 	}
 }
--- a/includes/resources/Design_Tokens/Admin/Feed/Font_Catalog.php
+++ b/includes/resources/Design_Tokens/Admin/Feed/Font_Catalog.php
@@ -1,0 +1,105 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Admin\Feed;
+
+/**
+ * Reads and normalizes the font catalog the Typography screen's searchable dropdown lists:
+ * every Google font family name plus every site-registered custom font name.
+ *
+ * The Google names come from includes/gfonts-names-array.php, the same generated, flat name
+ * list the block editor already localizes (class-kadence-blocks-editor-assets.php's g_font_names).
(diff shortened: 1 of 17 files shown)


includes/resources/Design_Tokens/Admin/Feed/Font_Catalog.php
```diff
@@ -0,0 +1,105 @@
...
+
+		$names = [];
+
+		foreach ( $fonts as $key => $value ) {
+			if ( is_string( $key ) ) {
+				$names[] = $key;
+			} elseif ( is_string( $value ) ) {
+				$names[] = $value;
+			}
+		}
```
d4mation, line 113: This allows for that weird quoted font family name with a fallback issue to happen. 

Compare this to `kadence_blocks_convert_custom_fonts()` which will use something like `"My Font", sans-serif` which becomes the array key, and `custom_names()` uses that key as the font name.

Clicking "Add Font" will set something like `$value: [ '"My Font", sans-serif' ]` which `Css_Renderer::font_family()` will convert to `font-family: "\"My Font\", sans-serif"` which is invalid.

`Font_CatalogTest` only covers a case without the fallback, so CI wasn't able to catch this.

includes/resources/Design_Tokens/Admin/Feed/Font_Catalog.php
```diff
@@ -0,0 +1,105 @@
...
+
+	/**
+	 * The site's custom font names, normalized from the custom-fonts filter's associative shape.
+	 *
+	 * @since TBD
+	 *
+	 * @return string[]
+	 */
+	private function custom_names(): array {
+		$fonts = apply_filters( self::CUSTOM_FONTS_FILTER, [] );
```
d4mation, line 99: This should have a phpdoc describing the filter since it is new. Description, `@since TBD`, `@param`s, and `@return`.

src/style-library/components/molecules/SearchableSelectDropdown.js
```diff
@@ -0,0 +1,134 @@
...
+						</span>
+						<span className="kadence-blocks-style-library__searchable-select-dropdown-icon">
+							<Icon
+								className="kadence-blocks-style-library__searchable-select-dropdown-chevron"
+								icon={chevronDown}
+							/>
+						</span>
+					</Button>
+				)}
+				renderContent={({ onClose }) => (
```
d4mation, line 84: It looks like if there's no results, there isn't a "No results"-type message but it is just blank instead, right? Maybe we should add one to make it super clear for users.

src/style-library/hooks/use-google-font-loader.js
```diff
@@ -0,0 +1,66 @@
...
+		if (!google.includes(familyName)) {
+			return;
+		}
+
+		loadedFamilies.add(familyName);
+
+		const link = document.createElement('link');
+		link.rel = 'stylesheet';
+		link.href = `https://fonts.googleapis.com/css2?family=${encodeURIComponent(familyName)}&display=swap`;
+		document.head.appendChild(link);
```
d4mation, line 64: You mentioned this in the PR description, but just noting again for when this is merged that you should record a Loom explaining this for Donaldo that we can watch when he's back so he can decide what we may want to do with this.

includes/resources/Design_Tokens/Admin/Feed/Font_Catalog.php
pauloiankoski: Reproduced with a test first — a filter entry keyed `"My Font", sans-serif` was catalogued verbatim, so the pick wrote that whole stack as one family. Fixed in a42b523: the catalog now takes the first family and strips its quotes, so `My Font` is what is listed and stored. That also fixes the Google dedupe, which never matched a stack key. The registered fallback is deliberately not carried into the token — a font primitive stays a single-family stack, same rule as not inventing a generic for a Google family. Two tests added (fallback key, and a fallback key that must dedupe away).

includes/resources/Design_Tokens/Admin/Feed/Font_Catalog.php
pauloiankoski: Added in a42b523: description, `@since TBD`, `@param` and `@return`, documenting both key shapes (family name or whole stack expression).

src/style-library/components/molecules/SearchableSelectDropdown.js
pauloiankoski: Confirmed on the Typography screen — a query with no matches left the menu blank, just the search box. Added a "No fonts found" row in a42b523, styled like the truncation footer but with no top border since it replaces the list rather than sitting under it.

c-catalog-no-results

src/style-library/hooks/use-google-font-loader.js
pauloiankoski: Noted — I will record the Loom on the Google Fonts loading approach and share it so Donaldo can pick it up when he is back.

d4mation (approved):

PR #1240, opened 2026-08-07
[DTM] SOFT-4075: Style Library shared components & screen scaffold (complete)

pauloiankoski:
**Consolidation PR for SOFT-4075.** This is the complete ticket as one unit: the Style Library app shell, navigation and screen registry, library selector, shared content primitives, the settings panel and field library, and explicit library activation.

It exists so `feature/design-tokens-module` receives the whole ticket rather than six partial slices.

## Constituent PRs

Reviewed individually, in order:

| PR | Phase |
|---|---|
| #1210 | App shell and routing |
| #1211 | Navigation and screen registry |
| #1212 | Library selector |
| #1213 | Shared content primitives |
| #1214 | Settings panel and field library |
| #1216 | Explicit library activation, rename, safer delete |

Plus the two commits on `SOFT-4075/base` that strip work-plan phase references and ticket numbers from shipped comments.

## How to use this

Review the phase PRs above — they carry the per-phase reasoning and screenshots. This PR is the merge unit: it is what lands on the feature branch, and it contains exactly the same commits.

Nothing here is new work; there is no commit in this PR that is not in one of the PRs listed above.


pauloiankoski:

PR #1242, opened 2026-08-07
[DTM] SOFT-4077: Typography screen (complete)

pauloiankoski:
**Consolidation PR for SOFT-4077.** The complete Typography ticket as one unit: the screen, the backend mapping that makes user-created font families possible, and the font catalog with Add and Delete.

## Constituent PRs

| PR | Part |
|---|---|
| #1237 | Typography screen, and the `primitive.fontFamily` → `primitive.font-family` rename that made the families registrable at all |
| #1238 | Mapping a DTCG `$type` to a registered kebab id segment (backend only) |
| #1239 | The font catalog, contextual Add / Delete, and the Google Fonts loader |

These were split deliberately: #1238 flips an invariant shipped days earlier and widens the set of creatable types from three to nine, which deserves review on its own terms rather than buried behind a font picker. #1239 carries an external-request decision (injecting Google Fonts stylesheets from wp-admin) that would be easy to miss in a large diff.

## How to use this

Review the three PRs above — that is where the reasoning and the screenshots live, and #1238 is the one to scrutinise. This PR is the merge unit so the feature branch receives Typography as a whole, and contains no commit that is not in one of them.

## Still open on this ticket

Design questions the boards raise and the implementation does not yet answer: the step names and px values versus the shipped rem scale, the `Body` chip placement, and whether the Google Fonts loader needs a local-hosting option.

### DIFF
--- a/includes/resources/Design_Tokens/Admin/Feed/Font_Catalog.php
+++ b/includes/resources/Design_Tokens/Admin/Feed/Font_Catalog.php
@@ -1,0 +1,105 @@
+<?php declare( strict_types=1 );
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Admin\Feed;
+
+/**
+ * Reads and normalizes the font catalog the Typography screen's searchable dropdown lists:
+ * every Google font family name plus every site-registered custom font name.
+ *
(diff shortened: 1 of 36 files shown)


pauloiankoski:

PR #1243, opened 2026-08-07
[DTM] SOFT-3906: Blocks-side token seam + singlebtn proving ground (complete)

pauloiankoski:
**Consolidation PR for SOFT-3906.** The complete ticket as one unit: the integration base, the blocks-side token seam registration with the `singlebtn` proving ground, and per-corner preset values with the save-as-preset fix.

## Constituent PRs

| PR | Part |
|---|---|
| #1201 | Integration base for the proving ground |
| #1173 | Blocks-side token seam registration + `singlebtn` proving ground |
| #1202 | Per-corner preset values and the save-as-preset 422 |

## How to use this

**Keep reviewing the PRs above** — #1173 in particular already has review history, and this PR is not a place to restart that conversation. This exists only so `feature/design-tokens-module` receives SOFT-3906 as a whole rather than three partial slices. It contains no commit that is not already in one of them.

## Note on the stack

#1203 (SOFT-4085, responsive preset values) is a separate ticket branched from `SOFT-3906/preset-per-corner-values`. It is left targeting that branch, since retargeting it is not required for this consolidation to be valid and would disturb an unrelated ticket's review.


pauloiankoski:

PR #1259, opened 2026-08-12
Fix TypeError in header sticky cleanup branch (parent used before assignment)

### LINKED ISSUE #1258: Header block: TypeError on every scroll when sticky is disabled for the active screen size ((void 0).classList in shipped bundle)
### Environment

- Kadence Blocks 3.7.9 — also verified in the 3.7.9.1 release bundle from wordpress.org
- WordPress 7.0.3, block-based header (Advanced Header block)
- Header configured with sticky enabled for desktop only (tablet/mobile sticky off)

### Steps to reproduce

1. Build a header with the Advanced Header block and enable **Sticky** for desktop only.
2. Open the site in a tablet or mobile viewport (any breakpoint where sticky is disabled).
3. Scroll.

### Actual result

Every scroll event throws:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'classList')
    at KBHeader.updateSticky (.../includes/assets/js/kb-header-block.min.js:1:7925)
```

### Analysis

The shipped bundle literally contains an unconditional crash in the `updateSticky` cleanup branch (the path taken when sticky is *not* enabled for the active size):

```js
...this.stickyWrapper.style.position="initial",
(void 0).classList.remove("child-is-fixed"),
void document.body.classList.remove("header-is-fixed");
```

`(void 0).classList` throws every time this branch executes, which suggests a reference was statically resolved to `undefined` at build time (the expression sits where an element reference — presumably the inner header element that owns `child-is-fixed` — should be).

Verifiable directly in the current release file:
https://plugins.svn.wordpress.org/kadence-blocks/tags/3.7.9.1/includes/assets/js/kb-header-block.min.js (search for `child-is-fixed`).

### Impact

- Console error on **every scroll event** for all visitors on breakpoints where sticky is off (easily hundreds of errors per session; also inflates client-side error monitoring).
- Because the statement chain aborts at the throw, `document.body.classList.remove("header-is-fixed")` never runs, so the `header-is-fixed` body class can incorrectly stick around after resizing/breakpoint changes.

### Workaround

Enabling sticky for all breakpoints (desktop, tablet, mobile) avoids the broken branch entirely.

Contributolo:
Fixes #1258

### Problem

In `updateSticky()`, the early-return cleanup branch (taken whenever sticky is not enabled for the active screen size) references `parent` before the `var parent = this.stickyWrapper.parentNode;` assignment further down in the method:

- `src/assets/js/kb-header-block.js` line 488: `parent.classList.remove('child-is-fixed');` (use)
- line 550: `var parent = this.stickyWrapper.parentNode;` (assignment)

Because of `var` hoisting, `parent` is `undefined` at line 488. The bundler statically resolves this and ships `(void 0).classList.remove("child-is-fixed")` in `includes/assets/js/kb-header-block.min.js`, so the branch throws unconditionally:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'classList')
    at KBHeader.updateSticky
```

On any site where sticky is enabled for some breakpoints only (e.g. desktop-only), every scroll event on the other breakpoints throws, and `document.body.classList.remove('header-is-fixed')` on the following line never runs, so the body class can incorrectly persist across breakpoint changes.

### Fix

Use `this.stickyWrapper.parentNode` directly in the cleanup branch — the method's opening guard (`if (!this.stickyWrapper) return;`) guarantees it exists, and it is exactly the element the later `var parent` assignment refers to. The second `parent.classList.remove('child-is-fixed')` (line 726) runs after the assignment and stays unchanged.

Note: I have only changed the source file; the minified bundle under `includes/assets/js/` is not rebuilt in this PR.

### DIFF
--- a/src/assets/js/kb-header-block.js
+++ b/src/assets/js/kb-header-block.js
@@ -485,7 +485,7 @@ class KBHeader {
 			this.stickyWrapper.style.height = null;
 			this.stickyWrapper.style.top = null;
 			this.stickyWrapper.style.position = 'initial';
-			parent.classList.remove('child-is-fixed');
+			this.stickyWrapper.parentNode.classList.remove('child-is-fixed');
 			document.body.classList.remove('header-is-fixed');
 			return;
 		}


Contributolo:
@rahulacaleffi since you have this part of the codebase fresh from #968 — would you mind glancing at this one as well? It is a one-line change.

In `updateSticky()` the early-return cleanup branch uses `parent` before its `var` assignment further down, so `var` hoisting leaves it `undefined`. The bundler resolves that statically and ships `(void 0).classList.remove('child-is-fixed')`, meaning the branch throws on every scroll whenever sticky is off for the active screen size — verified in the 3.7.9.1 bundle from wordpress.org.

Repro steps and the shipped-bundle evidence are in #1258.

Contributolo:
Quick update after 3.7.10: the release bundle on wordpress.org still ships `(void 0).classList.remove("child-is-fixed")` in `includes/assets/js/kb-header-block.min.js` (verified against `tags/3.7.10`), so the scroll-event TypeError from #1258 is unchanged for sites with sticky enabled on some breakpoints only. The branch still merges cleanly against master. Could this one-liner be picked up for the next release?

PR #1269, opened 2026-08-17
[DTM] SOFT-4083 [2/23]: Preset helpers and fetch hook

pauloiankoski:
Pure preset payload/row/slug helpers, plus the `useButtonPresets` fetch-and-bind hook. Nothing renders them yet — this is the data layer the screen above consumes.

## Test instructions

1. Run `npx wp-scripts test-unit-js src/style-library/__tests__/presets.test.js`.
2. Read `presetInitialValues` and the slug helpers against the payload shape from the endpoint in the slice below — the contract between them is the thing worth reviewing here.
3. There is no UI to click at this slice.

<details>

```bash
git checkout SOFT-4083/phase-1b-preset-helpers-hook
bun run build-wp
```

</details>

---

Slice 2 of 23 in the SOFT-4083 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

* **New Features**
  * Added support for loading and displaying button presets from the style library.
  * Added preset data handling, including token values, editable drafts, saved settings, and automatic preset naming.
  * Added loading and error states for preset retrieval.

* **Tests**
  * Added comprehensive coverage for preset conversion, resolution, mapping, validation, serialization, and slug generation.

### DIFF
--- a/src/style-library/helpers/presets.js
+++ b/src/style-library/helpers/presets.js
@@ -1,0 +1,201 @@
+/**
+ * Pure preset payload/row/slug helpers for the Button preset screen (and, unchanged, any preset
+ * screen that reuses this contract): mapping a preset GET payload to row view models, the
+ * alias<->id codec the preset write surface needs (presets store `{dot.path}` aliases, not bare
+ * ids), resolving a stored token value (alias or literal) against the feed's resolved value map,
+ * seeding a settings-panel draft, and minting the next free preset slug. No React, no JSX, no
+ * REST — see `hooks/use-button-presets.js` for the state binding and `api/client.js` for the REST
+ * wrappers this module's output feeds.
+ */
+
+/**
+ * Internal dependencies
+ */
+import { nextScaleSlug } from './scale';
+
+/**
+ * The frozen bound surface a button preset defines, in the order the panel and previews read it.
+ * The single JS spelling of the properties `declarations.php` binds for the button block — every
+ * seed, save, and preview walk goes through this list, so this app can never write a property the
+ * server's `guard_surface` would reject as unbound.
+ *
+ * @since TBD
+ */
+export const BUTTON_PRESET_PROPERTIES = [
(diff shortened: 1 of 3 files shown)


src/style-library/helpers/presets.js
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * @since TBD
+ */
+export const BUTTON_PRESET_PROPERTIES = [
+	'button-bg',
+	'button-text',
+	'button-bg-hover',
+	'button-text-hover',
+	'button-radius',
+];
```
d4mation: Aren't we able to get this from the preset bindings in `declarations.php`? Each of these is a key there. If we can avoid a duplicate list it would be ideal. It should already be localized using `Preset_Catalog::properties_for()`.

d4mation (approved):

src/style-library/hooks/use-button-presets.js
```diff
@@ -0,0 +1,90 @@
...
+
+		return () => {
+			cancelled = true;
+		};
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [namespace, block, slug, version]);
+
+	const rows = useMemo(() => presetRows(payload, library?.values), [payload, library?.values]);
+
+	const initialValuesFor = (presetSlug) => presetInitialValues(payload, presetSlug);
```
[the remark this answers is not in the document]
pauloiankoski: Stale — phase-9 already does this. `presetInitialValues` takes `properties` as an argument now and the hook is `usePresets(library, preset)`. Nothing calls it with another block at this slice.

src/style-library/helpers/presets.js
pauloiankoski: Fixed in https://github.com/stellarwp/kadence-blocks/pull/1269/commits/1f37ca53403f58deeee14b36fab52b9c41d31324

PR #1271, opened 2026-08-17
[DTM] SOFT-4083 [4/23]: Token color select field

pauloiankoski:
Adds the `token-color-select` field type and narrows `pickableTokensForType` by an optional token role, so a field only offers tokens from its own role.

## Test instructions

1. Run `npx wp-scripts test-unit-js src/style-library/__tests__/tokens.test.js`.
2. The narrowing is the part to review: confirm passing a role filters the pool, and passing none leaves it unchanged, so existing callers are unaffected.

<details>

```bash
git checkout SOFT-4083/phase-3a-token-color-select
bun run build-wp
```

</details>

---

Slice 4 of 23 in the SOFT-4083 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

* **New Features**
  * Added a color token selector with swatches, resolved values, selected states, and read-only support.
  * Added role-based filtering for selectable design tokens.
  * Added reusable button settings panels for Normal and Hover states.
  * Added token color selection to the demo Color panel.

* **Tests**
  * Added coverage for color resolution, button settings, token filtering, and active-library values.

### DIFF
--- a/src/style-library/components/molecules/fields/TokenColorSelectField.js
+++ b/src/style-library/components/molecules/fields/TokenColorSelectField.js
@@ -1,0 +1,111 @@
+/**
+ * A round-swatch token color picker (e.g. Button's Text / Background rows): a bordered row whose
+ * toggle is the field's own name beside a swatch of the currently selected token's resolved color,
+ * opening a menu of every pickable color token — each option showing its own swatch, label, and
+ * resolved value. Closes the gap between `ColorListField` (writes a raw CSS string, no token
+ * reference) and `TokenSelectField` (writes a token id but renders no swatch): this field writes a
+ * token id AND shows its color.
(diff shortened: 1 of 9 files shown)


src/style-library/components/molecules/fields/TokenColorSelectField.js
d4mation, line 1: This looks a bit off compared to the other field replacements we've done.

I figure we need to have the "Style Library" and "Custom" tabs along with spacing things out nicely, correct?

image

d4mation (approved):
I'm fine with updating this in another PR. It shouldn't meaningfully impact later PRs in the Stack so it isn't worth dealing with rebasing this all over.

PR #1279, opened 2026-08-17
[DTM] SOFT-4083 [12/23]: Button token declarations and bindings

pauloiankoski:
PHP data layer only. Rebuilds the radius scale on Tailwind's steps and values, binds button padding and margin (valued only when a preset sets them), and documents a preset token value as possibly non-scalar.

This sits at the bottom of the remaining stack because the JS slices above bind to it.

## Test instructions

1. Run `vendor/bin/phpstan analyse` and `slic run wpunit Resources/Design_Tokens/Registry/Preset_BindingsTest`.
2. Go to **Style Library → Border Radius** and confirm the scale reads None 0, XS 0.125rem, SM 0.25rem, MD 0.375rem, LG 0.5rem, XL 0.75rem, 2XL 1rem, 3XL 1.5rem, Full 9999px.
3. Confirm a button with no preset spacing set renders **no** padding/margin declarations — the bindings are deliberately valued only when a preset sets them, so this slice should change nothing visually on the front end.
4. **If the site errors after checking this branch out**, the cached baseline is stale: flush it (`lando ssh -s redis -c 'redis-cli flushall'`, or whatever your object cache is). Editing `baseline.json` does not bust that cache on its own.

<details>

```bash
git checkout SOFT-4083/phase-8-button-token-declarations
bun run build-wp
```

</details>

---

Slice 12 of 23 in the SOFT-4083 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

* **New Features**
  * Added extra border-radius options, including `2xl` and `3xl`.
  * Added responsive Button controls for padding and margin across desktop, tablet, and mobile.
  * Expanded preset support for responsive values and per-corner settings.
* **Bug Fixes**
  * Corrected radius values for small and extra-large sizing options.
  * Improved recognition of padding-related settings as dimensional values.

### DIFF
--- a/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
+++ b/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
@@ -245,7 +245,7 @@
 				},
 				"sm": {
 					"$type": "dimension",
-					"$value": "0.1875rem"
+					"$value": "0.25rem"
 				},
 				"md": {
 					"$type": "dimension",
@@ -257,7 +257,15 @@
 				},
 				"xl": {
 					"$type": "dimension",
+					"$value": "0.75rem"
+				},
+				"2xl": {
+					"$type": "dimension",
 					"$value": "1rem"
+				},
+				"3xl": {
+					"$type": "dimension",
+					"$value": "1.5rem"
 				},
 				"full": {
 					"$type": "dimension",
@@ -486,7 +494,7 @@
 		"radius": {
 			"control": {
 				"$type": "dimension",
-				"$value": "{primitive.dimension.radius.sm}"
+				"$value": "0.1875rem"
 			},
 			"media": {
 				"$type": "dimension",
--- a/includes/resources/Design_Tokens/Registry/declarations.php
+++ b/includes/resources/Design_Tokens/Registry/declarations.php
@@ -63,6 +63,8 @@ $radius_labels = [
 	'md'   => __( 'Medium', 'kadence-blocks' ),
 	'lg'   => __( 'Large', 'kadence-blocks' ),
 	'xl'   => __( 'Extra Large', 'kadence-blocks' ),
+	'2xl'  => __( '2X Large', 'kadence-blocks' ),
+	'3xl'  => __( '3X Large', 'kadence-blocks' ),
 	'full' => __( 'Full', 'kadence-blocks' ),
 ];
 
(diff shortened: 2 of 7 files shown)


includes/resources/Design_Tokens/Registry/declarations.php
```diff
@@ -499,6 +501,27 @@ static function ( string $slug, string $label ): array {
...
+					],
+				],
+				'button-margin'     => [
+					'css_var'          => 'kb-btn-margin',
+					'control_attr'     => 'margin',
+					'responsive_attrs' => [
+						'tablet' => 'tabletMargin',
+						'mobile' => 'mobileMargin',
+					],
+				],
```
d4mation: While this is true, it looks like it tries to initialize an empty value when creating a new Preset which results in a validation failure:

```json
{
  "data": {
    "status": 422,
    "block": "kadence/singlebtn",
    "errors": [
      {
        "path": "$extensions.com.kadence.designTokens.presets.kadence/singlebtn.button.tokens.button-padding",
        "code": "value_invalid",
        "message": "A foundation-preset/block-preset token value must be an alias, a non-empty literal, a slot list, or a responsive entry."
      },
      {
        "path": "$extensions.com.kadence.designTokens.presets.kadence/singlebtn.button.tokens.button-margin",
        "code": "value_invalid",
        "message": "A foundation-preset/block-preset token value must be an alias, a non-empty literal, a slot list, or a responsive entry."
      }
    ]
  }
}
```

includes/resources/Design_Tokens/Registry/Baseline/baseline.json
```diff
@@ -245,7 +245,7 @@
 				},
 				"sm": {
 					"$type": "dimension",
-					"$value": "0.1875rem"
+					"$value": "0.25rem"
 				},
```
d4mation, line 249: Wasn't this intentionally `0.1875rem` (`3px`) so that the default styling for Button Border Radius was preserved?

pauloiankoski: Makes sense, semantics as role-based defaults, primitives as the selectable set. Padding and margin get the same treatment, landing in #1290 rather than reopening this slice.

includes/resources/Design_Tokens/Registry/Baseline/baseline.json
```diff
@@ -486,7 +494,7 @@
 		"radius": {
 			"control": {
 				"$type": "dimension",
-				"$value": "{primitive.dimension.radius.sm}"
+				"$value": "0.1875rem"
```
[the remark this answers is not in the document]
pauloiankoski: We are going to replace that again with a semantic token but not in this slice. That fix is landing in https://github.com/stellarwp/kadence-blocks/pull/1290 instead.

d4mation (approved):

PR #1280, opened 2026-08-17
[DTM] SOFT-4083 [13/23]: Generic preset screen and sidebar

pauloiankoski:
Extracts the Button-specific screen into a reusable preset screen and sidebar, and adds a `beforeTabs` slot to the settings panel template. Pure refactor — no behavior change.

## Test instructions

1. **Turn on rename detection before reading this diff.** Most of it is a move git did not detect, so without it the diff reads as far more new code than it is.
2. On **Block Presets → Button**, confirm the **Name field now sits above** the Normal/Hover tabs — that is the one visible consequence of the refactor.
3. Re-run the slice-4b checks (add, rename, save, delete) and confirm nothing regressed in the move.
4. Review question worth answering: the generic screen is meant to serve future block presets, so check the Button-specific bits really did move into `button-preset.js` rather than staying behind.

<details>

```bash
git checkout SOFT-4083/phase-9-generic-preset-screen
bun run build-wp
```

</details>

## Screenshot

Name field now above the Normal/Hover tabs, in the beforeTabs slot
phase-9-generic-preset-screen

---

Slice 13 of 23 in the SOFT-4083 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

* **New Features**
  * Added a reusable preset management experience for listing, creating, selecting, reordering, editing, saving, and deleting style presets.
  * Added support for button preset configuration, including normal and hover states, previews, colors, radius, and labels.
  * Added unsaved-change safeguards and clearer handling for loading, empty, stale, and error states.
  * Added documentation covering the Style Library architecture and preset workflows.

* **Improvements**
  * Refactored button management screens to use the shared preset experience.
  * Improved preset creation and preview behavior for future block-specific configurations.


src/style-library/presets/button-preset.js
```diff
@@ -0,0 +1,168 @@
...
+ * `guard_surface` rejects. The preset name is not here; it is tab-independent and comes from
+ * `presetNameSchema()`.
+ *
+ * @param {string} tab The active tab name (`'normal'` or `'hover'`).
+ *
+ * @since TBD
+ *
+ * @return {{panels: Array<Object>}} The settings-form schema for the active tab.
+ */
+function schemaFor(tab) {
```
d4mation, line 96: It looks like the old copy of this was left in `helpers/presets.js` as `buttonSettingsSchema(tab)`. I assume it should be deleted?

pauloiankoski: Deleted. phase-9 moved the schema into `button-preset.js` and left this behind. No consumers anywhere in the chain.

pauloiankoski:
All five addressed in `5c7de0c75`:

- **Spread before the fixed prop** — `<PresetScreen {...props} preset={BUTTON_PRESET} />`, so the binding can no longer be overridden.
- **`properties` getter at render scope** — dropped from both destructurings and read as `preset.properties` inside `addPreset` / `initialValuesFor`, so the throwing getter only fires where the value is used. This was the one with real behaviour behind it: the laziness the config documents was not actually holding.
- **Docblock** — documents the `preset` config and its three fields; the stale `useButtonPresets` reference is gone.
- **`createPresetFlow` hardcoding the Button label** — the flow now takes `newLabel`, supplied by the preset config and threaded through the hook.
- **Duplicate `BUTTON_BLOCK`** — one definition, in `helpers/presets.js`, re-exported from the config. Kept it there rather than in the config because `getButtonPresetProperties` needs it, and the other direction would make `helpers` import from `button-preset`, which already imports `helpers`.

Also fixed a separate save failure this branch inherits, found while testing: `button-padding` / `button-margin` became bound in the slice below, so every preset draft carried an empty entry for them and `presetSaveTokens` sent `""`. The server rejected the document and a label-only save failed with "The design token document failed validation". The guard that omits unset properties existed six slices higher, so #1279 through #1289 were all affected — it now lives in the slice that introduces the binding.

d4mation (approved):

PR #1284, opened 2026-08-17
[DTM] SOFT-4083 [17/23]: Box control and public barrel

pauloiankoski:
The first control built on the shell, plus the library's public barrel — the surface both hosts import from.

## Test instructions

1. Still unwired; the UI should be unchanged.
2. Review `src/token-controls/index.js` as an API decision: anything exported there becomes something the block editor and the Style Library may depend on, and anything omitted stays internal.

<details>

```bash
git checkout SOFT-4083/phase-10d-token-box-control
bun run build-wp
```

</details>

---

Slice 17 of 23 in the SOFT-4083 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

* **New Features**
  * Added controls for editing spacing and radius values as a single value or individually for each side.
  * Added linked and unlinked editing modes, custom numeric values, units, tokens, responsive breakpoints, reset options, and disabled states.
  * Added shared token-control components and utilities for consistent token selection, status indicators, and responsive value management.

### DIFF
--- a/src/token-controls/controls/BoxControl.js
+++ b/src/token-controls/controls/BoxControl.js
@@ -1,0 +1,153 @@
+/**
+ * The box-shaped token control: one value or four, linked or individual.
+ *
+ * Backs both the `radius` and `spacing` field types. Radius and spacing are the same component
+ * because they are the same value shape — four slots of one dimension. What differs is geometry
+ * (corners walk clockwise, sides do not), which token pool is offered, and whether negatives are
+ * allowed. Those are settings, not a reason to fork.
+ *
+ * **Linked state is controlled or uncontrolled, and the choice follows the host's storage.** Pass
+ * `onToggleLink` and the host owns it — the block editor does, because its attribute is always a
+ * four-element array and "linked" is a UI mode with nowhere to live in the value. Pass nothing and
+ * this derives it from the value's shape and collapses a uniform list back to a scalar on write,
+ * which is what the Style Library's stored preset shape wants (a scalar there reads as "every
+ * slot"). Mixing the two — a host that controls the toggle while values still collapse — would
+ * silently rewrite an array attribute into a string, so `collapse` defaults to the safe pairing.
+ *
+ * **The binding indicator and its reset are opt-in.** They exist for the block editor, where a
+ * control can override a selected preset. The Style Library passes no `status` and gets a bare
+ * label.
+ */
+
+/**
+ * Internal dependencies
+ */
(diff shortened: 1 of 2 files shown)


src/token-controls/controls/BoxControl.js
```diff
@@ -0,0 +1,153 @@
...
+						min={min}
+						max={max}
+						step={step}
+						// The field speaks three intents; a slot stores one value, so they collapse
+						// here rather than in every host. `onCustom` writes a bare number — the unit
+						// is the control's, shared across all four slots, exactly as the editor's
+						// measurement control treats it.
+						onPick={(alias) => !disabled && onSlotChange(alias)}
+						onClear={() => !disabled && onSlotChange('')}
+						onCustom={(next) => !disabled && onSlotChange(next)}
```
d4mation, line 150: `disabled` doesn't end up getting passed through to `<TokenSelector>` itself, so `<TokenPopover>` and `<TokenPickerButton>` technically aren't aware of that state except indirectly via these callbacks. 

This means that these elements are still interactive and writes to them are silently ignored. We should ensure they're correctly identified as disabled when `disabled === true`.

Maybe adding support for that in PR #1283 and then rebasing to pass it through here makes the most sense? Or you can do it here as this PR is pretty small. Whichever you prefer :)

pauloiankoski: Went with your first option, with one correction: `TokenSelector` and friends come from #1282, not #1283, so `disabled` landed there (`bcc847c48`) and this PR now passes it through (`f0b144f26`).

The prop only needs to reach the trigger button — it is the only interactive element outside the popover, so once it is inert the popover cannot open and nothing below it is reachable. Added two tests for it; #1282 had none before.

src/token-controls/controls/BoxControl.js
```diff
@@ -0,0 +1,153 @@
...
+	units,
+	onUnit,
+	defaultValue,
+	inherited = false,
+	slotIcons = null,
+	role = 'sides',
+	status = null,
+	onReset = null,
+	showReset = true,
+	indicator = null,
```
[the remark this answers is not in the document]
pauloiankoski: Documented. It was already destructured and forwarded to `ControlShell`, just missing from the table — which matters more here than usual, since `BoxControl` ships through the public barrel and that table is the contract for both hosts.

d4mation (approved):

PR #1285, opened 2026-08-17
[DTM] SOFT-4083 [18/23]: Editor adopts the token-control library

pauloiankoski:
The block editor extension switches to the shared barrel and the duplicated `component-token-ui` module is deleted (799 lines).

## Test instructions

1. Run `npx wp-scripts test-unit-js src/extension/design-tokens`.
2. In the block editor, add a **Kadence Single Button**, open **Style**, and confirm the token fields still behave exactly as before this slice: the trigger reads like an input, the popover opens with Style Library / Custom tabs, and Reset works.
3. **Behavior should be identical** — this is a swap, not a change. Any visible difference is a bug.
4. Confirm `component-token-ui.js` and its stylesheet are gone and nothing imports them.

<details>

```bash
git checkout SOFT-4083/phase-10e-editor-adopts-token-controls
bun run build-wp
```

</details>

---

Slice 18 of 23 in the SOFT-4083 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

* **Refactor**
  * Updated token-mapped component fields to use the shared token selector controls.
  * Preserved existing token selection, callbacks, and editor behavior.
  * Consolidated token-control styling and interactions for a more consistent editing experience.
  * Standardized responsive preset data handling to improve consistency across supported settings.


d4mation (approved):

pauloiankoski:
Fixed in `f2fb659d5` — the test imports through `../../../token-controls` now, the same path the production file uses.

Worth confirming it actually buys something rather than just matching style: with the change in place, deleting `export { TokenChip }` from the barrel fails `register-component-filters.test.js`. Before it, that deletion left the test green while the real import broke.

d4mation (approved):

PR #1288, opened 2026-08-17
[DTM] SOFT-4083 [21/23]: Editor canvas radius control

pauloiankoski:
Renders Border Radius in the block editor with the same shared box control the Style Library uses, so both hosts present the control identically.

## Test instructions

1. In the block editor, select a **Kadence Single Button** and open **Style**.
2. Border Radius now renders as the corner grid with token names and values, matching the Style Library.
3. Set a corner to a token, then to a custom value, and confirm both render on the canvas.
4. Compare side by side with the Style Library's radius field — presenting the same control in both hosts is the point of the slice.

<details>

```bash
git checkout SOFT-4083/phase-13-editor-radius-box-control
bun run build-wp
```

</details>

## Screenshot

Border Radius in the block editor, rendered by the shared box control
phase-13-editor-radius-box-control

---

Slice 21 of 23 in the SOFT-4083 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

* **New Features**
  * Added responsive border-radius controls with linked or individual corner editing.
  * Added token selection, inherited values, units, reset options, and device-specific settings.
  * Improved handling of inherited values across responsive breakpoints.

* **Bug Fixes**
  * Preserved per-device overrides when linking or changing border-radius values.

* **Tests**
  * Added coverage for responsive breakpoint switching and inherited corner values.

### DIFF
--- a/src/blocks/singlebtn/edit.js
+++ b/src/blocks/singlebtn/edit.js
@@ -86,6 +86,7 @@ import BackendStyles from './components/backend-styles';
 import { PresetButton } from '../../extension/preset-picker/PresetButton';
 import { usePresetBinding, resetAttr } from '../../extension/token-indicators';
 import {
+	anyCornerInherited,
 	deriveMeasureMode,
 	inheritedMeasureSlots,
 	measureAttrsForDevice,
@@ -1307,7 +1308,9 @@ export default function KadenceButtonEdit(props) {
 															onDeviceChange={setPreviewDevice}
 															tokens={borderRadiusTokens}
 															defaultValue={inheritedBorderRadius.values}
-															inherited={!!inheritedBorderRadius.inherited}
+															inherited={anyCornerInherited(
+																inheritedBorderRadius.inherited
+															)}
 															state={tokenBinding.borderRadius}
 															onReset={() => resetToken('borderRadius')}
 															isLinked={borderRadiusIsLinked}
--- a/src/blocks/singlebtn/edit.js
+++ b/src/blocks/singlebtn/edit.js
@@ -91,8 +91,9 @@ import {
 	measureAttrsForDevice,
 	presetValueForDevice,
 } from '../../extension/token-indicators/normalize';
-import { TokenLabel } from '../../extension/token-indicators/components/TokenLabel';
 import { TokenControlRow } from '../../extension/token-indicators/components/TokenControlRow';
+import { EditorBoxControl } from '../../extension/design-tokens/components/EditorBoxControl';
+import { pickableTokensForControl } from '../../extension/token-picker';
 
 export default function KadenceButtonEdit(props) {
 	const { attributes, setAttributes, isSelected, context, clientId, name } = props;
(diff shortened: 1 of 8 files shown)


src/extension/design-tokens/components/EditorBoxControl.js
```diff
@@ -0,0 +1,130 @@
...
+				// controls show, and two marks meaning the same thing should not look different.
+				indicator={<TokenIndicator state={state} onReset={onReset} />}
+				isLinked={isLinked}
+				onToggleLink={onToggleLink}
+				// Never collapse: the attribute is always a four-element array, so folding a uniform
+				// value down to a scalar would write a shape the block cannot read back.
+				collapse={false}
+				role={role}
+				breakpoints={Object.values(BREAKPOINT_FOR_DEVICE)}
+				breakpoint={breakpoint}
```
d4mation, line 129: This needs `onBreakpointChange` as well, correct?

src/blocks/singlebtn/edit.js
```diff
@@ -1217,145 +1297,30 @@ export default function KadenceButtonEdit(props) {
...
+															label={__('Border Radius', 'kadence-blocks')}
+															value={borderRadiusForDevice.value}
+															onChange={(next) =>
+																setAttributes({ [borderRadiusForDevice.attr]: next })
+															}
+															previewDevice={previewDevice}
+															onDeviceChange={setPreviewDevice}
+															tokens={borderRadiusTokens}
+															defaultValue={inheritedBorderRadius.values}
+															inherited={!!inheritedBorderRadius.inherited}
```
d4mation: This ends up always being `true` because `inheritedBorderRadius.inherited` is always a four element `boolean[]`. On Desktop, it should say `Default (...)` but this will make it say `Inherited (...)` instead. 

In current use, the `Default (...)` text doesn't actually show because the Presets are seeded with the `3px` border radius for Desktop, so you would need to create a new Preset to have it get seeded empty to see this. The issue mentioned in https://github.com/stellarwp/kadence-blocks/pull/1289#discussion_r3797641579 will need addressed before you can easily test that though.

src/extension/design-tokens/components/EditorBoxControl.js
pauloiankoski: Wired it up. ControlShell calls the prop, not the context, so the switcher was doing nothing. Pulled the breakpoint→device mapping out so it isn't duplicated.

src/blocks/singlebtn/edit.js
pauloiankoski: Fixed. Checks if any corner actually inherits now instead of coercing the array, so Desktop says Default again.

src/blocks/singlebtn/edit.js
```diff
@@ -317,6 +319,85 @@ export default function KadenceButtonEdit(props) {
...
+		? borderRadiusPresetValue
+		: inheritedBorderRadius.values;
+	const inheritedCornersDiffer =
+		Array.isArray(inheritedCorners) && inheritedCorners.some((corner) => corner !== inheritedCorners[0]);
+	// The preset surface carries resolved literals, not aliases (`blockPresetValues` flattens them so the
+	// overridden-check can be a plain compare), so the literal is matched back to the token that produced
+	// it. Writing the alias keeps the collapsed corner bound to `None` rather than landing as a custom
+	// `0px` that merely happens to look the same.
+	const aliasForValue = (value) => borderRadiusTokens.find((token) => token.value === value)?.alias ?? value ?? '';
+	const inheritedFirstCorner = Array.isArray(inheritedCorners) ? aliasForValue(inheritedCorners[0]) : '';
```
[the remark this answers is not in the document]
pauloiankoski: Fixed, but split into two things.

The narrow bug: `inheritedCorners` short-circuited to the raw preset array whenever the preset was per-corner, bypassing `inheritedBorderRadius.values` entirely — which already falls through to the preset correctly via `presetSlotAt`, but also correctly keeps a partially-stored breakpoint corner where one exists. Now always reads `inheritedBorderRadius.values`.

The unit-mismatch part: traced it end to end. The real frontend was never at risk — `render_measure_side()` (PHP) and the JS `KadenceBlocksCSS.render_measure_output()` both already resolve a `{alias}` corner to `var(--kb-token--…)`, tested in `alias-css-output.test.js`. What was actually broken is `singlebtn`’s own editor-canvas preview, which used a separate, older `getPreviewSize` + `render_size` pair with no alias awareness — a picked token would show `{alias}px` (invalid CSS) while editing, even though the saved page rendered fine. Swapped it to the same `render_measure_output()` call PHP already uses. Verified live: picking a token now emits `border-top-right-radius:var(--kb-token--…)` in the canvas, computed style matches the token exactly.

src/blocks/singlebtn/edit.js
```diff
@@ -1217,145 +1298,32 @@ export default function KadenceButtonEdit(props) {
...
+															}
+															previewDevice={previewDevice}
+															onDeviceChange={setPreviewDevice}
+															tokens={borderRadiusTokens}
+															defaultValue={inheritedBorderRadius.values}
+															inherited={anyCornerInherited(
+																inheritedBorderRadius.inherited
+															)}
+															state={tokenBinding.borderRadius}
+															onReset={() => resetToken('borderRadius')}
```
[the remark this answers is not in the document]
pauloiankoski: Checked both parts separately.

Reset: `resetAttrPatch("borderRadius", "dimension")` clears `borderRadius`, `tabletBorderRadius` and `mobileBorderRadius` together, by explicit design (the docblock: "so their clearing convention cannot drift"). It does not target desktop only — every dimension control in this codebase resets its whole responsive stack at once, not this slice's doing.

Indicator: this part is real. `usePresetBinding()` computes `overridden` from the base (desktop) attribute regardless of `previewDevice`, so `state={tokenBinding.borderRadius}` can show the wrong mark on Tablet/Mobile. But that's a property of `usePresetBinding` itself, shared by every dimension control it feeds — not something specific to this slice's Border Radius integration. Leaving it out here; worth a follow-up against `usePresetBinding` directly rather than a one-off patch on this control.

pauloiankoski:
The two nitpicks from the review body are also fixed: `changeBreakpoint` is now a single named handler shared by `BreakpointProvider` and `ControlShell`'s switcher, and `EditorBoxControl.test.js` covers the forward device -> breakpoint mapping plus the `collapse={false}` invariant, neither of which had a test before.

src/blocks/singlebtn/edit.js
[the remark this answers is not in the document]
pauloiankoski: Implemented in usePresetBinding() on this slice: added a previewDevice param, a dimension control now compares its own device attribute against the preset value at that device instead of always the desktop one. Propagated through phase-14.

d4mation (approved):

PR #1289, opened 2026-08-17
[DTM] SOFT-4083 [22/23]: Button preset padding and margin

pauloiankoski:
Adds the responsive Padding and Margin fields to the button preset and renders them on the block, only when the preset sets them. Binds to the declarations in the PHP slice at the bottom of this stack.

## Test instructions

1. Open a Button preset — **Spacing** now has Padding and Margin, both responsive, both empty.
2. Set a padding, save, and confirm the button changes **both in the editor canvas and on the front end**.
3. Clear it again and confirm the button returns to its size/style padding rather than to zero — spacing is rendered *only* when the preset sets it, which is the part most likely to break.
4. Repeat for Margin, and check one non-desktop breakpoint.

<details>

```bash
git checkout SOFT-4083/phase-14-button-spacing
bun run build-wp
```

</details>

## Screenshot

Spacing section with Padding and Margin fields, both unset
phase-14-button-spacing

---

Slice 22 of 23 in the SOFT-4083 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

- **New Features**
  - Added responsive spacing controls for button presets, including padding and margin.
  - Button styles now support preset-defined spacing while preserving explicit per-side overrides.
  - Added spacing token support to the style library and demo settings.
  - Preset spacing can automatically apply to buttons in both editing and published views.
  - Added safeguards so invalid or unavailable preset data leaves existing spacing unchanged.

### DIFF
--- a/src/blocks/singlebtn/components/backend-styles/index.js
+++ b/src/blocks/singlebtn/components/backend-styles/index.js
@@ -7,6 +7,30 @@ import {
 	getBorderColor,
 	getSpacingOptionOutput,
 } from '@kadence/helpers';
+import { blockDefaultPreset, blockPresetValues } from '../../../../extension/preset-picker';
+
+/**
+ * Whether the button's active preset resolves a padding and/or a margin.
+ *
+ * Reads the same preset surface the inspector does, so the canvas and the panel cannot disagree about
+ * whether a preset carries spacing. A block with no explicit selection follows the block's default
+ * preset, exactly as the server does.
+ *
+ * @param {Object} attributes The block attributes.
+ *
+ * @since TBD
+ *
+ * @return {{padding: boolean, margin: boolean}} Which spacing properties the preset defines.
+ */
+function presetSpacingProperties(attributes) {
+	const preset = attributes?.kbPreset || blockDefaultPreset('kadence/singlebtn');
+	const tokens = blockPresetValues('kadence/singlebtn')?.[preset] ?? {};
+
+	return {
+		padding: 'button-padding' in tokens,
+		margin: 'button-margin' in tokens,
+	};
+}
 
 export default function BackendStyles(props) {
 	const { attributes, isSelected, previewDevice, currentRef, context } = props;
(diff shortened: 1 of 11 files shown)


src/style-library/__tests__/presets.test.js
```diff
@@ -223,6 +223,10 @@ describe('presetInitialValues', () => {
 				'button-bg-hover': 'semantic.color.action-primary-hover',
 				'button-text-hover': 'semantic.color.on-primary',
 				'button-radius': '0.5rem',
+				// Bound but unvalued by this preset, so they seed empty — the shape that keeps a button's
+				// spacing owned by its size class until someone sets one.
+				'button-padding': '',
```
d4mation, line 308: This _is_ how it works, but right now if you call `presetSaveTokens()` it fails because it tries to initialize these to an empty string which causes a validation failure.

We need to fix that and add tests for this type of case to ensure that these values are not incorrectly saved when unset.

```json
{
  "data": {
    "status": 422,
    "block": "kadence/singlebtn",
    "errors": [
      {
        "path": "$extensions.com.kadence.designTokens.presets.kadence/singlebtn.button.tokens.button-padding",
        "code": "value_invalid",
        "message": "A foundation-preset/block-preset token value must be an alias, a non-empty literal, a slot list, or a responsive entry."
      },
      {
        "path": "$extensions.com.kadence.designTokens.presets.kadence/singlebtn.button.tokens.button-margin",
        "code": "value_invalid",
        "message": "A foundation-preset/block-preset token value must be an alias, a non-empty literal, a slot list, or a responsive entry."
      }
    ]
  }
}
```

pauloiankoski: Fixed, unset properties get omitted instead of sent as empty strings. Heads up though: omitting isn't clearing — merge() keeps the stored value, so clearing a saved padding still won't revert to the default.

src/blocks/singlebtn/components/backend-styles/index.js
```diff
@@ -7,6 +7,30 @@ import {
...
+ * @return {{padding: boolean, margin: boolean}} Which spacing properties the preset defines.
+ */
+function presetSpacingProperties(attributes) {
+	const preset = attributes?.kbPreset || blockDefaultPreset('kadence/singlebtn');
+	const tokens = blockPresetValues('kadence/singlebtn')?.[preset] ?? {};
+
+	return {
+		padding: 'button-padding' in tokens,
+		margin: 'button-margin' in tokens,
+	};
```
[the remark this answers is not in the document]
pauloiankoski: Fixed, extracted to a shared activePresetFor() helper in preset-picker/index.js (validated against the preset map, mirroring has_preset()/default_preset()) and used at all 4 places that had the same unvalidated selected || blockDefaultPreset(...) pattern: this file, PresetButton.js (x2), and capture.js. Added a dedicated test file plus verified live in the editor that a deleted-preset kbPreset still falls back to the default's spacing.

src/style-library/__tests__/presets.test.js
pauloiankoski: That landed in 4a39665f2 on phase-8 (PR #1279), not on this slice directly — every downstream branch here inherits it through the base chain, so it won't show up in #1289's own commit list.

d4mation (approved):

PR #1290, opened 2026-08-17
[DTM] SOFT-4083 [23/23]: Field built-in defaults

pauloiankoski:
Shows a field's built-in default muted when the preset sets nothing — in the default's own unit, with the picker kept to that field's role scale — and spaces the screen header from the content below it.

## Test instructions

1. Open a Button preset with Padding unset: it should read **`Default 0.4em 1em`**, muted, in the default's own `em` unit rather than a bare number.
2. Click that unset field. It must open on the **Style Library tab (the token list)**, not Custom — the fallback is the block's own value, not a choice the user made.
3. Confirm the picker offers the **spacing** scale, not the radius scale.
4. Switch to the Custom tab and confirm it seeds from that same default, and that the unit shown matches it.
5. Set a real value and confirm the label stops reading `Default`; clear it and confirm the muted default comes back.

<details>

```bash
git checkout SOFT-4083/phase-15-field-defaults-polish
bun run build-wp
```

</details>

## Screenshots

Padding and Margin showing their built-in defaults muted
phase-15-field-defaults-polish

An unset spacing field opens on the token list, scoped to the spacing scale
phase-15-token-popover

---

Slice 23 of 23 in the SOFT-4083 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

* **New Features**
  * Added semantic design-token controls for button padding and margins, including independent top, right, bottom, and left values.
  * Button spacing can now update dynamically from configured design-token values while preserving existing defaults.
  * Added consistent button radius defaults matching the built-in styling.
  * Style controls now display inherited breakpoint values and retain appropriate units when values are unset.

* **Bug Fixes**
  * Improved handling of empty, cleared, and partially populated spacing presets.


src/style-library/presets/button-preset.js
d4mation, line 1: This hardcoding of default values could be problematic.

We have cases like 

```scss
border-radius: var(--kb-btn-radius, var(--kb-token--semantic--radius--control, 3px));
```

where the default styling is explicitly pointed to a Token as a fallback so that we can ensure that if that Token is modified, it is reflected globally properly while still having the same CSS value as `master` does now.

If that Token were to be updated, then this hardcoded default value would be incorrect. 

Padding and Margin should be given the same type of fallback as Border Radius where its default CSS uses a CSS Variable which points to matching semantic Tokens (that apparently don't exist yet) to ensure that if those Tokens are updated, a button without customized Padding/Margins inherits them properly too.

Then this configuration will need to be updated to account for Token-backed default values rather than hardcoding them.

There is a _potential weirdness_ with how `.kb-button.kb-btn-global-outline` has a `calc()` to account for border width, but since we would need to also redirect Default Border Width to a Token Variable here too I don't think it is unsolvable (and that case wouldn't be shown in the Preset Editor anyway, as it is specific to the "Button Inherit Style: Outline" setting).

pauloiankoski: Added 8 per-side semantic tokens and routed the default CSS through them. Preset defaults resolve from tokens now instead of hardcoded values.

includes/resources/Design_Tokens/Registry/Baseline/baseline.json
```diff
@@ -489,6 +489,38 @@
...
+				"$type": "dimension",
+				"$value": "0"
+			},
+			"button-margin-bottom": {
+				"$type": "dimension",
+				"$value": "0"
+			},
+			"button-margin-left": {
+				"$type": "dimension",
+				"$value": "0"
```
d4mation, line 523: The only concern I have with this is that the other Button-specific semantic we have right now uses `control` with the thought process that it would apply to all similar "form control"-type elements, but I suppose they likely wouldn't share these paddings 🤔 This is probably fine. It may be worthwhile checking to see if that `3px` button border radius is used anywhere else and if not, just renaming that one to be `button` as well.

src/blocks/advancedbtn/style.scss
```diff
@@ -72,7 +72,7 @@
 	border: 2px solid var(--global-palette-btn-bg, #3633e1);
 	background: transparent;
 	color: var(--global-palette-btn-bg, #3633e1);
-	padding: calc(0.4em - 2px) calc(1em - 2px);
+	padding: calc(var(--kb-token--semantic--spacing--button-padding-top, 0.4em) - 2px) calc(var(--kb-token--semantic--spacing--button-padding-right, 1em) - 2px);
```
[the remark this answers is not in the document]
pauloiankoski: Fixed in 64d8f6980 — all four outline-button padding rules (style.scss and its editor.scss mirror) now use the full 4-value shorthand with all four semantic tokens instead of the 2-value form.

src/style-library/__tests__/presets.test.js
```diff
@@ -426,6 +426,61 @@ describe('presetSaveTokens', () => {
...
+	 *
+	 * @return void
+	 */
+	it('omits an already-stored property the draft cleared back to empty', () => {
+		const initialTokens = { 'button-padding': '0.4em' };
+		const storedTokens = { 'button-padding': '0.4em' };
+
+		const result = presetSaveTokens({ 'button-padding': '' }, initialTokens, storedTokens);
+
+		expect(result).not.toHaveProperty('button-padding');
```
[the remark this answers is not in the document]
pauloiankoski: Fixed in 2c5c21d3e — Presets_Controller::create_item() now replaces a preset's token map wholesale instead of merging it property by property, so an omitted (cleared) property is actually removed. Gated on the raw tokens request param, not the synthesized block node, so a label-only rename still leaves stored tokens untouched. Added a regression test.

d4mation (approved):

PR #1293, opened 2026-08-17
[SMTNC-2035] WP admin notice prompting site administrators to update WordPress

claudiosanches:
[SMTNC-2035](https://linear.app/nexcess/issue/SMTNC-2035)

https://www.loom.com/share/caa2ad962db9462ba3aa892aa4fc1d7e

Shows a dismissible notice prompting site administrators to update WordPress when the install is out of date, using `stellarwp/core-update-notice` 1.0.0.

The notice is limited to this plugin's own screens: Home, Settings, and the screens for the post types Kadence Blocks puts under its own menu (`kadence_form`, `kadence_navigation`, `kadence_header`). Those three are matched on post type rather than screen ID, so a list table and its add and edit screens are covered by one entry. `kadence_lottie` and `kadence_vector` are excluded, being registered with `show_in_menu => false`.

Dismissal state lives in an option shared with the other StellarWP plugins carrying the package, so a site running several of them only dismisses once, and it is keyed per WordPress version so dismissing one release does not hide a later security release.

Unrelated to this change, but found while confirming the screen IDs: `includes/assets/css/admin-dashboard.min.css` styles `.kadence-blocks_page_kadence-blocks-home`, while the real screen is `kadence_page_kadence-blocks-home`. That selector has never matched.

### Checklist


includes/resources/Notice/Notice_Provider.php
```diff
@@ -0,0 +1,102 @@
...
+				'heading' => __( 'Keep your site protected. Update to the latest version of WordPress.', 'kadence-blocks' ),
+				'body'    => __( 'Your site is running on an outdated version of WordPress, which can leave it vulnerable to security issues. To decrease your risk of exposure, please update your WordPress install to the latest version.', 'kadence-blocks' ),
+				'dismiss' => __( 'Dismiss this notice.', 'kadence-blocks' ),
+			]
+		);
+
+		// Bind the registered instance so the rest of the plugin resolves the same one.
+		$this->container->singleton( CoreUpdateNotice::class, $notice );
+
+		Register::notice( $notice, [ $this, 'is_plugin_admin_screen' ] );
```
defunctl:
```suggestion
		Register::notice( $this->container->get( CoreUpdateNotice::class ), [ $this, 'is_plugin_admin_screen' ] );
```

includes/resources/Notice/Notice_Provider.php
```diff
@@ -0,0 +1,102 @@
...
+	/**
+	 * Whether the current screen belongs to this plugin.
+	 *
+	 * Runs during admin_notices, when get_current_screen() is available.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public function is_plugin_admin_screen(): bool {
```
defunctl: This shouldn't be in a Provider, a Provider is only for registering container definitions and WordPress hooks.

Make a new class with `__invoke()` and use that instead. It can have its own unit tests as well.

defunctl (changes requested):

includes/resources/Notice/Notice_Provider.php
```diff
@@ -0,0 +1,63 @@
...
+					[
+						'heading' => __( 'Keep your site protected. Update to the latest version of WordPress.', 'kadence-blocks' ),
+						'body'    => __( 'Your site is running on an outdated version of WordPress, which can leave it vulnerable to security issues. To decrease your risk of exposure, please update your WordPress install to the latest version.', 'kadence-blocks' ),
+						'dismiss' => __( 'Dismiss this notice.', 'kadence-blocks' ),
+					]
+				);
+			}
+		);
+
+		$this->container->singleton( Is_Plugin_Admin_Screen::class, Is_Plugin_Admin_Screen::class );
```
defunctl:
```suggestion
		$this->container->singleton( Is_Plugin_Admin_Screen::class );
```

Don't need the second param if it's exactly the same as the first.

defunctl (changes requested):
Just one minor thing otherwise this looks good now

defunctl (approved):

dpanta94 (approved):

dpanta94:
Not being shipped

PR #1294, opened 2026-08-17
[DTM] SOFT-4083 [24/24]: Loading skeletons

pauloiankoski:
Replaces the two screen-level loading spinners in the Style Library with skeleton placeholders that mirror the content they stand in for, so the layout does not jump when data lands.

Adds a shared `Skeleton` atom, then adopts it on the preset row list and the colour swatch grid. Both skeletons reuse the real row/swatch layout classes rather than duplicating their dimensions, so they stay in sync if those change.

The inline `isBusy` spinner in `SelectDropdown` and the app-boot spinners in `StyleLibraryApp`/`AppShell` are deliberately untouched — an inline indicator is not a skeleton case, and a whole-shell skeleton is a separate design decision.

## Test instructions

1. Go to **Style Library → Block Presets → Button** and reload. While the presets fetch, the panel shows row-shaped placeholders rather than a spinner, and the layout does not shift when the rows arrive.
2. Do the same on **Color Palette** — the swatch grid shows card-shaped placeholders.
3. Throttle the network (DevTools → Network → Slow 3G) if the fetch is too fast to see locally.
4. Turn on **Reduce Motion** (macOS: System Settings → Accessibility → Display) and reload. The shimmer stops; the static placeholder remains.
5. With a screen reader, confirm the loading region is announced as busy and the placeholder shapes are not read out as content.

<details>

```bash
git checkout SOFT-4083/phase-16-ux-polish
bun run build-wp
```

</details>

## Screenshot

Preset row skeletons while the presets load

---

Slice 24 of 24 in the SOFT-4083 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

### DIFF
--- a/src/style-library/__tests__/preset-screen.test.js
+++ b/src/style-library/__tests__/preset-screen.test.js
@@ -24,9 +24,9 @@ jest.mock('../hooks/use-preset-screen', () => ({
 // `jest.config.js` maps the `@wordpress/components` specifier to the copy nested under
 // `@kadence/components/node_modules`, which resolves its own nested `react`/`react-dom` — a
 // different module instance than the top-level `react-dom/client` this test renders with. Mounting
-// a real `Button`/`Notice`/`Spinner` from that nested copy under the top-level renderer trips
-// React's "Invalid hook call" guard. Simple stand-ins sidestep the cross-copy mismatch; this test
-// only needs to tell the loading/empty/populated states apart, not exercise the real controls.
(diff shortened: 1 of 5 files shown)


pauloiankoski:
Closing to fold this work into the SOFT-4200 stack (shared @wordpress/data store for the Style Library) as its first slice — the skeleton loading UI this PR added is a direct prerequisite for that stack's sidebar-loading-state task.

PR #1299, opened 2026-08-18
SOFT-4087 [4/6]: apply per-block palette and preset markup generically for dynamic blocks

d4mation:
🎫  https://linear.app/nexcess/issue/SOFT-4087/color-control-soft-3995-style-token-picker-in-picker-palette-switching
📹  https://www.loom.com/share/0c508363df8841a09bc11b577d113547

Applies a block's per-block design-token selections to dynamic blocks **generically**, in the abstract block's `render_css()`:

- `render_palette_attribute()` sets `data-kb-palette="<id>"` from `kbPalette`.
- `render_preset_class()` adds the `kb-preset--<slug>` class from `kbPreset`.

Kept as two methods since palette and preset are distinct concerns. Each uses `WP_HTML_Tag_Processor`, stops at the opening tag (nested children are not walked), and short-circuits before any parsing when nothing is pinned. A dynamic block now opts in with just a `block.json` support line — no per-block PHP.

Opts Row Layout (`kadence/rowlayout`) and Advanced Text (`kadence/advancedheading`) into the palette override this way, and `kadence/singlebtn` drops its per-block palette/preset merges to render through the generic path.

Stacked on #1298.

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Advanced Heading blocks now support color palette settings.
  * Row Layout blocks now support color palettes and enhanced block metadata.
  * Block styling information is applied consistently during dynamic rendering.

* **Bug Fixes**
  * Improved handling of palette and preset styling so invalid or empty values do not affect rendered content.
  * Button blocks no longer receive unintended palette or preset styling attributes.

### DIFF
--- a/includes/blocks/class-kadence-blocks-abstract-block.php
+++ b/includes/blocks/class-kadence-blocks-abstract-block.php
@@ -358,8 +358,9 @@ class Kadence_Blocks_Abstract_Block {
 	 * `kbPalette` attribute (via `kbPalette` block support) and needs no per-block PHP. A no-op when no palette
 	 * is pinned, the content is empty, or it has no opening tag to carry the attribute.
 	 *
-	 * Only the opening tag is touched: the tag processor stops at the first tag, so a block with many nested
-	 * children is not walked, and the whole method short-circuits before any parsing when nothing is pinned.
+	 * The attribute lands on the block's own root element (the tag carrying its `wp-block-<namespace>-<name>`
+	 * class), not an outer wrapper such as an animation `data-aos` div, and the whole method short-circuits
+	 * before any parsing when nothing is pinned.
 	 *
 	 * @since TBD
 	 *
@@ -381,7 +382,7 @@ class Kadence_Blocks_Abstract_Block {
 
 		$tags = new WP_HTML_Tag_Processor( $content );
 
-		if ( ! $tags->next_tag() ) {
+		if ( ! $tags->next_tag( [ 'class_name' => 'wp-block-' . $this->namespace . '-' . $this->block_name ] ) ) {
 			return $content;
 		}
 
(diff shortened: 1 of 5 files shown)


src/blocks/advancedheading/block.json
d4mation, line 1: This originally was only added in this PR because an earlier iteration of these changes injected the Palette Attribute into specifically this Block due to how it was built. However, we are able to control this more generically with the changes to the Abstract class now and as a result, the only required step to "opt-in" to Palette support is this change to `block.json`.

includes/blocks/class-kadence-blocks-abstract-block.php
```diff
@@ -343,6 +351,86 @@ public function build_html( $attributes, $unique_id, $content, $block_instance )
...
+
+		$classes = $this->preset_classes( $attributes['kbPreset'] ?? '' );
+
+		if ( $classes === [] ) {
+			return $content;
+		}
+
+		$tags = new WP_HTML_Tag_Processor( $content );
+
+		if ( ! $tags->next_tag() ) {
```
pauloiankoski: Found a bug testing this: when a Single Button has both a scroll animation and a preset, the preset class ends up on the AOS wrapper div instead of the button. Tested with `kbPreset: "secondary"` + `kadenceAnimation: "fade-up"`:

```html
<div data-aos="fade-up" class="kb-blocks-button-aos ... kb-preset--secondary">
  <span class="... wp-block-kadence-singlebtn"></span>
</div>
```

The preset CSS is `.wp-block-kadence-singlebtn.kb-preset--secondary`, so it never matches — checked with `getComputedStyle`, the override is genuinely lost, not just cosmetic.

Fix: target the tag by its `wp-block-...` class instead of taking the first one:
```php
$processor->next_tag( [ 'class_name' => 'wp-block-' . str_replace( '/', '-', $this->block_name ) ] );
```

d4mation: Fixed in https://github.com/stellarwp/kadence-blocks/pull/1299/commits/cf79fba6e68c756feec85a9fff5dae7604305b0d

Note: That's a Kadence Blocks Pro feature, so it is only testable if Kadence Blocks Pro is also active. Good catch!

pauloiankoski (approved):

PR #1304, opened 2026-08-19
SOFT-4199: Token Picker: order the pickable pool by the stored token sort order

d4mation:
🎫  https://linear.app/nexcess/issue/SOFT-4199/token-picker-pickable-pool-ignores-the-token-sort-order

:video_camera:  https://www.loom.com/share/3ca51151f2c24880b8428a2be0555882

The editor Token Picker lists tokens in registry-insertion order, so a custom token dragged into the middle of a group on the Style Library screen shows up last in the picker: user primitives register after the baseline, and nothing between the registry and the picker reorders anything.

`Token_Sorter` owns the per-group permutation the admin feed applies (listed ids first by position, unmentioned ids appended in declaration order so a newly added token is never hidden). Both `Admin\Feed\Builder` and `Editor\Pickable_Tokens_Catalog` go through it, so the picker and the Style Library screen cannot drift apart.

The pool groups by each token's UI group, the bucket `to_ui_schema()` keys the feed's schema by and the one the reorder endpoint writes against, permutes each group by the **active** library's stored `tokenOrder`, and writes the result back into the slots that group already occupied. The flat pool therefore only ever moves within a group, leaving the picker's type filter and its semantic-before-primitive ranking untouched. No JS, REST or schema changes.

Also here: `Token_Store::get_decoded_document()` replaces nine copies of the read-then-decode-then-`is_array` body spread across the REST, Resolver, Registry, Admin and Foundation_Presets layers.

Stacked on #1303 (SOFT-4194).

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Token groups now follow saved library ordering in the editor and administrative feeds.
  * Active library ordering is applied consistently to available tokens and presets.
  * User-created tokens can appear at their saved positions within groups.

* **Bug Fixes**
  * Unrecognized or outdated ordering entries no longer remove tokens.
  * Missing or invalid token-library documents are handled safely without disrupting results.

* **Refactor**
  * Document loading and token ordering are now handled consistently across editing, resolving, and REST operations.

### DIFF
--- a/includes/resources/Design_Tokens/Admin/Feed/Feed_Assembler.php
+++ b/includes/resources/Design_Tokens/Admin/Feed/Feed_Assembler.php
@@ -152,7 +152,7 @@ final class Feed_Assembler {
 			$resolved = false; // Corrupt stored document. Fail open: ship structure only.
 		}
 
-		$document = $this->read_document( $slug );
+		$document = $this->store->get_decoded_document( $slug );
 
 		return $this->builder->build(
 			$values,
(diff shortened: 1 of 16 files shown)


includes/resources/Design_Tokens/Admin/Feed/Feed_Assembler.php
```diff
@@ -204,27 +204,4 @@ private function rest(): array {
...
-		$raw = $this->store->get_document( $slug );
-
-		if ( $raw === '' ) {
-			return [];
-		}
-
-		$decoded = json_decode( $raw, true );
-
-		return is_array( $decoded ) ? $decoded : [];
-	}
```
d4mation, line 229: This exact same pattern was in the code 6 times and we needed it a 7th time. Took the opportunity to clean it up while we can.

pauloiankoski (approved):

PR #1305, opened 2026-08-20
fix: pin react and react-dom to matching versions

pauloiankoski:
## Summary

`react` and `react-dom` were resolving to mismatched versions (`react@19.2.6` vs `react-dom@18.3.1`) because neither is a direct dependency of this plugin — both were purely transitive, pulled in at conflicting version ranges by different `@wordpress/*` packages, and the package manager hoisted them independently. This broke every React-rendering unit test with errors like "Objects are not valid as a React child" and un-`act()`-wrapped state-update warnings.

Pins both as direct `devDependencies` at `^18.3.1`, matching what `@wordpress/element@6.46.0` and `@wordpress/scripts@30.19.0`'s peer dependencies actually expect. `react`/`react-dom` are never bundled at runtime regardless — `DependencyExtractionWebpackPlugin` externalizes every `@wordpress/*` import (and by extension `@wordpress/element`'s own React usage) to WordPress's own runtime globals, so this only affects local tooling resolution (Jest, editor intellisense).

## Test plan

- [x] `npx jest` — 48/48 suites, 800/800 tests passing (previously 10 suites / 36 tests failing with this exact mismatch)
- [x] Confirmed no duplicate nested `react` copy remains under `node_modules/react-dom/node_modules/`

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -67,6 +67,8 @@
         "gulp-sass": "^6.0.0",
         "install": "^0.13.0",
         "prettier": "^3.6.0",
+        "react": "^18.3.1",
+        "react-dom": "^18.3.1",
       },
     },
   },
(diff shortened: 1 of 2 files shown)


pauloiankoski:
Closing to reopen as SOFT-4200/phase-1 of the shared-data-store stack, since every phase after it needs this fix.

PR #1306, opened 2026-08-20
[DTM] SOFT-4200 [1/7]: Fix react/react-dom version mismatch

pauloiankoski:
## Summary

`react` and `react-dom` were resolving to mismatched versions (`react@19.2.6` vs `react-dom@18.3.1`) — neither is a direct dependency of this plugin, so the package manager hoisted each independently at conflicting version ranges pulled in by different `@wordpress/*` packages. This broke every React-rendering unit test (`Objects are not valid as a React child`, un-`act()`-wrapped state-update warnings).

Pins both as direct `devDependencies` at `^18.3.1`, matching what `@wordpress/element@6.46.0` and `@wordpress/scripts@30.19.0`'s peer dependencies actually expect. Neither is bundled at runtime regardless — `DependencyExtractionWebpackPlugin` already externalizes every `@wordpress/*` import (and by extension `@wordpress/element`'s own React usage) to WordPress's own runtime globals, so this only affects local tooling resolution.

Folded into this stack as its first slice, rather than a standalone PR, since every later slice depends on Jest passing cleanly.

## Test plan

- [x] `npx jest` — every suite passes (previously several suites failed with this exact mismatch)
- [x] Confirmed no duplicate nested `react` copy remains under `node_modules/react-dom/node_modules/`

<details>

```bash
bun install
```

</details>

---

Slice 1 of 7 in the SOFT-4200 stack. Read bottom-up: this PR is understandable from its own diff.

## Summary by CodeRabbit

* **Chores**
  * Added React and React DOM development dependencies.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -67,6 +67,8 @@
         "gulp-sass": "^6.0.0",
         "install": "^0.13.0",
         "prettier": "^3.6.0",
+        "react": "^18.3.1",
+        "react-dom": "^18.3.1",
       },
     },
   },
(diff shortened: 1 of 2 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1307, opened 2026-08-20
[DTM] SOFT-4200 [2/7]: Loading skeletons, everywhere a real fetch is in flight

pauloiankoski:
## Summary

Replaces spinners with skeleton placeholders across the Style Library page, scoped to exactly the controls that do a real, awaitable fetch — and nowhere else.

Adds a shared `Skeleton` atom, then adopts it on the preset row list and the color-palette swatch grid, replacing their screen-level spinners. A generic `isLoading` capability is added to the shared `SelectDropdown` molecule (skeleton rows in the open menu instead of an empty list) and wired into the library-picker dropdown and the color-palette picker. Both preset and palette settings panels also gain a loading skeleton for their cold-load state.

Deliberately does **not** touch any of the six "scale-type" screens (Border Radius, Border Width, Spacing, Icon Sizes, Typography, Shadow) — they read synchronously off the already-loaded design-token feed and never fetch, so a loading state there would be theater. The inline `isBusy` spinner in `SelectDropdown` and the app-boot spinners in `StyleLibraryApp`/`AppShell` are also untouched — an inline write-in-flight indicator and a whole-shell boot state are different design decisions.

## Screenshots

**Block Presets panel — spinner (before) vs. skeleton rows (after), network throttled:**
_paste screenshot here_

**Color Palette swatch grid — skeleton cards, network throttled:**
_paste screenshot here_

**Library / Color Palette dropdown — skeleton rows in the open menu:**
_paste screenshot here_

## Test instructions

1. Go to **Style Library → Block Presets → Button** and reload with the network throttled (DevTools → Slow 3G). The panel shows row-shaped placeholders instead of a spinner.
2. Do the same on **Color Palette** — the swatch grid shows card-shaped placeholders.
3. Open the library dropdown, and the Color Palette picker, before either has loaded this session — confirm skeleton rows instead of an empty menu.
4. Open a preset's settings panel, and a palette swatch's settings panel, before the data has loaded — confirm skeleton fields.
5. Un-throttle, reload, and repeat each — confirm real content shows immediately once already loaded, no skeleton flash on warm data.
6. Turn on **Reduce Motion** (macOS: System Settings → Accessibility → Display) and reload — the shimmer stops, the static placeholder remains.

<details>

```bash
bun install
```

</details>

---

Slice 2 of 7 in the SOFT-4200 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

* **New Features**
  * Added skeleton loading placeholders across style library screens, dropdowns, preset panels, and color palette settings.
  * Added accessible loading states with localized status text.
  * Added busy-state feedback when creating presets, preventing duplicate submissions.

* **Bug Fixes**
  * Improved loading behavior and accessibility for library and preset selection interfaces.

### DIFF
--- a/src/style-library/components/atoms/Skeleton.js
+++ b/src/style-library/components/atoms/Skeleton.js
@@ -1,0 +1,36 @@
+/**
+ * A single shimmering placeholder shape, used to build screen-level loading skeletons (a preset
+ * row, a swatch card, …). Purely decorative — `PresetScreen`/`ColorPaletteScreen` wrap the
+ * composed skeleton in its own `role="status"`/`aria-live` container, so this atom marks itself
+ * `aria-hidden` and carries no accessible name of its own.
+ */
+
+/**
+ * Internal dependencies
+ */
+import './Skeleton.scss';
+
+/**
+ * Render one skeleton placeholder shape.
+ *
+ * @param {Object} [props]           The component props.
+ * @param {string} [props.className] An additional class name — typically a real layout class (e.g.
+ *                                   `list-row-label`) so the shape inherits that column's exact
+ *                                   width instead of a duplicated literal.
+ * @param {Object} [props.style]     Inline style, e.g. a `width`/`height` pinned to a shared size
+ *                                   token (`var(--kb-sl-size-row-preview)`) for a shape with no
+ *                                   layout class of its own to borrow from.
+ *
+ * @since TBD
+ *
+ * @return {JSX.Element} The placeholder shape.
+ */
+export function Skeleton({ className = '', style }) {
+	return (
+		<span
+			className={`kadence-blocks-style-library__skeleton ${className}`.trim()}
+			style={style}
+			aria-hidden="true"
+		/>
+	);
+}
--- a/src/style-library/components/pages/ColorPaletteScreen.js
+++ b/src/style-library/components/pages/ColorPaletteScreen.js
@@ -181,6 +181,7 @@ export function ColorPaletteScreen({ label, route, navigate, library }) {
 							// someone merely looks at a different palette; see `openPaletteFlow`'s own docblock.
 							onChange={(id) => palettes.openPalette(id).catch(() => {})}
 							isBusy={palettes.isBusy}
+							isLoading={palettes.isLoading}
 							error={palettes.openError}
 							onClearError={palettes.clearOpenError}
 							trailingAction={{
--- a/src/style-library/components/pages/ColorPaletteScreen.js
+++ b/src/style-library/components/pages/ColorPaletteScreen.js
@@ -10,7 +10,7 @@
  * WordPress dependencies
  */
 import { useMemo, useState } from '@wordpress/element';
-import { Button, DropdownMenu, MenuGroup, MenuItem, Notice, Spinner } from '@wordpress/components';
+import { Button, DropdownMenu, MenuGroup, MenuItem, Notice } from '@wordpress/components';
 import { __, sprintf } from '@wordpress/i18n';
 import { moreVertical, plus } from '@wordpress/icons';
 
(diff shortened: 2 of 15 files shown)


pauloiankoski:
@coderabbitai review

src/style-library/components/pages/ColorPaletteScreen.js
```diff
@@ -39,6 +40,56 @@ import {
...
+				{/* No real heading class carries a width of its own — `SectionHeading`'s width is
+				 * whatever its text measures — so this bar's width is a plain literal, not a reused
+				 * layout value. */}
+				<Skeleton className="kadence-blocks-style-library__skeleton--bar" style={{ width: '8rem' }} />
+				<div className="kadence-blocks-style-library__swatch-group-row">
+					{SKELETON_SWATCH_IDS.map((id) => (
+						<div key={id} className="kadence-blocks-style-library__swatch-card">
+							<div className="kadence-blocks-style-library__swatch-card-main">
+								<Skeleton className="kadence-blocks-style-library__swatch-card-preview" />
+								<Skeleton className="kadence-blocks-style-library__swatch-card-name kadence-blocks-style-library__skeleton--bar" />
```
d4mation: `&__swatch-card-name` declares only `max-width: 100%` without a `width` and since the parent here has `flex-direction: column; align-items: flex-start` it results in an empty `<span>` for the skeleton rendering as `1.25rem x 0`. We likely need a similar fix here as the group heading to force a width.

src/style-library/components/pages/PresetSidebar.js
```diff
@@ -204,6 +205,15 @@ export function PresetSidebar({ route, navigate, screen, preset }) {
 		}
 	}, [id, screen.isLoading, screen.loadError, hasInitialValues, navigate]);
 
+	if (id && screen.isLoading) {
+		return (
```
d4mation, line 216: >  The panel, or null while a stale `kb-item` self-heals for a tick, or while a valid one's presets are still loading.

While a preset is still loading we no longer return null, so we need to fix this.

src/style-library/components/atoms/Skeleton.js
```diff
@@ -0,0 +1,36 @@
...
+ * @since TBD
+ *
+ * @return {JSX.Element} The placeholder shape.
+ */
+export function Skeleton({ className = '', style }) {
+	return (
+		<span
+			className={`kadence-blocks-style-library__skeleton ${className}`.trim()}
+			style={style}
+			aria-hidden="true"
```
d4mation, line 33: If this is not in an `aria-live` region, then on load we aren't announcing anything. Worth double-checking to make sure we're doing that in every case. It makes sense for the skeleton itself to be invisible, but we need to let screen readers know once it has loaded.

src/style-library/components/pages/ColorPaletteScreen.js
pauloiankoski: Fixed in 6ccbd8d33. Added an explicit `style={{ width: '70%' }}` on that skeleton, mirroring the group-heading bar's own existing `8rem` fix right above it.

src/style-library/components/pages/PresetSidebar.js
pauloiankoski: Fixed in 6ccbd8d33 — corrected the `@return` to describe the panel's actual three states (real panel, loading skeleton, or null while a stale item self-heals) in the code's own order.

src/style-library/components/atoms/Skeleton.js
pauloiankoski: Fixed in 6ccbd8d33, across all 5 skeleton sites in this app (`ColorPaletteScreen`, `ColorPaletteSettings`, `PresetScreen`, `PresetSidebar`, `SelectDropdown`). Added a new `useLoadingAnnouncement(isLoading, message)` hook — on mount it creates its own persistent, visually-hidden `role="status" aria-live="polite"` node (removed on unmount), and writes `message` into it only on a true→false loading transition. Each skeleton's own `role="status"` region still announces "Loading X…" while it's showing (unchanged); this closes the gap where nothing announced completion once that region unmounted. Deliberately one node per hook instance rather than a shared global one, so two skeletons finishing in the same tick can't clobber each other's announcement.

d4mation (approved):

PR #1308, opened 2026-08-20
[DTM] SOFT-4200 [3/7]: Store scaffold + libraries slice

pauloiankoski:
## Summary

Adds a real `@wordpress/data` store for the Style Library page (`kadence-blocks/style-library`) — a plain reducer, action creators, thunk resolvers that call the existing REST client functions, and selectors — registered so it's inspectable in Redux DevTools and via `wp.data.select('kadence-blocks/style-library')` from the console.

Migrates the first resource, `useLibraries()`, to read through the store instead of local `useState`/`useEffect`. Reads go through `useSelect`; the framework auto-triggers a resolver the first time its selector is read and tracks `isResolving`/resolution-finished state per exact argument tuple — no manual fetch-on-mount effect needed. `isLoading` is deliberately gated on "no data yet," not raw `isResolving` alone, so a write-triggered background reload never re-flashes the loading skeleton once data has already loaded.

`helpers/library-flows.js` — the existing write orchestration (create/rename/delete/activate) — is **not modified**. It already takes a `loadLibraries`/`refreshFeed` callback as its invalidation signal; only what `useLibraries()` passes into that slot changed, from a local re-fetch to `registry.dispatch(STORE_NAME).invalidateResolution(...)` + `registry.resolveSelect(STORE_NAME)...`.

`@wordpress/data` is added as a `devDependency` only — it's externalized to WordPress's own runtime global at build time like every other `@wordpress/*` import in this plugin, so this only affects local tooling resolution, never the shipped bundle.

## Test instructions

1. Go to **Style Library**, open the library dropdown — confirm the list loads and behaves exactly as before (open/create/rename/delete a library).
2. With Redux DevTools installed, confirm `kadence-blocks/style-library` appears as a store and `RECEIVE_LIBRARIES` fires as the list loads.
3. In the browser console: `wp.data.select('kadence-blocks/style-library').getLibraries()` returns the current list.

<details>

```bash
bun install
```

</details>

---

Slice 3 of 7 in the SOFT-4200 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

* **Improvements**
  * Improved Style Library loading and refresh behavior.
  * Library data now stays synchronized across multiple views, reducing duplicate loading requests.
  * Background updates refresh results without unnecessarily showing a loading state.
  * Library changes now trigger refreshed results automatically.

* **Bug Fixes**
  * Improved error handling when Style Library data cannot be loaded.

* **Tests**
  * Added coverage for loading, refreshing, deduplication, invalidation, and error scenarios.

### DIFF
--- a/src/style-library/hooks/use-libraries.js
+++ b/src/style-library/hooks/use-libraries.js
@@ -79,7 +79,10 @@ export function useLibraries(feed, refreshFeed) {
 	// re-resolves after create/rename/delete) — the store keeps serving the previous rows while that
 	// re-fetch is in flight, so there is already data to render and no loading state to show for it.
 	const { libraryRows, isLoading, loadFailure } = useSelect((select) => {
-		const rows = select(STORE_NAME).getLibraries();
+		// Guarded the same way `sortLibraries(libraryRows ?? [])` below already is — the selector's
+		// raw result is trusted nowhere else in this hook, and a throw here would crash the render
+		// instead of leaving the empty-list rendering `sortLibraries`'s own guard already handles.
+		const rows = select(STORE_NAME).getLibraries() ?? [];
 		return {
 			libraryRows: rows,
 			isLoading: select(STORE_NAME).isResolving('getLibraries', []) && rows.length === 0,
--- a/src/style-library/hooks/use-libraries.js
+++ b/src/style-library/hooks/use-libraries.js
@@ -74,14 +74,18 @@ export function useLibraries(feed, refreshFeed) {
 	const [activeSlug, setActiveSlug] = useState(editingSlug);
 
 	const registry = useRegistry();
-	const { libraryRows, isLoading, loadFailure } = useSelect(
-		(select) => ({
-			libraryRows: select(STORE_NAME).getLibraries(),
-			isLoading: select(STORE_NAME).isResolving('getLibraries', []),
+	// `isLoading` only reflects a cold load with nothing to show yet. `isResolving` alone would also
+	// flip true for every write-triggered background reload (`loadLibraries()`, below, invalidates and
+	// re-resolves after create/rename/delete) — the store keeps serving the previous rows while that
+	// re-fetch is in flight, so there is already data to render and no loading state to show for it.
+	const { libraryRows, isLoading, loadFailure } = useSelect((select) => {
+		const rows = select(STORE_NAME).getLibraries();
+		return {
+			libraryRows: rows,
+			isLoading: select(STORE_NAME).isResolving('getLibraries', []) && rows.length === 0,
 			loadFailure: select(STORE_NAME).getResolutionError('getLibraries', []),
-		}),
-		[]
-	);
+		};
+	}, []);
 	const libraries = useMemo(() => sortLibraries(libraryRows ?? []), [libraryRows]);
 
 	// The original hook's mount-effect `.catch()` wrote a failed list fetch into the SAME `openError`
--- a/src/style-library/hooks/use-libraries.js
+++ b/src/style-library/hooks/use-libraries.js
@@ -1,12 +1,12 @@
 /**
  * WordPress dependencies
  */
-import { useCallback, useEffect, useState } from '@wordpress/element';
+import { useCallback, useEffect, useMemo, useState } from '@wordpress/element';
+import { useSelect, useRegistry } from '@wordpress/data';
 
 /**
  * Internal dependencies
  */
-import { fetchLibraries } from '../api/client';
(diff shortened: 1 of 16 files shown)


pauloiankoski:
@coderabbitai review

src/style-library/hooks/use-libraries.js
```diff
@@ -74,16 +73,41 @@ export function useLibraries(feed, refreshFeed) {
...
+	// The flows in `helpers/library-flows.js` call this after create/rename/delete to refresh the list —
+	// unchanged from their point of view. Internally: `invalidateResolution` clears the "this selector
+	// call is already resolved" flag the framework tracks per argument tuple, and `resolveSelect` then
+	// re-runs the `getLibraries` resolver (rather than returning the now-invalidated cached result),
+	// resolving once the fresh rows land in the store — every mounted `useLibraries` instance re-renders
+	// from the same updated state automatically.
+	const loadLibraries = useCallback(() => {
+		registry.dispatch(STORE_NAME).invalidateResolution('getLibraries', []);
+		return registry.resolveSelect(STORE_NAME).getLibraries();
+	}, [registry]);
```
d4mation, line 113: This used to swallow failures, but now it propagates them. 

This can cause problems where a failure after a write can happen which causes the modal to stay open confusingly.

Reproduction:

1. Open the Style Library with at least two libraries; make sure you're editing a non-active, non-default one.
2. Open devtools and set up a fault that fails only the list read — block `GET /kadence-blocks/v1/design-tokens/libraries` (devtools Network → request blocking, or a `wp_die` in the collection endpoint's `get_items` permission callback). Leave `DELETE .../libraries/<non_default_slug>` working. Real-world equivalents: a Cloudflare 5xx on a single request, a session expiring between the two calls, or an nginx timeout under load.
3. Click Delete, confirm.
4. The row is gone from the database, but the modal stays open with an error, and the library dropdown also shows the same error. Click Delete again and you get a 404 — you're retrying a delete that already succeeded.

Expected: the delete succeeds, the modal closes, and the stale list error surfaces only in the dropdown.

src/style-library/hooks/use-libraries.js
```diff
@@ -74,16 +73,41 @@ export function useLibraries(feed, refreshFeed) {
...
+	const registry = useRegistry();
+	// `isLoading` only reflects a cold load with nothing to show yet. `isResolving` alone would also
+	// flip true for every write-triggered background reload (`loadLibraries()`, below, invalidates and
+	// re-resolves after create/rename/delete) — the store keeps serving the previous rows while that
+	// re-fetch is in flight, so there is already data to render and no loading state to show for it.
+	const { libraryRows, isLoading, loadFailure } = useSelect((select) => {
+		const rows = select(STORE_NAME).getLibraries();
+		return {
+			libraryRows: rows,
+			isLoading: select(STORE_NAME).isResolving('getLibraries', []) && rows.length === 0,
```
d4mation, line 88: `rows.length` is unguarded on the raw selector result, but `sortLibraries()` guards it. A throw from a `useSelect` mapper would crash the render instead of gracefully showing an empty list.

src/style-library/hooks/use-libraries.js
pauloiankoski: Fixed in 65b14d75a. `deleteLibraryFlow`'s post-delete `loadLibraries()` call is now wrapped in its own `.catch(() => {})` — a failed refetch there no longer falls into the flow's outer `catch`, which was setting `deleteError` and re-throwing even though the delete itself had already succeeded. A stale list still surfaces separately, through `getResolutionError('getLibraries', [])` feeding `openError` in `use-libraries.js`, rendered by the library dropdown — so it matches your expected behavior exactly: the delete succeeds, the modal closes, and the stale-list error shows only in the dropdown. Added a regression test pinning this: `deleteLibrary`/`refreshFeed` succeed, `loadLibraries` rejects, and the flow's own promise still resolves with `onError` never called.

One thing worth flagging separately: `createLibraryFlow` and `renameLibraryFlow` both still chain an unguarded post-write `loadLibraries()` into their own outer `.catch()`, so the identical bug is reachable there too (a failed refetch after a successful create/rename would misreport as a failure). Left out of scope here since your finding was specifically about delete — happy to file a follow-up for those two if you'd like.

src/style-library/hooks/use-libraries.js
pauloiankoski: Fixed in 65b14d75a — guarded with `?? []`, matching `sortLibraries(libraryRows ?? [])` right below it.

d4mation (approved):

PR #1309, opened 2026-08-20
[DTM] SOFT-4200 [4/7]: Presets slice

pauloiankoski:
## Summary

Adds the presets slice to the store (`getBlockPresets(namespace, block, slug)` resolver + selector) and migrates `usePresets()`/`usePresetScreen()` to read through it.

This is a genuine bug fix, not just a refactor: `ButtonScreen.js` and `ButtonSettings.js` each independently called `usePresetScreen()` today, so opening a preset's settings panel triggered its own separate REST fetch of the same preset collection the list screen had already loaded. Both now read the same store selector call, so only one of them ever issues the request, and a write from either instance updates what both render.

`helpers/preset-flows.js` is **not modified**. `usePresetScreen()`'s `refreshFeed` wrapper now does both the existing feed refresh (still needed — preset row previews read token values off the feed) and a store invalidate + re-resolve, but every write flow's own call site is unchanged.

Reorder's version tracking (the client-sent `version` a preset reorder must match, to guard against a stale concurrent write) is sourced from the shared feed's own `library.version`, not from the store's cached preset payload — every write anywhere in the library (scale, typography, palettes, presets) bumps `library.version` via `refreshFeed`, while the cached payload only catches up when this screen's own wrapped `refreshFeed` explicitly invalidates it. A write on any OTHER screen would otherwise leave the payload's version stale, and the next preset reorder would 409 against a version the server had already moved past. The refs this tracking depends on sync from a `useLayoutEffect` keyed on the feed version, not the render body, so a discarded or replayed render (React 18 concurrent rendering) can't leak a stale value into a queued reorder.

## Test instructions

1. Go to **Style Library → Block Presets → Button**, throttle the network, open a preset's settings panel — confirm it doesn't re-fetch the list you can already see (only one network request for presets total).
2. Create, save, delete, and reorder presets — confirm everything behaves exactly as before.
3. With Redux DevTools installed, confirm `RECEIVE_BLOCK_PRESETS` fires once per library/block combination, not once per panel open.
4. On **Block Presets → Button**, change a value on a different screen that shares this library (e.g. **Border Radius**), then go back and reorder a preset — confirm the reorder succeeds instead of failing with a version conflict.

<details>

```bash
bun install
```

</details>

---

Slice 4 of 7 in the SOFT-4200 stack. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

* **New Features**
  * Added centralized loading and caching for block presets.
  * Preset data remains visible during background refreshes.
  * Multiple views requesting the same presets share a single fetch.
  * Preset updates from other screens are detected and reflected automatically.

* **Bug Fixes**
  * Switching libraries clears stale preset data and displays the appropriate loading state.
  * Improved preset refresh behavior, version tracking, and handling of missing or unresolved data.

### DIFF
--- a/src/style-library/hooks/use-preset-screen.js
+++ b/src/style-library/hooks/use-preset-screen.js
@@ -18,7 +18,7 @@
 /**
  * WordPress dependencies
  */
-import { useCallback, useEffect, useMemo, useRef, useState } from '@wordpress/element';
+import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from '@wordpress/element';
 import { useRegistry } from '@wordpress/data';
 
 /**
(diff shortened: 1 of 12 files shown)


pauloiankoski:
@coderabbitai review

src/style-library/hooks/use-presets.js
```diff
@@ -38,61 +38,29 @@ export function usePresets(library, preset) {
...
+				Boolean(namespace && block && slug) &&
+				select(STORE_NAME).isResolving('getBlockPresets', [namespace, block, slug]) &&
+				!select(STORE_NAME).getBlockPresets(namespace, block, slug),
+			loadError:
+				namespace && block && slug
+					? select(STORE_NAME).getResolutionError('getBlockPresets', [namespace, block, slug])
+					: null,
+		}),
+		[namespace, block, slug]
+	);
```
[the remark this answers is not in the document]
pauloiankoski: Confirmed and fixed. `usePresetScreen`'s reorder version-tracking now sources from `library.version` (the shared feed's own version, bumped by every write anywhere in the library — scale, typography, palettes, presets) instead of the store's cached `getBlockPresets` payload version, matching `use-scale-screen.js`'s existing pattern. That payload only advanced when this screen's own wrapped `refreshFeed` explicitly invalidated it, so a write on any other screen left it stale and the next reorder here would 409 against a version the server had already moved past. See `src/style-library/hooks/use-preset-screen.js` and its updated tests in `use-preset-screen.test.js` (including a new regression test for the exact stale-payload-on-mount scenario).

pauloiankoski:
@coderabbitai review

src/style-library/hooks/use-preset-screen.js
```diff
@@ -52,37 +58,58 @@ export function usePresetScreen(library, preset) {
...
 
 	// Mirrors the version the queued reorder continuation below must send, so it always
 	// dereferences the live value at execution time, never one captured when the drop was enqueued
-	// — see `use-scale-screen.js`'s identical comment for why.
-	const feedVersionRef = useRef(payloadVersion);
-	feedVersionRef.current = writtenVersionRef.current ?? payloadVersion;
+	// — see `use-scale-screen.js`'s identical comment for why. See the module docblock for why this
+	// is sourced from `library.version`, not the store's cached preset payload.
+	const feedVersionRef = useRef(feedVersion);
+	feedVersionRef.current = writtenVersionRef.current ?? feedVersion;
```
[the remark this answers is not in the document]
pauloiankoski: Fixed. The four refs (`feedVersionSnapshotRef`, `writtenVersionRef`'s retirement check, `feedVersionAtWriteRef`, `feedVersionRef`) now sync from a `useLayoutEffect` keyed on `feedVersion`, not the render body. A layout effect still runs before the browser paints and before any user-triggered event handler (e.g. a drop that calls `reorderPresets`), so nothing that reads these refs can observe a value from a render React replayed or discarded. The direct writes inside `onVersion` and the failure handler are unchanged, since those already run outside render.

src/style-library/hooks/use-presets.js
```diff
@@ -38,61 +38,29 @@ export function usePresets(library, preset) {
...
+			// loaded once, the store keeps serving that payload while it revalidates, so there is data to
+			// keep rendering — a loading state should only ever show up before the first payload lands.
+			isLoading:
+				Boolean(namespace && block && slug) &&
+				select(STORE_NAME).isResolving('getBlockPresets', [namespace, block, slug]) &&
+				!select(STORE_NAME).getBlockPresets(namespace, block, slug),
+			loadError:
+				namespace && block && slug
+					? select(STORE_NAME).getResolutionError('getBlockPresets', [namespace, block, slug])
+					: null,
```
d4mation: `isLoading` reads `false` before resolution actually starts, and that can wipe a valid deep link (`?kb-item=<preset>`). 

It will start to load and see no data and then assume the preset has been deleted and not open it directly on load.

A similar `hasFinishedResolution` from `use-palettes` would fix it here.

src/style-library/hooks/use-preset-screen.js
```diff
@@ -52,37 +58,63 @@ export function usePresetScreen(library, preset) {
...
+
+			let storeRefresh = Promise.resolve();
+			if (namespace && block && slug) {
+				registry.dispatch(STORE_NAME).invalidateResolution('getBlockPresets', [namespace, block, slug]);
+				storeRefresh = registry.resolveSelect(STORE_NAME).getBlockPresets(namespace, block, slug);
+			}
+
+			return Promise.all([feedRefresh, storeRefresh]);
+		},
+		[library, registry, namespace, block, slug]
```
d4mation: This should depend on `library.refreshFeed` directly rather than `library` since it doesn't care if the rest of the object were to change. Otherwise this churns every render.

src/style-library/hooks/use-presets.js
pauloiankoski: Fixed in fb6548600 — switched to `hasFinishedResolution`, mirroring `use-palettes.js`'s pattern. Added a test proving `isLoading` is true on the very first render, before the resolver dispatch fires.

src/style-library/hooks/use-preset-screen.js
pauloiankoski: Fixed in fb6548600 — narrowed the dep to `library.refreshFeed` directly, since that's the only thing the callback reads off `library`. Confirmed no other `library.*` reads inside that callback body.

d4mation (approved):

PR #1313, opened 2026-08-20
[DTM] SOFT-4083 [24/38]: token-controls groundwork for border and shadow

pauloiankoski:
## Summary

Shared-library groundwork for the upcoming border and box-shadow token controls:

- `isUniformSlotList()` — a shallow-equal collapse helper for composite per-side values (e.g. `{width, style, color}`), additive alongside the existing `writeSlot()`/`toSlotList()` (which are left byte-for-byte unchanged — radius and spacing depend on their exact current `===`-based collapse behavior).
- `TokenPopover`'s Custom tab gains an optional `renderCustom` prop, letting a control supply its own Custom-tab body (e.g. a composite shadow editor) instead of the default numeric editor. Purely additive — every existing consumer (radius, spacing, border-width) that doesn't pass it renders exactly as before.

Both changes are prerequisites for `BorderControl` and `BoxShadowControl`, added in the following stacked PRs.

## Test plan

- [x] `npx jest src/token-controls` — full suite green, including every pre-existing test (proves radius/spacing/border-width paths are unaffected)
- [x] cspell clean
- [x] `grep -rn "SOFT-" src/token-controls` — no hits

## Summary by CodeRabbit

* **New Features**
  * Added an optional custom renderer for the Custom tab in token selection popovers.
  * Custom tab content can now be tailored while retaining the existing default display when no renderer is provided.
  * Added support for detecting uniform slot values, including scalar and simple composite values.

* **Tests**
  * Added coverage for custom tab rendering and slot-value comparisons, including differing values and key sets.

### DIFF
--- a/src/token-controls/helpers/value-shapes.js
+++ b/src/token-controls/helpers/value-shapes.js
@@ -133,53 +133,6 @@ export function writeSlot(value, index, next, collapse = false) {
 }
 
 /**
- * Whether every slot in a list is shallow-equal to the first — same top-level keys, same primitive
- * values. Used to collapse a composite per-side value (e.g. `{width, style, color}`) back to a
- * scalar the way `writeSlot`'s `===` check already does for plain string/number slots. `writeSlot`
- * itself is not touched: radius and spacing rely on its exact current behavior, and a composite
- * slot shape is unique to `BorderControl`.
- *
- * @param {Array} slots A four-element slot list.
- *
- * @since TBD
- *
- * @return {boolean} True when every slot has the same shape and values as the first.
- */
-export function isUniformSlotList(slots) {
-	const [first, ...rest] = slots;
-
-	return rest.every((slot) => shallowEqual(slot, first));
-}
-
-/**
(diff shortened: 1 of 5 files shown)


pauloiankoski:
@coderabbitai review

src/token-controls/helpers/value-shapes.js
```diff
@@ -132,6 +132,53 @@ export function writeSlot(value, index, next, collapse = false) {
...
+ *
+ * @since TBD
+ *
+ * @return {boolean} True when every slot has the same shape and values as the first.
+ */
+export function isUniformSlotList(slots) {
+	const [first, ...rest] = slots;
+
+	return rest.every((slot) => shallowEqual(slot, first));
+}
```
d4mation: This never ends up being used in any PR. Do we want to keep it?

pauloiankoski: Confirmed it's not called anywhere in this stack — removed `isUniformSlotList` and its now-unused `shallowEqual` helper, plus the dedicated test block, in d5d538316.

d4mation (approved):

PR #1325, opened 2026-08-21
[DTM] SOFT-4200 [7/7]: Optimistic writes + transaction feedback

pauloiankoski:
## Summary

Makes every write in the Style Library's Color Palette editor and scale-type screens (Scale, Spacing, Typography, Border Radius, Border Width, Icon Sizes, Shadow) feel instant, with a Snackbar confirming or reverting each one — the same pattern the Gutenberg post editor uses for "Post updated."

- **Save** updates the value on screen immediately; the write happens in the background, and a Snackbar confirms it or reverts the value and explains why.
- **Delete** shows a "Deleting…" dimmed, disabled state (not an instant vanish) until the write confirms — for both a swatch and an entire color group, whose heading and actions also dim while its swatches are removed.
- **Add color** / **Add group** / **Add** on every scale-type screen (Border Radius, Border Width, Spacing, Icon Sizes, Shadow, Typography's font size) show up immediately, with the write completing in the background; a duplicate group's "Add" modal now validates and closes on submit instead of waiting on the request. The new item's settings panel opens right away too, showing its Save/Delete buttons disabled until the write confirms.
- Every settings panel now shows "Saving…"/"Deleting…" on its footer buttons while a write is in flight, wired the same way the preset settings panel already does it.
- The app's cold-start and library-switch loading states are now a shell-shaped skeleton instead of a spinner, consistent with the rest of this app's loading states.
- The transaction Snackbar is pinned to the viewport and centered — without this, it rendered in normal document flow and was entirely invisible below the fold on this admin page (`@wordpress/components`' own `SnackbarList` styling assumes the block editor's chrome, which isn't present here).

## A related fix found along the way

A swatch's Delete button always wrote to the *default* palette regardless of which palette was actually open, so deleting an override on a non-default palette silently touched the wrong palette and the override was never reset. The dedicated revert-to-inherited endpoint already existed on the backend but nothing in the frontend called it. Delete is now Reset for a built-in swatch showing the open palette's own override (reverts to inherited, targeting that palette), and stays Delete for a custom, user-created swatch (removes it entirely, with a best-effort primitive cleanup) — neither button shows when there's nothing to do (editing the default palette itself, or a non-overridden built-in swatch).

## Also fixed during review

- A pending optimistic swatch edit was scoped to the whole library, not the specific palette being edited. Switching palettes while a write was still in flight — reachable via the browser's back/forward buttons, which bypass the palette selector's own busy-disabled state — could apply a pending edit to the wrong palette. Pending edits are now scoped per palette.
- `usePalettes()` and `useScaleScreen()` are each called as separate sibling instances by a screen and its settings panel, and each tracked its own "busy" flag locally, so a write started in one instance didn't block a conflicting write starting in the other. Both hooks now share write-lock state through the store, so a write in either instance blocks a concurrent one in its sibling.

## Test instructions

1. Open a non-default color palette, edit an overridden swatch's value or name, hit Save — confirm the swatch updates immediately and a "Swatch saved." toast appears at the bottom of the screen.
2. Delete a custom swatch — confirm it dims and disables instead of vanishing instantly, and is gone once the write confirms.
3. Add a color and add a color group — confirm both show up immediately, with no duplicate tile ever appearing, and the new swatch's settings panel opens right away with Save/Delete disabled until the "Color added."/"Color group added." toast appears.
3a. On a scale-type screen (e.g. Border Radius), click "+ Add" — confirm the new row shows up immediately and its settings panel opens right away with Save/Delete disabled until the "Token created." toast appears.
4. On a non-default palette, open a built-in swatch that's overridden there — confirm you see **Reset**, not Delete; click it and confirm the swatch reverts to the default palette's value and the network request targets the palette you're editing (not `default`).
5. Drag-reorder a group's swatches, then immediately save or delete a *different* swatch before the reorder's own write resolves — confirm the reorder order is not reverted.
6. Delete a color group — confirm the group's own heading and its Rename/Delete/Add-color controls dim and disable while its swatches are removed underneath.
7. Reload the Style Library page and switch libraries — confirm both loading states show a shell-shaped skeleton, not a spinner.
8. Trigger any write and confirm the resulting Snackbar is visible, centered, at the bottom of the screen — not below the fold.
9. Start editing a swatch on one palette, then — before the save resolves — switch to a different palette (e.g. the browser's Back button) — confirm the pending edit does not leak onto the palette you land on.

<details>

```bash
git checkout SOFT-4200/phase-7-optimistic-writes
bun install
```

</details>

---

Slice 7 of this stack, stacked on `SOFT-4200/phase-6-feed-slice`. Read bottom-up: this PR is understandable from its own diff plus the slice below it.

## Summary by CodeRabbit

- **New Features**
  - Added accessible loading skeletons for the Style Library shell.
  - Added centered snackbar notifications for successful and failed actions.
  - Added optimistic updates for palette and scale edits, additions, deletions, and reordering.
  - Added validation for new color-group labels, including empty and duplicate names.
  - Added support for resetting built-in swatches and deleting custom swatches.

- **Improvements**
  - Pending deletions remain visible but appear dimmed and cannot be edited or dragged.
  - Save, delete, reset, and add controls now show appropriate busy and disabled states.
  - Preset saves and deletions now provide confirmation messages.


pauloiankoski:
@coderabbitai review

src/style-library/components/organisms/AppShellSkeleton.scss
```diff
@@ -0,0 +1,23 @@
...
+// height) — it renders standalone at cold start (`StyleLibraryApp.js`, before `feed.isReady`) and,
+// reused as-is, inside `AppShell.js`'s `isBlocked` overlay (`.kadence-blocks-style-library__blocker`,
+// a `display: flex; align-items: center; justify-content: center` box with no width of its own).
+// Without sizing of its own, the skeleton shrink-wraps to its content and renders as a small,
+// centered box instead of filling the frame the way the spinner it replaced did.
+.kadence-blocks-style-library__shell-skeleton {
+	display: flex;
+	flex-direction: column;
+	width: 100%;
+	// Fills the blocker overlay, which already gives `height: 100%` a real ancestor to resolve
```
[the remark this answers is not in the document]
pauloiankoski: Fixed in dac7b6a59.

src/style-library/hooks/use-palettes.js
```diff
@@ -129,19 +146,34 @@ export function usePalettes(feed, refreshFeed, route, navigate) {
...
+				: EMPTY_OPTIMISTIC_SWATCH_EDIT,
+		[namespace, slug]
+	);
+
+	const editingRow = useMemo(
 		() => listing.palettes.find((row) => row.id === editingId) ?? null,
 		[listing.palettes, editingId]
 	);
 
+	const palette = useMemo(() => applyOptimisticOverlay(editingRow, overlay), [editingRow, overlay]);
```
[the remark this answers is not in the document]
pauloiankoski: Fixed in 4fcbd2323. Added `paletteEditKey(namespace, slug, paletteId)` and threaded `editingId` through every optimistic-swatch overlay read/write site in `use-palettes.js`, so a pending patch/deletion/addition is now scoped to the palette it targets, not just the library. Added a regression test that starts a write on one palette, switches `editingId` to a different one before the write settles (mirroring the `popstate` path that bypasses the busy-disabled selector), and asserts the other palette's data isn't touched by the still-pending patch.

src/style-library/hooks/use-scale-screen.js
```diff
@@ -39,19 +44,20 @@ import {
...
 	const [addError, setAddError] = useState(null);
-	const [saveError, setSaveError] = useState(null);
-	const [deleteError, setDeleteError] = useState(null);
 	const [orderError, setOrderError] = useState(null);
 	const [pendingOrder, setPendingOrder] = useState(null);
 
 	const feed = library.feed;
 	const feedVersion = feed?.version;
 
+	const overlay = useSelect((select) => select(STORE_NAME).getOptimisticScaleEdit(library.slug), [library.slug]);
```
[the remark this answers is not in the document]
pauloiankoski: Fixed in 4fcbd2323, for both `useScaleScreen` and `usePalettes` — the same local-`isBusy`-per-instance pattern existed in both hooks. Moved the write lock into the store as new `paletteBusy`/`scaleBusy` reducer slices (`setPaletteBusy`/`setScaleBusy` actions, `getPaletteBusy`/`getScaleBusy` selectors), keyed by library for palettes and by slug for scale-type. Both hooks now read `isBusy` via `useSelect` and dispatch through it, so a write started in one sibling instance (e.g. a settings panel) is now visible to the other (e.g. the screen). Added a regression test mounting two sibling instances against one registry, proving instance B's `isBusy` becomes true while instance A's write is in flight and both clear once it settles.

src/style-library/hooks/use-scale-screen.js
```diff
@@ -114,12 +118,27 @@ export function useScaleScreen(config, library, route, navigate) {
 			refreshFeed: library.refreshFeed,
 			onBusy: setIsBusy,
 			onError: setAddError,
+		}).then((result) => {
+			notifySuccess(__('Token created.', 'kadence-blocks'));
+			return result;
 		});
 	}, [config, library, feedVersion]);
```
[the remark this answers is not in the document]
pauloiankoski: Leaving this as-is — it's not a gap specific to `addToken`, it's the same choice applied uniformly across every non-optimistic write flow in this PR: only success gets a Snackbar; failures stay on the inline `addError`/`saveError`/`deleteError`/`orderError` state the settings panel already renders next to the control that triggered them, since that's a more useful location for a validation-style failure than a transient toast. Routing `addToken`'s failure to the Snackbar too would make it the one write in the PR that behaves differently from its siblings.

src/style-library/store/resolvers.js
```diff
@@ -61,18 +61,29 @@ export const getPaletteListing =
...
+ *
+ * This app can have multiple independent `@wordpress/data` registries alive at once (see
+ * `store/test-utils.js`'s `createTestRegistry()` and how `style-library.js` mounts a fresh
+ * registry per app instance). A plain module-global `Map` would let a request in one registry
+ * bump the shared revision counter and wrongly discard an unrelated request in a different
+ * registry as "stale", even though the two have no ordering relationship. Keying a `WeakMap` on
+ * the thunk's `registry` object scopes the revision tracking to the registry it actually belongs
+ * to — `registry` is the one thing in `@wordpress/data`'s thunk context that is both stable and
+ * unique per registry instance (see `@wordpress/data`'s `redux-store/index.js`, where `registry`
+ * is closed over when the store is instantiated).
```
d4mation: This seems to overstate itself. 

> how `style-library.js` mounts a fresh registry per app instance

 The production application doesn't do this as it is a plain `createRoot(container).render(<StyleLibraryApp />)` with no `RegistryProvider` or `createRegistry`.
 
`notify.js:6-8` in this PR actually relies on the opposite of what this doc says to justify its default-registry `dispatch()`.

Using a `WeakMap` here is good for tests, but the doc here is otherwise incorrect based on the code.

src/style-library/helpers/palette-flows.js
```diff
@@ -638,59 +680,48 @@ export function addColorFlow({
...
-
-	if ((palette?.groups ?? []).some((group) => group.id === groupId)) {
-		const message = __('A color group with that name already exists.', 'kadence-blocks');
-		onError({ message });
-		return Promise.reject(new Error(message));
-	}
-
-	const colorSlug = nextCustomColorSlug(tokens);
-	const value = newSwatchValue(palette?.groups, groupId);
-	const swatchLabel = __('New Color', 'kadence-blocks');
```
d4mation, line 693: Is it expected for us to remove this client-side validation? Without it, `addGroupFlow` could end up posting an empty or duplicate `groupId`.

Arguably it is better for the server-side REST API to return an error for stuff like this anyway, but wanted to double-check as it is referenced in the doc for `validateNewGroupLabel()`

src/style-library/app/StyleLibraryApp.js
```diff
@@ -71,6 +73,15 @@ export function StyleLibraryApp() {
...
 	const libraries = useLibraries(feed.feed, feed.refreshFeed);
 
+	const snackbarNotices = useSelect(
+		(select) =>
+			select('core/notices')
+				.getNotices()
+				.filter((notice) => notice.type === 'snackbar'),
+		[]
+	);
+	const { removeNotice } = useDispatch('core/notices');
```
d4mation, line 83: Likely an edge case in practice, but `wp-notices` isn't technically registered, so if it weren't already loaded it could white screen the application. 

`import '@wordpress/notices'` in `notify.js` should ensure it is always loaded properly so if it suddenly _isn't_ loaded elsewhere, we don't run into issues.

src/style-library/components/pages/ColorPaletteScreen.js
```diff
@@ -283,12 +285,15 @@ export function ColorPaletteScreen({ label, route, navigate, library }) {
...
-								// Swallowed: a failure already lands in `structureError`, rendered above.
+								// Opens the new swatch's settings panel the moment it exists in the store as an
+								// optimistic addition, not after the write confirms — see `addColor`'s own
+								// `onOptimistic` docs.
+								.addColor(groupId, (newToken) => navigate({ item: newToken }))
+								// Swallowed: a failure already surfaces as a toast via `notifyError`.
 								.catch(() => {})
 						}
 						addLabel={__('Add color', 'kadence-blocks')}
+						addingGroupIds={palettes.addingGroupIds}
```
d4mation, line 296: This locks the add tile per group, but `use-palette'`s docblock for the shared flag says that a write to any palette in the library must block a write to any other palette in the same library from a sibling instance.

`addingGroupIds` only locks the group that is in-flight (`SwatchGrid.js:135`). Group B's tile stays live while A's add is running. 

The same gap exists on `onReorder`, but it doesn't check against anything at all. If you save a color edit from the settings panel, then drag a swatch while that's still in-flight, the reorder reads the palette _without_ the edit and its save then lands which writes the stale copy back to the swatch. Both `reorderSwatches` and `addColor` set the busy flag but never actually read it.

src/style-library/store/resolvers.js
pauloiankoski: Fixed as part of the phase-7 rebase merge — the docblock no longer claims a fresh registry per app instance and now correctly describes the shared, per-registry-scoped revision counter.

src/style-library/helpers/palette-flows.js
pauloiankoski: Restored client-side validation: empty groupId rejects before any network call, duplicate groupId is checked against the freshly-fetched palette inside the write flow itself. 7739979bb.

src/style-library/app/StyleLibraryApp.js
pauloiankoski: Added a side-effect `import '@wordpress/notices'` to notify.js so it's a declared script dependency instead of relying on load order. 51f69409e.

src/style-library/components/pages/ColorPaletteScreen.js
pauloiankoski: The shared busy flag is now actually read before every write in use-palettes.js (12 functions), not just written — a write on any palette blocks a write on any sibling. 10144ec46.

d4mation (approved):

PR #1327, opened 2026-08-21
SOFT-4205 [2/4]: lock the shipped swatches on the default palette write

d4mation:
🎫  https://linear.app/nexcess/issue/SOFT-4205/palette-swatch-deletion-lock-baseline-swatches-revert-instead-of

:video_camera:  https://www.loom.com/share/692dba0a66ba48c8a13c4ed53db90c78

The default palette is the structure template every other palette is projected from — the effective view renders each palette against the default's groups — so a swatch row removed there disappears from every palette at once, with nothing left to recover it from. Nothing stopped a write from doing that.

`update_item()` now guards a default-palette write against dropping a swatch the shipped baseline defines, answering **403 `rest_design_tokens_locked`** naming the token — the same shape `User_Primitives_Controller` uses to refuse a system primitive. A baseline swatch is reverted, never removed; that revert is the swatch DELETE route, in [3/5].

Scoped to the default palette on purpose: a non-default palette is stored as deltas, so it omits nearly every baseline token by design, and applying the guard there would reject every ordinary palette write. A swatch the baseline does **not** define stays deletable from the default palette too, which is what keeps a user-added color removable.

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Protected the default color palette from removing required built-in swatches.
  * Added clear forbidden errors identifying any locked swatch that is missing.
  * Continued allowing custom swatches to be added or removed from the default palette.
  * Non-default palettes can still omit built-in swatches.

* **Bug Fixes**
  * Prevented invalid default-palette changes from being saved.

### DIFF
--- a/includes/resources/Design_Tokens/Rest/V1/Palettes_Controller.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Palettes_Controller.php
@@ -37,6 +37,10 @@ use WP_REST_Server;
  * and `label` must be a string. Because a palette's values live under `$extensions`, the resolver dry-run
  * never sees them, so alias targets are guarded here explicitly.
  *
+ * One further guard is specific to the DEFAULT palette: it is the structure template every other palette is
+ * projected from, so a write against it may not drop a swatch the shipped baseline defines
+ * ({@see guard_baseline_swatches()}). Those rows are permanent — reverted, never removed.
+ *
  * @since TBD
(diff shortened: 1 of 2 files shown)


pauloiankoski (approved):
Approved. Just confirm if you wanna apply the fix proposed by CodeRabbit or we can resolve that.

PR #1330, opened 2026-08-21
SOFT-4205 [5/5]: split the swatch panel into Reset and Delete

d4mation:
🎫  https://linear.app/nexcess/issue/SOFT-4205/palette-swatch-deletion-lock-baseline-swatches-revert-instead-of

:video_camera:  https://www.loom.com/share/692dba0a66ba48c8a13c4ed53db90c78

> [!IMPORTANT]
> **This PR exists primarily so the behavior can be validated, and may be pulled out of the stack
> and closed without merging.** It wires the server work below it up to the Style Library so
> `@d4mation` can exercise the whole Reset/Delete flow in the UI and record a Loom of it — [1/5]–[4/5]
> are server-side only, so there is nothing to click through without this.
>
> `@pauloiankoski` has related changes in flight on another branch. Where the two overlap, that work
> is the one to accept — do not merge this PR without confirming which version of the Style Library
> changes is being taken.
>
> Dropping it costs the rest of the stack nothing. This is the only PR that touches `src/` and it
> sits at the top, so [1/5]–[4/5] stay intact and mergeable on their own. The guards and revert
> semantics below stand regardless of which UI drives them, and the REST contract this PR consumes
> (`baseline` / `overridden` on the effective view, and `DELETE /palettes/{id}/swatches/{token}`)
> lands in [4/5] and [3/5] — so the replacement work can build on it unchanged.

The Color Palette settings panel offered Delete on every swatch, and its handler rewrote the **default** palette node with the row stripped — whatever palette was open. Two problems in one: a swatch the shipped palette defines is not removable at all (its row is the shared structure every palette is projected from), and a delete inside a non-default palette took the color out of every palette instead of dropping that one palette's override.

The panel's destructive action now routes by where the swatch came from, mirroring the Delete/Reset split `ColorPaletteScreen` already makes at palette scope:

- A **baseline** swatch gets **Reset**, through the new `resetSwatchFlow`. It is the one flow besides `saveSwatchEditsFlow`'s color half that targets `editingId` rather than `defaultId`, and that single difference is what scopes the undo to the palette on screen: the shipped color comes back on the default palette, the delta is dropped (back to inherited) on any other. Offered only when there is something to undo, and it leaves the panel open on the row.
- A **user-added** swatch keeps **Delete** and the existing `removeSwatchFlow`, primitive cleanup and all.

This is the first caller of `deleteSwatch`, which the client has exported since the swatch route shipped without a single call site.

`baseline` rides through `mapPaletteToSwatchGroups` so the group overflow menu can gate on it too: a group holding any baseline swatch is no longer deletable, since group delete would otherwise do exactly what swatch delete now refuses. Absence rather than a disabled item, matching the guard already there for the last remaining group. Both the new `isDeletableGroup` and the hook's `isBaselineSwatch` fail **closed** like `helpers/token-capabilities`: an unflagged swatch reads as baseline, so a view that predates the flag is offered a reset, never a delete it cannot undo.

`SettingsPanel` grows `isReset` and `isDeleteDisabled` for the label and the enabled state; a Reset is also not rendered destructive, since it only undoes a value.

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added reset behavior for baseline swatches, while user-created swatches continue to be deleted.
  * Reset actions now refresh palette data and keep the settings panel open.
  * Added safeguards to prevent deletion of groups containing baseline swatches.
  * Improved action labels and loading states for reset and delete operations.

* **Bug Fixes**
  * Swatch and palette updates now correctly preserve baseline status and handle reset failures with appropriate error reporting.

### DIFF
--- a/src/style-library/__tests__/palette-flows.test.js
+++ b/src/style-library/__tests__/palette-flows.test.js
@@ -11,6 +11,7 @@ import {
 	renameGroupFlow,
 	renamePaletteFlow,
 	reorderSwatchesFlow,
+	resetSwatchFlow,
 	saveSwatchEditsFlow,
 	writeDefaultPaletteFlow,
 } from '../helpers/palette-flows';
@@ -24,6 +25,7 @@ import * as client from '../api/client';
 jest.mock('../api/client', () => ({
 	createUserPrimitive: jest.fn(),
 	deletePalette: jest.fn(),
+	deleteSwatch: jest.fn(),
 	deleteUserPrimitive: jest.fn(),
 	fetchPalette: jest.fn(),
 	savePalette: jest.fn(),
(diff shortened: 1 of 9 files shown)


d4mation:
@pauloiankoski You can ignore this one. I'm going to remove it from the Stack as your other changes should cover this I think. Regardless, it'd be easier to fix this on your branch if necessary.

PR #1332, opened 2026-08-21
[DTM] SOFT-4083 [36/38]: fix the border-width token's raw-id label fallback

pauloiankoski:
## Summary
- Fixes a display bug on the Style Library's Button preset screen: the Border field's width trigger showed the raw token id `semantic.border-width.default` instead of its real label ("Border Width").
- Root cause: `BorderField.js` called `pickableTokensForType('dimension', 'border-width')` without a `selected` argument, so the primitive-narrowing step (`preferPrimitiveTokens`) dropped the bound semantic token from the pickable list — every other field adapter (e.g. `BoxTokenField.js`) passes `selected` to exempt the currently-bound token from that narrowing, but this one didn't.
- Fix: pass the currently-bound width token id(s) as `selected`, reusing `BoxTokenField`'s existing `boundTokenIds()` helper (now exported) rather than duplicating its logic. Handles both a scalar width and a per-side unlinked width (all four bound tokens exempted).

## Test plan
- [x] `npx jest src/token-controls src/style-library src/extension/design-tokens` — full suite green
- [x] `bun run lint-js`, `bun run build`
- [x] cspell on changed files
- [x] `grep -rn "SOFT-" src/style-library` — no hits
- [ ] Manual verification on the dev site (Button preset screen's Border field width trigger)

## Summary by CodeRabbit

- **Bug Fixes**
  - Border width fields now keep currently selected semantic width tokens available after filtering.
  - Preserved token labels remain visible when narrowing primitive border-width options.

- **Tests**
  - Added coverage verifying semantic border-width token selection and label preservation.

### DIFF
--- a/src/style-library/__tests__/border-shadow-field-rendering.test.js
+++ b/src/style-library/__tests__/border-shadow-field-rendering.test.js
@@ -52,6 +52,12 @@ describe('BorderField', () => {
 		window[PICKABLE_TOKENS_GLOBAL] = {
 			tokens: [
 				{ id: 'primitive.dimension.border-width.sm', label: 'Small', type: 'dimension', role: 'border-width' },
+				{
+					id: 'semantic.border-width.default',
+					label: 'Border Width',
+					type: 'dimension',
+					role: 'border-width',
+				},
 				{ id: 'primitive.dimension.radius.sm', label: 'Radius Small', type: 'dimension', role: 'radius' },
 			],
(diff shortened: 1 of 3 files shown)


pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1334, opened 2026-08-22
[DTM] SOFT-4083 [29/38]: EditorBorderControl replaces the native border control on the canvas

pauloiankoski:
## Summary
- Adds `EditorBorderControl`, a thin block-editor wrapper around `BorderControl` that resolves/writes `kadence/singlebtn`'s native border attribute shape.
- Wires it into `singlebtn/edit.js` in place of the block's six `ResponsiveBorderControl` usages.
- Border color editing is wired back in through the same mechanism the native control used, unchanged in behavior.

## Test plan
- [x] Component tests for `EditorBorderControl`.
- [x] `npx jest` — full suite green.
- [ ] Manual verification on the dev site (block editor canvas, Style tab).

## Summary by CodeRabbit

* **New Features**
  * Added improved responsive border controls for the single-button block across desktop, tablet, and mobile views.
  * Supports border width presets, shared or individual border colors, and linked or unlinked side settings.
  * Border settings are available for normal, hover, transparent, and sticky button states.
  * Added support for preserving border design tokens while editing responsive settings.
* **Bug Fixes**
  * Preserves existing border colors, widths, styles, and units when adjusting responsive settings.

### DIFF
--- a/src/blocks/singlebtn/edit.js
+++ b/src/blocks/singlebtn/edit.js
@@ -26,7 +26,6 @@ import {
 	ResponsiveRangeControls,
 	IconRender,
 	HoverToggleControl,
-	ResponsiveBorderControl,
 	KadenceIconPicker,
 	KadencePanelBody,
 	URLInputControl,
@@ -94,6 +93,7 @@ import {
 } from '../../extension/token-indicators/normalize';
 import { TokenControlRow } from '../../extension/token-indicators/components/TokenControlRow';
 import { EditorBoxControl } from '../../extension/design-tokens/components/EditorBoxControl';
+import { EditorBorderControl } from '../../extension/design-tokens/components/EditorBorderControl';
 import { pickableTokensForControl } from '../../extension/token-picker';
 
 export default function KadenceButtonEdit(props) {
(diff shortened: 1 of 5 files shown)


pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai review

src/blocks/singlebtn/edit.js
```diff
@@ -1073,7 +1127,7 @@ export default function KadenceButtonEdit(props) {
 																/>
 															</TokenControlRow>
 														)}
-														<ResponsiveBorderControl
+														<EditorBorderControl
```
d4mation, line 1144: I can't get the Width picker to trigger. Is it the square-ish icon to the left of the color picker?

image

src/extension/design-tokens/components/EditorBorderControl.js
```diff
@@ -0,0 +1,250 @@
...
+		acc[side] = [
+			nextColor || existingColor,
+			readSlot(value.style, index) || 'none',
+			toNativeSize(readSlot(value.width, index), unit),
+		];
+		return acc;
+	}, {});
+
+	return [{ ...sides, unit }];
+}
```
d4mation: `nextColor || existingColor` treats `''` as "no update" and writes the old color back, which means that the border color cannot be cleared.

src/blocks/singlebtn/edit.js
pauloiankoski: All the area at the right side of the style selector is where we trigger the width selector. I've not faced a situation where we had nothing in there. I agree that is hard to realize there's a trigger in that area. Any ideas?

src/extension/design-tokens/components/EditorBorderControl.js
pauloiankoski: Confirmed and fixed. `toNativeBorder`'s `nextColor || existingColor` fallback is gone entirely — `value.color` is always authoritative (fromNativeBorder derives it from the real previous value, and BorderControl's patch() always merges the full current value on a width/style-only edit), so a genuine clear now writes `''` through instead of being silently reverted. Added a regression test covering the clear case.

src/blocks/singlebtn/edit.js
pauloiankoski: It's not the icon, it's that the width field had no visible content at all when unset (no default/fallback value shown, unlike Radius/Spacing), so it collapsed to zero height and looked like it wasn't there. Fixed by threading the active preset's border width through as a muted default, matching how the other dimension fields already show "Default"/"Inherited" — an unset width row now reads e.g. "SM 1px" muted instead of rendering nothing.

CleanShot 2026-08-24 at 18 52 01

d4mation: That seems to fix it! But one thing it exposes now is that the "Linked" button doesn't seem to collapse it down to a single field like the Border Radius one, for instance. Based on the CSS that would show a little preview icon of all four sides that I saw, I figure this is not expected, correct?

I think it would make sense to fix weird stuff like this in its own PR, so I'll approve this one and we can follow up on it.

d4mation (approved):

src/blocks/singlebtn/edit.js
pauloiankoski: Confirmed — not expected. Opened stellarwp/kadence-blocks#1345 to fix it: the Linked toggle now owns its own view state instead of trying to read it off the stored shape (which is always four sides once written, so it could never tell "every side matches" from "the user is editing them individually").

PR #1336, opened 2026-08-22
[DTM] SOFT-4083 [31/38]: render border and shadow from the selected preset

pauloiankoski:
## Summary
- Adds `render_preset_border()`/`render_preset_shadow()` (PHP) and `presetBorderProperties()`/`presetShadowProperties()` (JS), mirroring the existing `render_preset_spacing()`/`presetSpacingProperties()` pattern, so a selected (named, non-default) preset's border and shadow values actually reach rendered CSS on both the front end and the block-editor canvas preview — the same way padding and margin already work.
- An explicit per-instance override still wins over the preset's value, by CSS declaration order (preset emitted first, explicit attribute output after, in the same rule).

## Test plan
- [x] wpunit coverage proving a named, non-default preset changes emitted CSS, and that an explicit per-instance override still wins (declaration order asserted directly).
- [x] `npx jest` — full suite green.
- [ ] Manual verification on the dev site: two Button presets with different border widths render differently, and overriding one specific block instance's border still wins over its own preset.

## Summary by CodeRabbit

* **New Features**
  * Single-button presets now support border width, style, color, and shadow settings.
  * Preset styling is applied automatically when configured.
  * Explicit button border and shadow styles take precedence over preset values.
  * Missing or incomplete preset properties are handled without affecting existing styles.

* **Tests**
  * Added coverage for preset-driven border and shadow styling, fallback behavior, CSS output, and style precedence.

### DIFF
--- a/includes/blocks/class-kadence-blocks-singlebtn-block.php
+++ b/includes/blocks/class-kadence-blocks-singlebtn-block.php
@@ -127,10 +127,12 @@ class Kadence_Blocks_Singlebtn_Block extends Kadence_Blocks_Abstract_Block {
 		}
 		$css->render_typography( $attributes, 'typography' );
 		$css->render_measure_output( $attributes, 'borderRadius', 'border-radius', [ 'unit_key' => 'borderRadiusUnit' ] );
+		$this->render_preset_border( $css, $attributes );
 		$css->render_border_styles( $attributes, 'borderStyle', true );
 		$this->render_preset_spacing( $css, $attributes );
 		$css->render_measure_output( $attributes, 'padding', 'padding', [ 'unit_key' => 'paddingUnit' ] );
 		$css->render_measure_output( $attributes, 'margin', 'margin', [ 'unit_key' => 'marginUnit' ] );
+		$this->render_preset_shadow( $css, $attributes );
 		if ( isset( $attributes['displayShadow'] ) && true === $attributes['displayShadow'] ) {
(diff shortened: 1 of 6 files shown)


pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai review

includes/blocks/class-kadence-blocks-singlebtn-block.php
```diff
@@ -542,6 +544,118 @@ private function render_preset_spacing( Kadence_Blocks_CSS $css, array $attribut
...
+	 * border, which lands later in the same rule, still wins.
+	 *
+	 * @since TBD
+	 *
+	 * @param Kadence_Blocks_CSS   $css        The css object.
+	 * @param array<string, mixed> $attributes The block attributes.
+	 *
+	 * @return void
+	 */
+	private function render_preset_border( Kadence_Blocks_CSS $css, array $attributes ): void {
```
d4mation, line 565: At some point we may want to move `render_preset_spacing()`, `render_preset_border()`, and `render_preset_shadow()` to the `Kadence_Blocks_Abstract_Block` class so that they can be reused by other Blocks later. 

Nothing else can access these `private` ones, so we can remove/move them whenever though, so no need to rush it right now.

d4mation (approved):

PR #1346, opened 2026-08-25
[DTM] SOFT-4083 [40/40]: collapse the Border field's Linked view when every side matches

pauloiankoski:
Supersedes #1345, which auto-closed when its branch was renamed.

## Summary
- Follow-up to review feedback on #1334: `EditorBorderControl`'s Linked toggle never collapsed the Border field to one row, because the native attribute always stores four sides once written, so `BorderControl`'s own shape-based fallback (`isSlotList`) could never tell "every side happens to match" from "the user is editing them individually."
- `EditorBorderControl` now owns the linked/unlinked view as its own UI state, seeded from whether the stored sides currently agree and overridden per breakpoint once the user explicitly links or unlinks — the same pairing `singlebtn/edit.js`'s `borderRadiusModeOverride` already uses for Border Radius.
- Relinking collapses every side to the top side's value, matching `BorderControl`'s own uncontrolled relink rule ("the first side wins").

## Test plan
- [ ] Open the Button preset's Border field with all four sides matching and confirm it renders as one row.
- [ ] Click Unlink and confirm it expands into four independently-editable rows without changing the stored value.
- [ ] Diverge one side's width, style, or color, then click Link and confirm every side collapses to the top side's value and the field returns to one row.
- [ ] Switch breakpoints (Desktop/Tablet/Mobile) and confirm each one remembers its own linked/unlinked choice independently.

### DIFF
--- a/src/extension/design-tokens/components/EditorBorderControl.js
+++ b/src/extension/design-tokens/components/EditorBorderControl.js
@@ -13,8 +13,14 @@
  * - **once a side has ever been written, it stays a four-element array**, never collapsed back to a
  *   scalar — `fromNativeBorder` only answers the empty scalar (`''`/`'none'`) for a breakpoint that
  *   has never been saved at all; the moment any side is written, `toNativeBorder` always fills in
- *   all four, so every later read comes back as four-element width/style/color arrays and the
- *   control renders unlinked from then on. There is no separate "link" flag to keep in sync.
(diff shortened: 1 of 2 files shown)


pauloiankoski:
Renumbering branch to match phase sequence — replaced by a new PR from SOFT-4083/phase-33-editor-border-control-linked-state.

PR #1349, opened 2026-08-25
[DTM] SOFT-4083 [40/51]: collapse the Border field's Linked view when every side matches

pauloiankoski:
🎥  https://www.loom.com/share/6978ded2beb44d308c51e9acdabc43eb

Supersedes #1346, renumbered to keep the branch sequence at phase-33 (following phase-32).

## Summary
- Follow-up to review feedback on #1334: `EditorBorderControl`'s Linked toggle never collapsed the Border field to one row, because the native attribute always stores four sides once written, so `BorderControl`'s own shape-based fallback (`isSlotList`) could never tell "every side happens to match" from "the user is editing them individually."
- `EditorBorderControl` now owns the linked/unlinked view as its own UI state, seeded from whether the stored sides currently agree and overridden per breakpoint once the user explicitly links or unlinks — the same pairing `singlebtn/edit.js`'s `borderRadiusModeOverride` already uses for Border Radius.
- Relinking collapses every side to the top side's value, matching `BorderControl`'s own uncontrolled relink rule ("the first side wins").

## Test plan
- [ ] Open the Button preset's Border field with all four sides matching and confirm it renders as one row.
- [ ] Click Unlink and confirm it expands into four independently-editable rows without changing the stored value.
- [ ] Diverge one side's width, style, or color, then click Link and confirm every side collapses to the top side's value and the field returns to one row.
- [ ] Switch breakpoints (Desktop/Tablet/Mobile) and confirm each one remembers its own linked/unlinked choice independently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved border controls so linked and unlinked states are tracked independently for each breakpoint.
  * Added reliable detection of initially linked borders when all sides share matching values.
  * Relinking a breakpoint now consistently applies the first side’s settings across all sides.
  * Unlinking and relinking selections now persist correctly during editing and screen-size changes.

### DIFF
--- a/src/extension/design-tokens/components/EditorBorderControl.js
+++ b/src/extension/design-tokens/components/EditorBorderControl.js
@@ -13,8 +13,14 @@
  * - **once a side has ever been written, it stays a four-element array**, never collapsed back to a
  *   scalar — `fromNativeBorder` only answers the empty scalar (`''`/`'none'`) for a breakpoint that
  *   has never been saved at all; the moment any side is written, `toNativeBorder` always fills in
- *   all four, so every later read comes back as four-element width/style/color arrays and the
- *   control renders unlinked from then on. There is no separate "link" flag to keep in sync.
+ *   all four. So the linked/unlinked view cannot be read off storage the way `BorderControl`'s own
+ *   uncontrolled fallback tries to (it only checks "is this an array", not "do the four elements
+ *   agree") — this component owns that view as UI state instead, seeded from whether the stored
+ *   sides currently agree (`isUniformBorder`) and overridden per breakpoint once the user explicitly
+ *   links or unlinks. This mirrors `singlebtn/edit.js`'s own `borderRadiusModeOverride` for radius,
+ *   just owned locally: border's `defaultValue` is always one scalar for every side (see its own
+ *   docblock below), so there is no per-side preset difference an early "uniform" read could hide
+ *   the way an empty radius corner could.
  * - **the unit lives inside the native value itself** (`source.unit`), not a sibling attribute the
  *   way `BoxControl`'s radius keeps `borderRadiusUnit` — so, unlike `EditorBoxControl`, this
  *   component takes no separate `unit`/`units`/`onUnit` props at all.
@@ -25,6 +31,11 @@
  * still has to fall back to the native value's stored color when `renderColor` writes nothing for a
  * side — see its docblock for why.
  */
+
+/**
+ * WordPress dependencies
+ */
+import { useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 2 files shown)


src/extension/design-tokens/components/EditorBorderControl.js
```diff
@@ -230,6 +270,36 @@ export function EditorBorderControl({
...
+				{
+					width: readSlot(current.width, 0),
+					style: readSlot(current.style, 0),
+					color: readSlot(current.color, 0),
+				},
+				activeUnit
+			)
+		);
+		setLinkOverride((next) => ({ ...next, [breakpoint]: true }));
+	};
```
d4mation, line 306: On re-link, `toggleLink` sets `linkOverride[breakpoint]` to `true`, and it never goes back to `undefined`. The unlink branch only flips it to `false`. So once you re-link on a given breakpoint, that breakpoint stops deriving its linked/individual mode from the stored value for the rest of the session, and `isUniformBorder` is never consulted again.

The comment above points at `singlebtn/edit.js`'s `borderRadiusModeOverride`, which does the opposite on both counts: its re-link writes `undefined` (`edit.js:500-503`, forcing `'linked'` only for the blank-corner
case), and a `useEffect` clears the whole map on preset change (`edit.js:514-516`). Border has neither.

Reproduction:

1. Add an Advanced Button block. Style tab, Button Styles panel, Normal state, Border field (Desktop).
2. Unlink the field and set four different widths: top 1, right 2, bottom 3, left 4.
3. Click the link toggle. All four collapse to 1px, and `linkOverride.desktop` is pinned `true`.
4. Undo until the canvas button renders 1/2/3/4 again.
5. **Bug:** the Border field still shows one linked row reading 1px. The three divergent sides are hidden. Expected: `isUniformBorder` re-derives `false` and the four rows reopen.
6. **Data loss:** change that linked row's width to 5px. Linked mode calls `renderSlot` with `index: null`, so `applyToAxis` returns a bare scalar and `toNativeBorder`'s `readSlot(scalar, i)` writes it to all four sides. The 2/3/4 restored by the undo are gone, with nothing on screen indicating they were ever there.

It is scoped per breakpoint and per mount, so reselecting the block or switching device clears it which is likely why it has not turned up in testing. It repeats across all six button states.

Fix looks like one line: write `undefined` instead of `true`, or `true` only when `readSlot(current.width, 0)` is blank, matching the equal-but-empty case radius handles at `edit.js:502`.

pauloiankoski: On relink, `linkOverride[breakpoint]` now writes `undefined` instead of `true`, so it stops forcing linked once a later value (undo, a divergent write) no longer agrees. `isUniformBorder` derives the linked view from the real value on the next render, the same way it already does for a border that has never been touched. Added a test asserting a still-diverging value after relink reports unlinked, and updated the existing relink test to re-render with the collapsed value rather than assert an instant flip against a mock that never feeds props back.

Fixed in 25f4f4ac1.

d4mation (approved):

PR #1352, opened 2026-08-25
[DTM] SOFT-4207 [8/15]: a Favorites/Custom font-family picker

d4mation:
🎫  https://linear.app/nexcess/issue/SOFT-4207/font-families-become-favorites-not-tokens
:video_camera:  https://www.loom.com/share/2cde91916171494694080f3b7beda6e9

The shared field a host opens in place of a font select: a `Favorites` tab listing the families a site pinned, and a `Custom` tab searching the whole catalog behind a focused search input and the shared render cap.

A sibling of `TokenPopover` rather than a mode of it. The shape reads the same on screen, but what it picks is not a token: a favorite carries no alias, resolves through nothing, and emits no CSS variable. Both tabs write the same plain family string, so the control takes no token pool and never resolves. Folding this into `TokenPopover` would mean teaching the token picker to sometimes not pick tokens, which is the confusion dropping the font-family token layer was meant to end.

What it does borrow is the anatomy and the open-on-the-curated-list rule: an unset or favorited family opens on `Favorites`, anything else on `Custom`, which is the tab it was picked from and the only one that can show it in context.

Nothing renders it yet.

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a font-family selector with favorites and custom catalog views.
  * Added searchable font options with badges, empty states, and truncation messaging.
  * Added support for resetting selections and managing favorite fonts through an optional link.
  * Added inherited theme-default labeling and disabled-state support.
* **Style**
  * Improved catalog layout, scrolling, search field, option list, and footer presentation.


pauloiankoski (approved):
`isFavorite` (in `FontFamilySelector.js`) matches favorites with case-sensitive `===,` same gap as #1355. If a stored favorite differs in case from the set value, `initialTab` picks `custom` and the favorites row won't show as active. Not blocking, just tracking alongside #1355 so it's fixed once, not twice.

d4mation:
@pauloiankoski Addressed in [7141e67](https://github.com/stellarwp/kadence-blocks/pull/1352/commits/7141e672c3aa7fba9c378b92b85b6907a80e88c3)

PR #1373, opened 2026-08-26
[DTM] SOFT-4083 [48/51]: extract the 7 linked-state blocks into a useLinkedMeasureState hook

pauloiankoski:
Extracts the 7 duplicated linked-state blocks across Border Radius, Padding, and Margin into one shared `useLinkedMeasureState` hook.

## Summary by CodeRabbit

* **New Features**
  * Added consistent linked and individual controls for button border radius and padding.
  * Preserved responsive device-specific settings when switching between linked and individual modes.
  * Improved handling of inherited values, token aliases, and differing responsive values.

* **Bug Fixes**
  * Reset temporary overrides when presets or related settings change.
  * Prevented unintended changes to stored values when toggling linked mode.

### DIFF
--- a/src/blocks/singlebtn/edit.js
+++ b/src/blocks/singlebtn/edit.js
@@ -80,11 +80,11 @@ import {
 } from '@wordpress/components';
 import { addFilter, applyFilters, doAction } from '@wordpress/hooks';
 import BackendStyles from './components/backend-styles';
+import { useLinkedMeasureState } from './hooks/use-linked-measure-state';
 import { PresetButton } from '../../extension/preset-picker/PresetButton';
 import { usePresetBinding, resetAttr, presetPropertyValueForDevice } from '../../extension/token-indicators';
 import {
 	anyCornerInherited,
-	deriveMeasureMode,
 	inheritedMeasureSlots,
 	measureAttrsForDevice,
 	presetValueForDevice,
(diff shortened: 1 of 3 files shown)


pauloiankoski:
@coderabbitai review

src/blocks/singlebtn/edit.js
```diff
@@ -448,145 +438,42 @@ export default function KadenceButtonEdit(props) {
...
+	// `inheritedPadding`/`paddingPickableTokens` instead. Padding has no `resetOn`: unlike Border
+	// Radius's 6 states, its override has never been cleared on a preset change.
+	const { isLinked: paddingIsLinked, toggleLink: togglePaddingLink } = useLinkedMeasureState({
+		forDevice: paddingForDevice,
+		inherited: inheritedPadding,
+		previewDevice,
+		presetValue: paddingPresetValue,
+		tokens: paddingPickableTokens,
+		setAttributes,
+	});
```
d4mation, line 472: The Padding call omits `resetOn`, which drops a fix made two PRs down previously.

Before this refactor, the preset-change effect cleared all seven overrides together. The six border-radius maps and padding's (`edit.js:904-911` on the parent commit). `setPaddingModeOverride({})` on line 910 came from #1371's `3dfe442e4` ("clear the padding link-mode override when the preset changes"), which was itself the fix for a review comment on that PR.

The extraction moves the six radius states onto `resetOn: attributes.kbPreset` (`edit.js:456`, `:495`, `:517`, `:539`, `:560`, `:582`) but leaves it off Padding. Per the hook's own docblock, omitting it means "never auto-clear" (`use-linked-measure-state.js:50-53`), so nothing runs the `useEffect` at `:74-76` for padding and the override now survives a preset switch. Nothing else in the file clears it and the old effect is gone.

The comment on lines 461-462 states padding's override "has never been cleared on a preset change," which is what `3dfe442e4` did.

Reproduction:

1. Select an Advanced Button. Style tab, Button Styles panel, Normal state, Padding field.
2. Leave Padding unset so it inherits the preset, then click the link toggle to unlink. That records `'individual'` for this device (`use-linked-measure-state.js:134`).
3. Switch to a different preset whose padding differs from the first.
4. **Bug:** Padding stays in individual mode instead of deriving the new preset's. Border Radius, given the same sequence, snaps back correctly.

Fix is to add `resetOn: attributes.kbPreset` on line 469 and drop the comment claiming padding never had it.

pauloiankoski: Added `resetOn: attributes.kbPreset` to Padding's `useLinkedMeasureState` call and dropped the comment claiming it never had one — matches Border Radius's six states now.

Fixed in a6018e90f.

d4mation (approved):

PR #1376, opened 2026-08-26
[DTM] SOFT-4083 [51/51]: guard the token-indicators store against duplicate registration

pauloiankoski:
🎥  https://www.loom.com/share/685d696d65a14c9faed50cfabf528cec

Guards the `kadence/token-indicators` store's module-load registration against duplicate bundles registering it twice, and excludes `.worktrees/` from Jest's test-path discovery so a sibling worktree checkout doesn't get scanned twice.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate token-indicators store registration when multiple bundles are loaded.
  * Improved stability when the store is already registered.

* **Tests**
  * Added coverage for both new and existing store registration scenarios.

* **Chores**
  * Updated test discovery to ignore worktree directories.
  * Added “worktrees” to the spell-check dictionary.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -27,10 +27,18 @@ const baseConfig = require('@wordpress/scripts/config/jest-unit.config.js');
 
 module.exports = {
 	...baseConfig,
-	// `.worktrees/` holds sibling checkouts of this same repo (see `using-git-worktrees`); without
-	// this, jest walks into them too and runs every test a second time against a possibly different
-	// branch's source.
-	testPathIgnorePatterns: [...(baseConfig.testPathIgnorePatterns || []), '/\\.worktrees/'],
+	// `@wordpress/scripts/config/jest-unit.config.js` exports only `{ preset, reporters, transform }`,
+	// so `baseConfig.testPathIgnorePatterns` is always undefined and a spread of it contributes
+	// nothing — `testPathIgnorePatterns` is not one of the keys jest merges across preset and project
+	// (jest-config's normalize.js special-cases `setupFilesAfterEnv`, `modulePathIgnorePatterns`,
+	// `moduleNameMapper`, `transform`, and `globals`; everything else falls through to
+	// `{...preset, ...options}`, so the project value replaces the preset's outright). Written out
+	// literally so the list says what it actually is: the preset's own ignores
+	// (`@wordpress/jest-preset-default`'s `['/node_modules/', '<rootDir>/vendor/']`) plus this repo's
+	// own `.worktrees/` — `.worktrees/` holds sibling checkouts of this same repo (see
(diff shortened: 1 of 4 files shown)


pauloiankoski:
@coderabbitai review

jest.config.js
```diff
@@ -27,6 +27,10 @@ const baseConfig = require('@wordpress/scripts/config/jest-unit.config.js');
 
 module.exports = {
 	...baseConfig,
+	// `.worktrees/` holds sibling checkouts of this same repo (see `using-git-worktrees`); without
+	// this, jest walks into them too and runs every test a second time against a possibly different
+	// branch's source.
+	testPathIgnorePatterns: [...(baseConfig.testPathIgnorePatterns || []), '/\\.worktrees/'],
```
d4mation: Nit: the spread here is dead, and it silently drops the preset's own ignores.

`@wordpress/scripts/config/jest-unit.config.js` exports only `{ preset, reporters, transform }`, so `baseConfig.testPathIgnorePatterns` is `undefined` and the spread always contributes nothing. The preset's values (`['/node_modules/', '<rootDir>/vendor/']`) live on `@wordpress/jest-preset-default`, and `testPathIgnorePatterns` is not one of the keys jest merges across preset and project. `jest-config/build/normalize.js` special-cases `setupFilesAfterEnv`, `modulePathIgnorePatterns`, `moduleNameMapper`, `transform`, and `globals`; this one falls through to `{...preset, ...options}`. So the project value replaces the preset's outright:

    $ npx jest --showConfig | grep -A2 testPathIgnorePatterns
    "testPathIgnorePatterns": ["/\\.worktrees/"]

Nothing breaks right now. `--listTests` gives the same 80 suites with or without the preset entries restored, since jest-haste-map does not crawl `node_modules` regardless and `vendor/` currently holds no JS matching `testMatch`. It just means `vendor/` is only excluded by accident, and would start being crawled the moment a composer package ships a `__tests__` directory.

Worth writing the list literally so it says what it means:

    testPathIgnorePatterns: ['/node_modules/', '<rootDir>/vendor/', '/\\.worktrees/'],

The identical spread on `moduleNameMapper` below is equally dead but harmless, since jest does merge that one with the preset (`normalize.js:285`).

pauloiankoski: Rewrote `testPathIgnorePatterns` as a literal array: `['/node_modules/', '<rootDir>/vendor/', '/\\.worktrees/']`. Verified with `npx jest --showConfig | grep -A2 testPathIgnorePatterns` — resolved value is unchanged. Left `moduleNameMapper`'s spread as is since jest does merge that key.

Fixed in 1eedf501e.

d4mation (approved):

PR #1377, opened 2026-08-26
[DTM] SOFT-4167 [1/5]: stop filling per-corner gaps on a responsive override capture

pauloiankoski:
The preset-picker's `attrToLiteral()` no longer fills empty corners when capturing a responsive breakpoint override for a per-corner dimension value — an untouched corner is now written as a real gap (`''`) instead of being broadcast the touched corner's value. The base-value capture path is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Responsive border-radius settings now preserve intentionally unset corners as empty gaps at breakpoint overrides.
  * Breakpoint-specific values no longer incorrectly inherit corner values from base or earlier breakpoint settings.
  * Base preset fallback behavior remains supported where applicable.

### DIFF
--- a/src/extension/preset-picker/__tests__/capture.test.js
+++ b/src/extension/preset-picker/__tests__/capture.test.js
@@ -225,7 +225,7 @@ describe('capturedTokens', () => {
 		).toEqual({ mobile: '2px' });
 	});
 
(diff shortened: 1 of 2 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1378, opened 2026-08-26
[DTM] SOFT-4167 [2/5]: allow a gap slot only inside a responsive breakpoint override

pauloiankoski:
Adds a REST write guard (`Presets_Controller::guard_base_slot_gap()`) and a schema-level `$allow_gap` parameter (`Dtcg_Validator`) so an empty-string gap slot is legal only inside a responsive-override breakpoint, never in a base value.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for four-corner dimension presets.
  * Base values containing empty slots or spaces are now rejected with a validation error.
  * Responsive overrides may contain empty slots and are preserved correctly.
  * Responsive overrides now require a compatible four-corner base value.

### DIFF
--- a/includes/resources/Design_Tokens/Rest/V1/Presets_Controller.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Presets_Controller.php
@@ -1289,6 +1289,12 @@ final class Presets_Controller extends Controller {
 						return $error;
 					}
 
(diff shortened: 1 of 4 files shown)


pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1383, opened 2026-08-26
[DTM] SOFT-4211: reject a non-array $extensions metadata section

d4mation:
🎫  https://linear.app/nexcess/issue/SOFT-4211/dtcg-validator-reject-non-array-dollarextensions-metadata-sections

All four `$extensions.com.kadence.designTokens` metadata sections — `colorPalettes`, `tokenLabels`,
`tokenOrder`, `favoriteFonts` — were reached through a guard that required the stored value to
already be an array before it would run the section's validator. Every one of them contracts for an
array, so that guard let exactly the values that are wrong skip the check: a string, a number, a
bool or a null stored under any of those keys was reported as a valid document and accepted on
write.

Nothing broke downstream — the document indexes fail soft and degrade a malformed section to an
empty result — but a bad write was accepted silently instead of coming back as a `Validation_Error`
a caller could surface.

The guard now only asks whether the key is present, and each section validator takes its value
unshaped and reports the section itself as invalid when it is not an array. All four together:
tightening one would leave three siblings behind with nothing to explain the difference.

The fail-soft behavior in the document indexes is unchanged — this is about the validator reporting
a malformed section, not about how readers cope with one.

Based on the tip of the SOFT-4207 stack, because `favoriteFonts` is one of the four sections and
only exists there. Not part of that stack.

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved design-token validation for extension sections.
  * Invalid scalar and null values now produce clear validation errors instead of being skipped or rejected prematurely.
  * Valid and empty arrays continue to be accepted.

* **Tests**
  * Added comprehensive coverage for invalid and valid extension-section values.

### DIFF
--- a/includes/resources/Design_Tokens/Schema/Validation/Dtcg_Validator.php
+++ b/includes/resources/Design_Tokens/Schema/Validation/Dtcg_Validator.php
@@ -611,7 +611,7 @@ final class Dtcg_Validator {
 		// each swatch `$value` with the same alias-or-literal grammar here.
 		$palettes_section = Extensions::get_section_color_palettes();
 
-		if ( isset( $namespace[ $palettes_section ] ) && is_array( $namespace[ $palettes_section ] ) ) {
+		if ( array_key_exists( $palettes_section, $namespace ) ) {
 			$errors = array_merge(
 				$errors,
 				$this->validate_color_palettes( $namespace[ $palettes_section ], $base . '.' . $palettes_section )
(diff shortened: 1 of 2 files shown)


pauloiankoski (approved):

d4mation:
@pauloiankoski Somehow, the merge of the last few got messed up. I've fixed that and ensured the base was pointed correctly. No change to the PR itself otherwise, so I'm merging this now.

PR #1386, opened 2026-08-27
fix(box-control): stop a slot's picker from offering its siblings' own bound tokens

pauloiankoski:
`BoxTokenField`'s (and any future caller's) `boundTokenIds()` exempts every currently-bound token from the primitives-only narrowing so no corner's own binding silently disappears from the shared pool — but that exemption is computed once for all four corners together. The result: editing one corner's picker (e.g. Padding Top) also lists the other corners' own bound tokens (Padding Right/Bottom/Left) as pickable options, even though those are specific to corners the user isn't editing.

Fixes it at the shared `BoxControl` level (used by both the Style Library's `BoxTokenField` and the block editor's `EditorBoxControl`): each slot's picker now filters the shared pool down, dropping any token that's a *sibling* slot's own current value while always keeping a token that matches the slot's own value.

Found while recording review demos for the SOFT-4083 stack — unrelated to that ticket's own scope, so it's a standalone fix rather than another phase in that chain.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -187,6 +187,7 @@
         "webp",
         "woocommerce",
         "woodardmc",
+        "worktrees",
         "wpbody",
         "wpcontent",
         "wpdb",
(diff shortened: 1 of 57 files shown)


pauloiankoski:
Closed in favor of https://github.com/stellarwp/kadence-blocks/pull/1419 chain

PR #1420, opened 2026-08-28
[DTM] SOFT-4196 [6/8]: add a minimal editor-scoped color-groups fetch hook for ColorControl

pauloiankoski:
## Summary
- The block editor has no existing data source for `ColorControl`'s grouped `groups` prop — it only has `window.kadenceDesignTokensPalettes`, a flat slot→alias map, not the Accent/Contrast/Background/Notices grouped shape.
- Adds `useColorGroups(clientId)` (`src/extension/design-tokens/hooks/use-color-groups.js`): resolves the block's effective palette (its own `kbPalette` override, else the nearest ancestor's, else the site's current palette, via the existing `effectivePalette()`), fetches that palette's groups from the `Palettes_Controller` REST route via `@wordpress/api-fetch`, and shapes the response through `mapPaletteToColorControlGroups()` from the previous slices.
- Cached per palette id in a module-level `Map` (cleared on rejection so a transient failure doesn't strand every future instance) — this is read-only, rarely-changing data, so no store/reducer is needed.
- Still not wired into any block — the next slice consumes this hook in `kadence/singlebtn`'s inspector.

## Test plan
- `npx jest src/extension/design-tokens` — all green, no regressions.
- New tests: resolves the block's own palette override vs. falling back to the site's current palette, handles the not-yet-loaded state, fails open on a rejected fetch, and reuses the cache across instances sharing a palette id.

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving grouped color swatches based on the active design palette.
  * Automatically uses block-specific palette settings or site defaults.
  * Refreshes color groups when the selected client changes.
  * Provides graceful fallback behavior when color data is unavailable or cannot be loaded.

* **Tests**
  * Added coverage for palette resolution, loading states, errors, caching, and mapped color-group results.

### DIFF
--- a/src/extension/design-tokens/hooks/use-color-groups.js
+++ b/src/extension/design-tokens/hooks/use-color-groups.js
@@ -78,26 +78,31 @@ function resolveEffectivePaletteId() {
 
 /**
  * Fetch one palette's node and shape it into `ColorControl`'s `groups` prop, reusing an in-flight or
- * already-resolved request for the same palette id.
+ * already-resolved request for the same palette id AND library — the REST route's response depends on
+ * both (the same palette id can exist in more than one library), so caching on the id alone could
+ * return one library's groups for another.
  *
  * @param {string} paletteId The palette id to fetch.
+ * @param {string} library   The library slug the palette id is resolved against.
  *
  * @since TBD
  *
  * @return {Promise<Array>} The mapped groups, or an empty array when the registry is inactive or the
  *         fetch fails.
  */
-function fetchColorGroups(paletteId) {
+function fetchColorGroups(paletteId, library) {
 	if (!hasDesignTokensRest() || !paletteId) {
 		return Promise.resolve([]);
 	}
 
-	if (groupsCache.has(paletteId)) {
-		return groupsCache.get(paletteId);
+	const cacheKey = `${library}:${paletteId}`;
+
+	if (groupsCache.has(cacheKey)) {
+		return groupsCache.get(cacheKey);
 	}
 
 	const path = `/${designTokensNamespace()}/palettes/${encodeURIComponent(paletteId)}?library=${encodeURIComponent(
-		librarySlug()
+		library
 	)}`;
 
 	const request = apiFetch({ path })
@@ -105,12 +110,12 @@ function fetchColorGroups(paletteId) {
 		.catch(() => {
 			// Do not cache a failure — a transient error (e.g. a slow first-load auth race) should not
 			// permanently strand every future ColorControl instance on an empty list.
-			groupsCache.delete(paletteId);
+			groupsCache.delete(cacheKey);
 
 			return [];
 		});
 
-	groupsCache.set(paletteId, request);
+	groupsCache.set(cacheKey, request);
 
 	return request;
 }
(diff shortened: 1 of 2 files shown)


pauloiankoski:
@coderabbitai review

src/extension/design-tokens/hooks/use-color-groups.js
```diff
@@ -0,0 +1,150 @@
...
+		.then((palette) => mapPaletteToColorControlGroups(palette))
+		.catch(() => {
+			// Do not cache a failure — a transient error (e.g. a slow first-load auth race) should not
+			// permanently strand every future ColorControl instance on an empty list.
+			groupsCache.delete(paletteId);
+
+			return [];
+		});
+
+	groupsCache.set(paletteId, request);
```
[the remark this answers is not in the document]
pauloiankoski: Fixed both. `paletteId`/`library` are now derived on every render (not just inside the effect) and included in the effect's dependency array alongside `clientId`, so a `kbPalette` change on the same selected block re-fetches. The cache key is now `${library}:${paletteId}`, and `groups` resets to `[]` immediately on an identity change rather than briefly showing the previous palette's groups. Added regression tests for both: re-fetching on a palette-id change with no `clientId` change, and fetching separately (not from cache) for the same palette id in a different library.

src/extension/design-tokens/hooks/use-color-groups.js
```diff
@@ -0,0 +1,165 @@
...
+		fetchColorGroups(paletteId, library).then((resolved) => {
+			if (!cancelled) {
+				setGroups(resolved);
+			}
+		});
+
+		return () => {
+			cancelled = true;
+		};
+	}, [clientId, paletteId, library]);
```
d4mation, line 285: This effect's only re-run trigger is `[clientId, paletteId, library]`, and every not-ready condition resolves to `[]` — so the first attempt latches an empty list for the rest of the session. If `hasDesignTokensRest()` is false at mount, or the id falls back to `'default'` and 404s, or `effectivePalette()` returns `''` because nothing is selected yet, `setGroups([])` runs and nothing ever fires the effect again.

In practice that means a block's ColorControl opens with an empty Style Library tab and stays empty until you change the block's palette, which is the one action that changes `paletteId` and re-triggers the fetch. And if you only have the Default palette, it isn't possible to change the palette to correct this at all.

The `.catch()` above deletes the cache entry so a later fetch can retry, but nothing ever triggers a later fetch for that key, so the mitigation is inert for an already-mounted hook.

Worth noting the old path had no fetch at all — `register-color-control-filters.js` reads `window.kadenceDesignTokensPalettes.slots` synchronously, so it was never empty on first paint. A fix wants either a retry or a reactive palette read (`useSelect`), so that "not ready yet" and "genuinely empty" stop resolving to the same `[]`.

image

d4mation: Actually, turns out this was caused by my Button Blocks having a stored `kbPalette` that referenced a deleted Palette. This breaks the interface and could be a true failure case for users. Going to make it fallback to the active palette in these cases, as that's how the rest of the projection-related code works anyhow.

d4mation: Addressed a possible issue in [2a037c5](https://github.com/stellarwp/kadence-blocks/pull/1420/commits/2a037c51d243d9990535d101c1d0e303e98e0ae2), but fixed what I actually was encountering in [7326e4a](https://github.com/stellarwp/kadence-blocks/pull/1420/commits/7326e4abc7d62103e65972527809eea0f5a478da)

d4mation (approved):

PR #1421, opened 2026-08-28
[DTM] SOFT-4260 [2/19]: offer the fixed None/Auto sentinels in the block editor

pauloiankoski:

Prepends the shared fixed sentinels to the block editor's pickable pool, so every dimension/shadow control offers `None` and Margin also offers `Auto`. The block was building Margin's `Auto` row inline; that now comes from the shared helper, which is what keeps the editor's option list identical to the Style Library's.

Option order matches the Style Library too: the sentinels lead, then the role's own scale.

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved spacing, border-width, radius, and shadow token selection.
  * Corrected the placement and availability of “None” and “Auto” options.
  * Preserved relevant bound tokens while presenting semantic options first.
  * Added support for border-width presets and improved handling of radius tokens.
* **Tests**
  * Expanded coverage to verify token ordering, filtering, presets, and special entries.


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1423, opened 2026-08-28
[DTM] SOFT-4196 [7/8]: wire all 14 color attributes to the shared ColorControl

pauloiankoski:
## Summary
- Replaces every scalar (non-composite, non-gradient) \`PopColorControl\` in \`kadence/singlebtn\`'s inspector with the new \`ColorControl\` — 14 attribute slots: \`color\`/\`colorHover\`/\`background\`/\`backgroundHover\` (main panel), the same four for the Transparent and Sticky variants, and \`iconColor\`/\`iconColorHover\` (split from a single dual-swatch \`PopColorControl\` instance into two independent \`ColorControl\`s, matching how \`color\`/\`colorHover\` are already two separate controls elsewhere in the block).
- \`groups\` is fetched once per block instance via the previous slice's \`useColorGroups(clientId)\` and shared across all 14 \`ColorControl\`s.
- \`resolveLiteral\` (new \`src/blocks/singlebtn/color-control-adapter.js\`) reads a bound token's resolved value via \`getComputedStyle(document.documentElement)\`, so seeding the Custom tab from a token pick respects the block's own pinned palette (\`data-kb-palette\`, stamped on \`<html>\` by the existing inspector-preview mechanism) rather than the site default.
- Extends the design-token preset binding indicator (previously only on the 4 main-panel attributes) to all 12 scalar text/background attributes, since they're all the same component now.
- Fixes two pre-existing bugs found while rewriting this wiring: \`backgroundTransparentHover\`'s and \`backgroundStickyHover\`'s \`onChange\` handlers were writing to \`backgroundHover\` instead of their own attribute.
- Removes the now-dead \`colorResetNonce\`/remount-key workaround — it existed only to force-remount \`PopColorControl\` on an external reset, which \`ColorControl\` (fully value-driven from props) has no need for.
- Untouched: border color and box-shadow color (their own \`renderColor\` composite wiring — see the follow-up ticket SOFT-4261) and every gradient toggle branch.

## Test plan
- \`npx jest src/blocks/singlebtn\` — all green, no regressions.
- Manually verified in the block editor: all 14 color controls render, open the grouped popover, pick/custom-set correctly, and the binding indicator shows the right bound/modified state.

## Summary by CodeRabbit

- **New Features**
  - Improved button color controls with shared, consistent color selection across primary, hover, transparent, sticky, and icon states.
  - Added support for selecting and resetting palette colors, custom colors, and color tokens.
  - Color values now resolve correctly from the active theme palette, including aliased colors.
- **Bug Fixes**
  - Improved color handling to provide more reliable styling when switching between palette and custom color options.

### DIFF
--- a/src/blocks/singlebtn/edit.js
+++ b/src/blocks/singlebtn/edit.js
@@ -54,7 +54,7 @@ import { useDispatch, useSelect } from '@wordpress/data';
 import { tooltip as tooltipIcon } from '@kadence/icons';
 import { link as linkIcon } from '@wordpress/icons';
 import { displayShortcut, isKeyboardEvent } from '@wordpress/keycodes';
-import { useEffect, useRef, useState } from '@wordpress/element';
+import { useEffect, useState } from '@wordpress/element';
 import {
 	RichText,
 	InspectorControls,
@@ -93,7 +93,6 @@ import {
 	measureAttrsForDevice,
 	presetValueForDevice,
 } from '../../extension/token-indicators/normalize';
-import { TokenControlRow } from '../../extension/token-indicators/components/TokenControlRow';
 import { EditorBoxControl } from '../../extension/design-tokens/components/EditorBoxControl';
 import { EditorBorderControl } from '../../extension/design-tokens/components/EditorBorderControl';
 import {
@@ -102,6 +101,10 @@ import {
 	splitColorOpacity,
 } from '../../extension/design-tokens/components/EditorShadowControl';
 import { pickableTokensForControl, pickableTokensForKey } from '../../extension/token-picker';
+import { ColorControl } from '../../token-controls/controls/ColorControl';
+import { ColorControlGroup } from '../../token-controls/controls/ColorControlGroup';
+import { useColorGroups } from '../../extension/design-tokens/hooks/use-color-groups';
+import { resolveColorLiteral } from './color-control-adapter';
 
 /**
  * `EditorBorderControl`'s `renderColor` render-prop: reuses the block's existing `PopColorControl`
@@ -366,6 +369,9 @@ export default function KadenceButtonEdit(props) {
 	// reset that clears the mapped attribute back to the preset value (served by the existing scoped CSS).
 	const tokenBinding = usePresetBinding('kadence/singlebtn', attributes, undefined, previewDevice);
 	const resetToken = (attr) => resetAttr(attr, setAttributes, tokenBinding[attr]?.kind);
+	// One fetch of the block's effective palette groups, shared by every `ColorControl` instance on
+	// this block — the palette data is identical for all fourteen of them.
+	const colorGroups = useColorGroups(clientId);
 
 	const borderRadiusPresetValue = presetValueForDevice(
 		tokenBinding.borderRadius?.presetValue,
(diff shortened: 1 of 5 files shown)


pauloiankoski:
@coderabbitai review

src/blocks/singlebtn/edit.js
```diff
@@ -1578,15 +1550,19 @@ export default function KadenceButtonEdit(props) {
...
-																onChange={(value) =>
-																	setAttributes({ colorTransparentHover: value })
+																groups={colorGroups}
+																onPick={(alias) =>
+																	setAttributes({ colorTransparentHover: alias })
 																}
+																onCustom={(literal) =>
+																	setAttributes({ colorTransparentHover: literal })
+																}
+																resolveLiteral={resolveColorLiteral}
```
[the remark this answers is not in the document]
pauloiankoski: Not applying this — it would be inert dead code, not real parity. `usePresetBinding()` (`src/extension/token-indicators/index.js`) only produces a `tokenBinding[attr]` entry for properties `blockProperties()` declares in the active preset's server-side catalog (`src/extension/preset-picker/index.js`), and the button preset schema (`button-preset.js`) only has Text/Background for normal/hover — no Transparent/Sticky entries at all. `tokenBinding.colorTransparent`/`colorSticky`/etc. would always be `undefined`, so wiring `status`/`onReset` against them the same way as the main-panel attributes would always render no indicator and never do anything on click — it would look like a feature that silently doesn't work, not add one. Real parity would mean registering these 8 as bindable properties server-side, which is a separate, larger change than this slice's component swap.

src/blocks/singlebtn/edit.js
```diff
@@ -1578,15 +1550,19 @@ export default function KadenceButtonEdit(props) {
 												<HoverToggleControl
 													hover={
 														<>
-															<PopColorControl
+															<ColorControl
```
d4mation, line 1557: Four of the fourteen migrated colors pass `status`/`onReset` and get `BindingIndicator`'s reset, which clears the attribute to `''` via `resetAttr` — `colorHover` (1295), `backgroundHover` (1329), `color` (1435), `background` (1467). The other ten pass neither prop, so `BindingIndicator` receives `status = null` and returns `null`, leaving no way to clear the value back to unset: `colorTransparentHover` (1553), `backgroundTransparentHover` (1591), `colorTransparent` (1696), `backgroundTransparent` (1730), `colorStickyHover` (1836), `backgroundStickyHover` (1872), `colorSticky` (1972), `backgroundSticky` (2006), `iconColor` (2318), `iconColorHover` (2326).

That is a regression against `PopColorControl`, which rendered a native clear button whenever `value !== default` and reset to `''`; these call sites passed `default={''}` with no `hideClear`. `ColorControl` offers no equivalent — `ColorGroupList` only emits real aliases and `ColorPicker`'s Custom tab always emits a real color string, so once a user sets one of these ten there is no path back to empty.

Worth noting the four that do have it are still gated behind `if (!status?.bound) return null;`, so the same gap reopens on them whenever no preset binds that property.

d4mation: Addressed in [bb6fd8c](https://github.com/stellarwp/kadence-blocks/pull/1423/commits/bb6fd8c935f6cdd87b847e7940f0d96ef34ffd05)

d4mation (approved):

PR #1424, opened 2026-08-28
[DTM] SOFT-4196 [8/8]: wire ColorSelectField into the Button preset color panel

pauloiankoski:
## Summary
- Adds \`ColorSelectField\` (\`src/style-library/components/molecules/fields/\`), the Style Library adapter for \`ColorControl\`: reads the active palette's groups via \`getPaletteListing\`/\`mapPaletteToColorControlGroups\`, and bridges the stored bare-token-id format (\`semantic.color.accent.main\`) to \`ColorControl\`'s bracket-alias contract (\`{semantic.color.accent.main}\`) at the read/write boundary — the stored format itself doesn't change, so no data migration is needed.
- Wires it into the field-type registry (\`'color-select'\`, alongside the existing \`'token-color-select'\`, which stays untouched for \`BorderField\`/\`BoxShadowField\`'s own color sub-fields) and switches the Button preset page's Text/Background fields (\`button-preset.js\`) to it.
- No binding indicator — the preset page IS the source of truth, nothing to compare a preset color against.
- \`resolveLiteral\` is a plain \`(entry) => entry.value\`, since the Style Library's own token feed already resolves every color to a literal (no CSS-var indirection needed here, unlike the block editor).
- With this slice, the last remaining out-of-the-box color picker this ticket targets is replaced — the new \`ColorControl\` now backs every scalar color attribute in both \`kadence/singlebtn\`'s inspector and the Button preset page.

## Test plan
- \`npx jest src/style-library\` — all green, no regressions.
- New tests for \`ColorSelectField\`'s value-format bridging (bare id ↔ bracket alias) and rendering.
- Manually verified in the Style Library: the Button preset page's Text/Background fields render, open the grouped popover, pick/custom-set correctly, and write back the expected bare-id/literal format.

## Summary by CodeRabbit

* **New Features**
  * Added a color selector that uses the active site palette and supports custom color values.
  * Added support for palette color fields in the Style Library settings.
  * Updated button presets to use the new color selector for text and background colors.

* **Tests**
  * Added comprehensive coverage for palette mapping, color values, custom colors, change handling, and read-only behavior.

### DIFF
--- a/src/style-library/__tests__/ColorSelectField.test.js
+++ b/src/style-library/__tests__/ColorSelectField.test.js
@@ -1,0 +1,240 @@
+/* eslint-env jest */
+/**
+ * External dependencies
+ */
+import { act, createElement } from 'react';
+import { createRoot } from 'react-dom/client';
+
+/**
+ * Internal dependencies
+ */
+import {
+	ColorSelectField,
+	resolveLiteral,
+	toControlValue,
+	toStoredValue,
+} from '../components/molecules/fields/ColorSelectField';
+import { getDesignTokensFeed } from '../helpers/tokens';
+
+const NAMESPACE = 'kb-design-tokens/v1';
+const SLUG = 'default';
+
+const LISTING = {
+	defaultId: 'default',
+	currentId: 'default',
+	userCreated: [],
+	palettes: [
+		{
+			id: 'default',
+			label: 'Default',
+			groups: [
+				{
+					id: 'accent',
+					label: 'Accent',
+					swatches: [{ token: 'semantic.color.accent.main', label: 'Main', $value: '#3182ce' }],
+				},
+			],
+		},
+		{
+			id: 'sunset',
+			label: 'Sunset',
+			groups: [],
+		},
+	],
+};
+
+// A factory, not automock: `helpers/tokens.js` reaches the localized feed, and this test only
+// cares about the namespace/slug the field hands the store selector.
+jest.mock('../helpers/tokens', () => ({
+	getDesignTokensFeed: jest.fn(),
+}));
+
+let capturedProps = null;
+
+// Captures the props `ColorControl` receives rather than rendering the real control (already
+// covered by `ColorControl.test.js`) — this test is only about the adapter's own bridging.
+jest.mock('../../token-controls', () => ({
+	ColorControl: (props) => {
+		capturedProps = props;
+		return null;
+	},
+	isTokenId: (value) =>
+		typeof value === 'string' && (value.startsWith('primitive.') || value.startsWith('semantic.')),
+	mapPaletteToColorControlGroups: (palette) =>
+		palette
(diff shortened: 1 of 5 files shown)


pauloiankoski:
@coderabbitai review

src/style-library/__tests__/ColorSelectField.test.js
```diff
@@ -0,0 +1,240 @@
...
+
+let capturedProps = null;
+
+// Captures the props `ColorControl` receives rather than rendering the real control (already
+// covered by `ColorControl.test.js`) — this test is only about the adapter's own bridging.
+jest.mock('../../token-controls', () => ({
+	ColorControl: (props) => {
+		capturedProps = props;
+		return null;
+	},
```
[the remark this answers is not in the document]
pauloiankoski: Not applying this — verified against the actual repo config, not just plugin docs. Ran `npx jest src/style-library/__tests__/ColorSelectField.test.js` directly: it passes cleanly, no hoist error. `capturedProps` is only ever an ASSIGNMENT target inside the nested `ColorControl: (props) => { capturedProps = props; ... }` closure — babel-plugin-jest-hoist's out-of-scope check flags a top-level VALUE reference to an outer binding (e.g. `onToggle: someSpy`, which I did hit and had to rename in a sibling PR earlier in this stack), not a write target inside a function body that only executes at render time, well after the module has finished initializing. Happy to rename if it ever actually breaks, but there's nothing to fix right now.

src/style-library/constants/field-types.js
```diff
@@ -99,6 +100,7 @@ export const FIELD_TYPES = Object.freeze({
 	'color-list': ColorListField,
 	'token-select': TokenSelectField,
 	'token-color-select': TokenColorSelectField,
+	'color-select': ColorSelectField,
```
d4mation, line 106: Adding `color-select` brings `FIELD_TYPES` to 18 entries, but the header docblock on line 4 still says the registry supports "seventeen strings." The same drift was flagged on #1316 ("Line 4 states that the registry supports sixteen strings. `FIELD_TYPES` now contains seventeen entries after adding `border` and `box-shadow`") and fixed by bumping the count, so line 4 wants "eighteen." It would be better to just drop a count from the comment I think.

d4mation: Addressed in [2905d70](https://github.com/stellarwp/kadence-blocks/pull/1424/commits/2905d70d4bd4d5a2f76b6c16d8855710b728e59d)

d4mation:
GitHub was reporting a merge conflict that didn't exist due to a cache on their end. Only way to clear it was to unstack, flip-flop the base, then re-stack. ¯\\\_(ツ)\_/¯

Edit: Turns out it was a real conflict that GitHub didn't make obvious. Once rebasing the stack on the feature branch the issue was clear.

image

d4mation (approved):

PR #1425, opened 2026-08-28
[DTM] SOFT-4260 [3/19]: offer the fixed sentinels in the Style Library, fixing the Auto write path

pauloiankoski:

Offers the same fixed `None` / `Auto` sentinels in the Style Library's preset fields that the previous slice added to the block editor, so both hosts show one option list.

Also fixes two write-path bugs the sentinels exposed:

- **Auto wrote a corrupted value.** Every pool entry was re-bracketed into `{id}` form on its way to the picker. For a fixed entry that turned `ss-auto` into `"{ss-auto}"`, which the write path then unwrapped to the garbage id `ss-auto` rather than the slug. Fixed entries are now exempt from the re-bracketing, since their `alias` is already the value to store.
- **Border Width's None pick wrote the same corruption**, for the same reason, through `widthTokensForField`.

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of “None” and “Auto” options in border, margin, and padding controls.
  * Fixed storage and display of sentinel values so they remain consistent and resolve correctly.
  * Ensured role-specific “None” options appear appropriately without affecting shadow controls.

* **Tests**
  * Expanded coverage for token selection, field rendering, and sentinel value conversion.

### DIFF
--- a/src/style-library/__tests__/border-field.test.js
+++ b/src/style-library/__tests__/border-field.test.js
@@ -9,6 +9,7 @@ import {
 	toStoredStyleAxis,
 	toStoredWidth,
 	toStoredWidthAxis,
+	widthTokensForField,
 } from '../components/molecules/fields/BorderField';
 
 describe('toControlWidth', () => {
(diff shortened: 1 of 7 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1427, opened 2026-08-28
[SMTNC-2046] Fix cSpell failures on the navigation block files

rahulacaleffi:
### 🎫 Ticket

[SMTNC-2046](https://linear.app/nexcess/issue/SMTNC-2046)

### 🗒️ Description

**Tweak** (code quality - no runtime change)

Fixed the failing cSpell check on `release/M26.nidoking` introduced here #968 .

**Change in `src/assets/js/kb-navigation-block.js`:**
- **Comment typos:** `currrent` → `current`, `Fullwith` → `Full-width`, `fullwith` → `full-width`
- **Header directive added:** `cSpell:ignore contenteditable keymap navs`

**Change in `src/blocks/navigation/style.scss`:**
- **Comment typos:** `Mobiel` → `Mobile`, `specificty` → `specificity`, `specifity` → `specificity` (×2)
- **Header directive added:** `cSpell:ignore fullheight modbile navs radiusp uncollapse`

`fullwith` appears only in two JSDoc blocks and nowhere as an identifier, so correcting it is text-only. Ignores are scoped to these two files rather than added to `.cspell.json`, to avoid blessing these words repo-wide.

### ⚠️ Blockers

None.

### 🧪 Testing Instructions for QA

Provided in the task description.

### Checklist

### DIFF
--- a/src/blocks/navigation/style.scss
+++ b/src/blocks/navigation/style.scss
@@ -1,6 +1,8 @@
 /**
  * Frontend only CSS.
  * Styles for both the navigation and nav link block
+ *
+ * cSpell:ignore fullheight modbile navs radiusp uncollapse
  */
 
 @use "../../assets/css/variables" as vars;
@@ -268,7 +270,7 @@ $link-color-property: var(
 			@include parent-toggles-menus-styles();
 		}
 	}
-	//Mobiel parent toggle sub menus
+	//Mobile parent toggle sub menus
 	.navigation-mobile-parent-toggles-menus-true {
 		@media (max-width: vars.$phone-minus-query) {
 			@include parent-toggles-menus-styles();
@@ -910,7 +912,7 @@ $link-color-property: var(
 		position: var(--kb-nav-link-position);
 		border-bottom: var(--kb-nav-menu-item-border-bottom);
 
-		//active nav links (extra specificty to hack correct ordering against direct styling)
+		//active nav links (extra specificity to hack correct ordering against direct styling)
 		&.current-menu-item > .kb-link-wrap.kb-link-wrap.kb-link-wrap {
 			--kb-nav-link-color-active-lv: var(--kb-nav-link-color-active);
 			--kb-nav-link-background-active-lv: var(--kb-nav-link-background-active);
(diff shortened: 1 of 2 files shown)


src/blocks/navigation/style.scss
```diff
@@ -1,6 +1,8 @@
 /**
  * Frontend only CSS.
  * Styles for both the navigation and nav link block
+ *
+ * cSpell:ignore fullheight modbile navs radiusp uncollapse
```
pauloiankoski, line 5: One thought before these land on the ignore list: `radiusp` and `modbile` look like genuine typos rather than coined words, so neither name seems to exist outside these two lines, which would make them safe to correct. Worth a quick look?

rahulacaleffi: @pauloiankoski , great observation! This is feasible for sure, but if we correct them in this file instead of ignoring as we currently are in this PR (follow-up of an [external PR#968](https://github.com/stellarwp/kadence-blocks/pull/968)), those CSS rules will suddenly start matching in live. PHP is already emitting the correct class and variable names, it's the SCSS side that's misspelled, so the rules have simply never applied.

- Mobile orientation `modbile` : [1](https://github.com/stellarwp/kadence-blocks/blob/b8037a9a62ebf8773647887cade1574711ed302d/includes/blocks/class-kadence-blocks-navigation-block.php#L421) & [2](https://github.com/stellarwp/kadence-blocks/blob/b8037a9a62ebf8773647887cade1574711ed302d/src/blocks/navigation/edit.js#L168)
- Radius variable `radiusp` : [1](https://github.com/stellarwp/kadence-blocks/blob/b8037a9a62ebf8773647887cade1574711ed302d/includes/blocks/class-kadence-blocks-navigation-link-block.php#L245) & [2](https://github.com/stellarwp/kadence-blocks/blob/b8037a9a62ebf8773647887cade1574711ed302d/includes/class-kadence-blocks-css.php#L2185)

Let me confirm if we have any reports or tickets about the mobile stretch view never being rendered correctly, then I'll certainly address it. And if we don't, I'll check with QA too. What do you think?

gilbert-hernandez (changes requested):
... Originally I didn't see ticket numbers in your PR description. But they appear there now....

gilbert-hernandez (approved):

src/blocks/navigation/style.scss
rahulacaleffi: @pauloiankoski , I'm going to open a separate issue for properly testing these classes which will start to work once we remove those real words from the ignore rule. Thanks again.

PR #1429, opened 2026-08-28
[DTM] SOFT-4260 [6/19]: move the shadow shorthand parser into a shared module

pauloiankoski:

Moves the `box-shadow` shorthand parser out of `BoxShadowControl` into `src/token-controls/helpers/shadow-shorthand.js`, so the Style Library field and the block editor adapter can both read it instead of each carrying their own copy. Pure move plus its own tests — no behavior change.

Groundwork for the fixed `None` shadow pick, which needs both hosts to agree on what a shorthand resolves to.

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of shadow values, including inset shadows, colors, spread values, and invalid or missing inputs.
  * Standardized fallback behavior for incomplete or unsupported shadow definitions.

* **Refactor**
  * Centralized shadow parsing for more consistent behavior across editor controls.

* **Tests**
  * Added coverage for common shadow formats and fallback scenarios.

### DIFF
--- a/src/extension/design-tokens/components/EditorShadowControl.js
+++ b/src/extension/design-tokens/components/EditorShadowControl.js
@@ -57,20 +57,7 @@ import { __, sprintf } from '@wordpress/i18n';
  */
 import { BoxShadowControl } from '../../../token-controls/controls/BoxShadowControl';
 import { TokenControlRow } from '../../token-indicators/components/TokenControlRow';
-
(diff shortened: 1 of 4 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1432, opened 2026-08-28
[DTM] SOFT-4260 [9/19]: migrate legacy toggled-off shadow data via a block deprecation

pauloiankoski:

Migrates data written before the previous slice removed the six "Enable Box Shadow" toggles. A toggled-off shadow becomes the explicit zero/transparent composite — the shape the fixed `None` pick writes — and the toggle attribute is dropped either way.

Two things here are load-bearing and easy to mistake for boilerplate:

- **`isEligible`, not just `migrate`.** This block is fully dynamic (`save()` returns `null`), so Gutenberg's validation trivially passes against every prior version and never reaches `deprecated` on a mismatch. `isEligible` runs regardless, which is what lets an attribute-only migration fire at all.
- **The deprecation-local `attributes` and `supports`.** `migrate`'s return value replaces the block's *entire* attribute set, so without them it would be handed `{}` and wipe out text, colors, link and sizing.

**Known gap, accepted:** a button toggled on, customized, then toggled back off kept its value with no toggle key — byte-identical to a shadow picked on a new block. That shape cannot be migrated without erasing shadows nobody meant to lose, so those buttons will show a shadow they previously kept hidden. One-time, cosmetic, on old content only.

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when editing older Single Button blocks.
  * Preserved legacy shadow settings and converted disabled or missing shadow options to “None.”
  * Removed outdated shadow toggles while retaining other block attributes.

### DIFF
--- a/src/blocks/singlebtn/__tests__/deprecated.test.js
+++ b/src/blocks/singlebtn/__tests__/deprecated.test.js
@@ -1,0 +1,151 @@
+/* eslint-env jest */
+import metadata from '../block.json';
+import deprecated from '../deprecated';
+
+const [{ attributes, supports, isEligible, migrate }] = deprecated;
+
+const NONE_ITEM = { color: 'transparent', hOffset: 0, vOffset: 0, blur: 0, spread: 0, inset: false, opacity: 1 };
+
+const LEGACY_SHADOW_DEFAULTS = {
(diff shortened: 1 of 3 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1433, opened 2026-08-28
[DTM] SOFT-4260 [10/19]: un-register the unused button padding/margin tokens

pauloiankoski:

Un-registers the `button-padding` and `button-margin` primitive tokens. Nothing bound to them: the button's padding and margin come from its size and style classes, and the previous slices made that literal the field's muted default instead.

Registered-but-unbound tokens are worse than absent ones — they surface as editable rows on the Spacing screen, so a site owner can retune a value that reaches nothing.

The SCSS keeps the same literals it always used; only the token registration goes away.

### Checklist

## Summary by CodeRabbit

* **Style**
  * Updated button spacing to use consistent default padding and zero margins.
  * Standardized padding across primary, secondary, outline, large, and small button variants.
  * Preserved outline button border adjustments across sizes.
  * Removed obsolete semantic button spacing settings.

### DIFF
--- a/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
+++ b/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
@@ -455,38 +455,6 @@
 			"heading-padding": {
 				"$type": "dimension",
 				"$value": "0"
-			},
-			"button-padding-top": {
-				"$type": "dimension",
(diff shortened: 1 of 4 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1434, opened 2026-08-28
[DTM] SOFT-4260 [11/19]: show muted Default when unset and recognize fixed sentinel picks

pauloiankoski:

Two display bugs the fixed sentinels exposed in `BoxShadowControl`:

- **An unset shadow rendered as blank.** The trigger had no visible label at all, so the control relied on an `aria-label` for its accessible name. It now shows a muted "Default", matching every other token control's unset state — and naming itself the way a bound token or a composite already does.
- **A fixed pick was not recognized after the fact.** A sentinel resolves to its literal composite at pick time and keeps no live alias, so re-opening the control showed "Custom" instead of "None". A composite value is now matched back against each fixed entry's own resolved shorthand, so the trigger and the popover's active-row highlighting both recognize it.

`token-summary.js` also stops matching a fixed entry by label when resolving a field's default, which would otherwise let a sentinel's name leak into an unrelated field's muted text.

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved box-shadow token recognition for matching composite values.
  * Fixed default-value labels so they no longer incorrectly display fixed sentinel tokens.
  * Unset shadow controls now display a muted “Default” label.
  * Nonmatching composite values continue to display as “Custom.”

* **UI Improvements**
  * Matching shadow tokens now show their label, open in the Style Library, and appear selected.

### DIFF
--- a/src/token-controls/__tests__/box-shadow-control.test.js
+++ b/src/token-controls/__tests__/box-shadow-control.test.js
@@ -17,7 +17,13 @@ import { BoxShadowControl } from '../controls/BoxShadowControl';
 // the popover's tabs are reachable without simulating a real popover open, and `TabPanel` keeps its
 // own active-tab state so a test can click between `Style Library` and `Custom`.
 jest.mock('@wordpress/components', () => ({
-	Button: ({ children, isPressed, showTooltip, ...props }) => <button {...props}>{children}</button>,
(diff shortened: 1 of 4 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1435, opened 2026-08-28
[DTM] SOFT-4260 [12/19]: correct DEFAULT_COMPOSITE's stale black fallback

pauloiankoski:

`DEFAULT_COMPOSITE` still carried a black `#000000` fallback from before the fixed `None` pick defined "no shadow" as the zero/transparent shape. The two disagreed, so a composite falling back leg-by-leg could acquire a black colour the user never picked — and, once the None sentinel matched on shorthand, a value that should have read as "None" read as "Custom" instead.

Now the fallback is the same transparent shape the sentinel writes, so the two definitions of "no shadow" are one.

### Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Shadow controls now use a transparent color when no shadow is set or when shadow values cannot be parsed.
  - Resetting a shadow consistently restores the canonical “None” state with zero dimensions and no inset.
- **Tests**
  - Updated coverage to verify transparent defaults and reliable shadow reset behavior.

### DIFF
--- a/src/extension/design-tokens/components/__tests__/EditorShadowControl.test.js
+++ b/src/extension/design-tokens/components/__tests__/EditorShadowControl.test.js
@@ -3,7 +3,13 @@
 /**
  * Internal dependencies
  */
-import { EditorShadowControl, combineColorOpacity, splitColorOpacity, toNativeShadow } from '../EditorShadowControl';
+import {
+	EditorShadowControl,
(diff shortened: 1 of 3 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1440, opened 2026-08-28
[DTM] SOFT-4260 [17/19]: show an unset editor field as a muted Default or Inherited, never as bound

pauloiankoski:

Applies the same rule to the block editor that the previous slices applied to the Style Library, so the two hosts read identically.

`EditorBoxControl` and `EditorBorderControl` were substituting the fallback **into the control's `value`**. A non-empty value reads as bound, which suppressed the muted display entirely and produced a doubled unit — a `0.4em` fallback written into an attribute whose unit lives beside it renders as `0.4empx`. Removed; the fallback now shows muted via `defaultValue`, which is what the "Inherited"/"Default" labelling already keys off.

Two consequences fall out of the same cause:

- **"Inherited" vs "Default" works again.** A smaller breakpoint falling back to the next one up labels its muted fallback "Inherited"; one falling back to the preset/baseline says "Default". That wiring already existed — the substitution was bypassing it.
- **An unset measure control opens linked.** `deriveMeasureMode()` consulted the preset value, so an all-empty control opened unlinked across four blank rows, disagreeing with the Style Library. It now derives from what the device actually stores.

`useLinkedMeasureState.toggleLink()` also stops writing the inherited value into the attribute when relinking an empty device — that turned a display fallback into a real stored override off a single click, and was the other half of the doubled-unit bug.

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved linked dimension controls for borders, spacing, and corner values.
  * Unset device values now remain unset instead of being replaced by inherited or preset values.
  * Preserved stored overrides while providing fallback values only for display.
  * Improved preset indicators and device-specific linked/individual mode detection.

* **Documentation**
  * Clarified fallback, inherited, unset, and relinking behavior in border and spacing controls.

### DIFF
--- a/src/blocks/singlebtn/edit.js
+++ b/src/blocks/singlebtn/edit.js
@@ -479,40 +479,27 @@ export default function KadenceButtonEdit(props) {
 	// slots — otherwise an explicit "link" would stick and hide a new preset's per-slot value.
 	const { isLinked: borderRadiusIsLinked, toggleLink: toggleBorderRadiusLink } = useLinkedMeasureState({
 		forDevice: borderRadiusForDevice,
-		inherited: inheritedBorderRadius,
 		previewDevice,
-		presetValue: borderRadiusPresetValue,
-		tokens: borderRadiusTokens,
 		setAttributes,
 		resetOn: attributes.kbPreset,
 	});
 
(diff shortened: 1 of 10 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1454, opened 2026-08-29
[DTM] SOFT-4201 [2/6]: ScreenDescription molecule and header slot

pauloiankoski:
## Summary

Renders the copy from the slice below, and gives `ScreenHeader` somewhere to put it. No screen
uses it yet — the wiring is the next slice up. The only place it shows after this PR is the dev
primitives gallery.

**`ScreenDescription`** (molecule) takes a screen id, reads the catalog, and renders the sentence
with `ExternalLink` at the end of it, or `null` when the screen has no entry. It takes the id
rather than the strings so the copy keeps one home and every caller passes what it already has on
its route.

**`ScreenHeader`** (organism) becomes a block: today's flex row moves into
`.screen-header-row`, and a new optional `description` slot renders under it. The block keeps the
`margin-bottom` it always had, so a screen with no copy is spaced exactly as before. The slot is
rendered bare rather than wrapped like the other slots, because a slot that renders `null` has to
leave nothing behind.

It stays pure layout: it never reads the catalog itself, the same way it knows nothing about the
palette selector or the rename link it already hosts.

## On the styling

Values are the measured spec: 13px/20px in `--kb-sl-color-text-muted`, 8px under the title, and
the header block's existing 24px to the content below.

There is deliberately **no `max-width`**. A capped measure wrapped the longest sentence to a
second line on one screen and not the others, which made the header a different height per screen.
The trade-off is a long line on a wide window; the alternative was a header that moves when you
switch screens.

`ExternalLink` already ships the theme color, the underlined label and the "opens in a new tab"
arrow with its own accessible label, so the stylesheet sets only the medium weight the header's
other text links use, plus a colour and margin reset against a core rule that paints the class
white in another context.

## Test plan

- `npm run test -- src/style-library/__tests__/screen-description.test.js` — sentence and link
  render, an uncovered screen renders nothing, an entry with no usable URL renders the sentence
  alone.
- `npm run test -- src/style-library/__tests__/screen-header.test.js` — the row wraps title and
  actions, no description means no extra node, a description renders after the row inside the
  block, and a slot that renders `null` leaves the header with one child.

## Stack

2. **This PR** — the molecule and the header slot
3. `SOFT-4201/slice-3-screen-description-wiring`
4. `SOFT-4201/slice-4-action-tooltips`
5. `SOFT-4201/slice-5-modal-enter-submit`

## Summary by CodeRabbit

* **New Features**
  * Added screen-specific helper descriptions with optional “Learn more” documentation links.
  * Enhanced screen headers to display descriptions beneath the title and actions.
  * Updated the gallery example to demonstrate contextual helper copy.

* **Style**
  * Improved header spacing, alignment, typography, link styling, and text wrapping.

* **Tests**
  * Added coverage for descriptions, documentation links, empty states, and header layouts.

### DIFF
--- a/src/style-library/__tests__/screen-description.test.js
+++ b/src/style-library/__tests__/screen-description.test.js
@@ -1,0 +1,104 @@
+/* eslint-env jest */
+/**
+ * External dependencies
+ */
+import { act } from 'react';
+import { createRoot } from 'react-dom/client';
+
+/**
+ * WordPress dependencies
+ */
+import { addFilter, removeFilter } from '@wordpress/hooks';
+
+/**
+ * Internal dependencies
+ */
+import { ScreenDescription } from '../components/molecules/ScreenDescription';
+import { SCREEN_DOCS, SCREEN_DOCS_FILTER } from '../constants/screen-docs';
+
+// `jest.config.js` maps `@wordpress/components` to a copy that resolves its own react/react-dom,
+// a different module instance than the top-level `react-dom/client` this test renders with —
+// mounting the real `ExternalLink` under the top-level renderer trips React's "Invalid hook call"
+// guard. The stand-in keeps the anchor and the href, which is all this test asserts on.
+jest.mock('@wordpress/components', () => ({
+	ExternalLink: ({ children, ...props }) => <a {...props}>{children}</a>,
+}));
+
+describe('ScreenDescription', () => {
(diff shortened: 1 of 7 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1456, opened 2026-08-29
[DTM] SOFT-4201 [3/6]: Show the helper copy on every screen

pauloiankoski:
## Summary

Turns the copy on. Every Style Library screen now shows its sentence and documentation link under
the heading — the placement SOFT-4201 asks for.

Three containers, one line each, all keyed on `route.screen`:

| Container | Screens it covers |
| --- | --- |
| `ScaleScreen` | Typography, Border Radius, Border Width, Spacing, Icon Sizes, Shadow |
| `PresetScreen` | Button, Advanced Image, Row Layout, Section, Icon, Advanced Text |
| `ColorPaletteScreen` | Color Palette |

`route.screen` rather than each container's own identity (`config.id`, the preset's block name):
it is the one key all three already have, and it covers both id shapes, so all three read the
catalog the same way.

## Screenshots

**Border Radius**

after-border-radius

**Color Palette** — the crowded lead (palette select, Rename, Reset) is untouched; the sentence
sits under the whole row

after-color-palette

**Typography** — the FONT toolbar still follows the header block

after-typography

## Test plan

- `npm run test -- src/style-library/__tests__/scale-screen-description.test.js` — the route's
  screen id picks the sentence and the href; an uncovered id renders nothing.
- `npm run test -- src/style-library/__tests__/preset-screen.test.js` — the Button screen's copy
  and link render; a preset screen with no catalog entry renders no description. The file's
  existing render helper gained a `screenId` option, defaulting to what it hardcoded before, so
  every existing test is unchanged.
- Checked in the browser on all thirteen screens: 13px/20px `#757575`, 8px under the title, 24px
  to the content, `Learn more` in the admin theme colour with core's "opens in a new tab" arrow.

`ColorPaletteScreen` has no render test of its own. Mounting it means mounting the palette
dropdown, the activate button, the swatch grid and six modals, all of which need stubbing — the
test would assert on the stubs, not on a one-line prop. The behaviour is covered by the two files
above plus the molecule's own tests, and this screen was checked in the browser.

## Stack

1. `SOFT-4201/slice-1-screen-docs-catalog` — #1453
2. `SOFT-4201/slice-2-screen-description-slot` — #1454
3. **This PR** — the wiring
4. `SOFT-4201/slice-4-action-tooltips`
5. `SOFT-4201/slice-5-modal-enter-submit`
6. `SOFT-4201/slice-6-screen-header-height`

## Summary by CodeRabbit

* **New Features**
  * Added contextual descriptions to style library screens.
[...]

PR #1457, opened 2026-08-29
[DTM] SOFT-4201 [4/6]: Explain the set-active and favorite actions on hover

pauloiankoski:
## Summary

Three actions now say what they will do before you click them — SOFT-4201's scope item 3, plus two
neighbours with the same problem:

| Action | Sentence |
| --- | --- |
| Set as active (palette) | Makes this palette the one your site uses. Individual blocks can still be switched to another palette. |
| Set as active (library) | Makes this library the one your site uses. Your site switches to its colors, sizes and styles. |
| Add Favorite | Pins this font to the top of your font pickers, here and in the editor. It does not change any text on your site. |
| Remove Favorite | Unpins this font from the top of your font pickers. Text already using it stays as it is. |

The ticket names only the palette action. The library action beside it and the Typography favorite
action have the same "what will this actually do to my site" question, and the favorite one has a
second worry worth answering directly: it does not restyle anything.

Each is one `<Tooltip>` wrap around a button that is otherwise untouched.

## Two decisions worth review

**The documentation link is dropped from the tooltip.** The ticket's copy ends with `[Learn more]`,
but core's `Tooltip` takes a plain string and is neither hoverable nor focusable — it hides on blur
and on mouse-leave, so a link inside it cannot be reached by mouse or keyboard. Shipping one would
be an affordance nobody can use. The screen's documentation link is already on the page, in the
description row from the slices below. Putting the link in the tooltip means an info icon opening a
`Popover` — its own trigger, open state, focus management and dismiss handling — which is a feature,
not a tooltip, and belongs in its own ticket if it is wanted.

**`Tooltip`, not `Button`'s own `label` + `showTooltip`.** That pair sets the button's `aria-label`,
which would replace the accessible name "Set as active" with the whole sentence. `Tooltip` wires
`aria-describedby` instead, so the name stays and the sentence is read as its description. Verified
in the browser: the button keeps its name and picks up `aria-describedby` pointing at the tooltip
text while it is shown.

The two Set Active buttons use plain `disabled`, so their tooltips are suppressed while a write is
in flight — correct, there is nothing to explain about a control that cannot be used. The favorite
button is `accessibleWhenDisabled`, so it stays focusable and its tooltip keeps working.

## Screenshot

tooltip-add-favorite

## Test plan

- `npm run test -- src/style-library/__tests__/activate-button-tooltips.test.js` — each Set Active
  action carries its sentence and keeps "Set as active" as its visible label; neither renders at
  all while the thing being edited is already active.
- `npm run test -- src/style-library/__tests__/typography-screen.test.js` — the add and remove
  states each carry their own sentence; the file's existing six tests are unchanged.

## Note on `.cspell.json`

Two words added: `describedby` and `hoverable`. Both are in comments explaining the above.

## Stack

4. **This PR** — the tooltips
5. `SOFT-4201/slice-5-modal-enter-submit`

## Summary by CodeRabbit

* **New Features**
  * Added hover tooltips explaining how to activate libraries and palettes.
  * Added contextual tooltips for adding or removing favorite fonts.
  * Tooltips include localized, descriptive guidance while preserving existing controls and behavior.

* **Tests**
  * Added coverage confirming tooltip text, button labels, and active-state behavior.


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1458, opened 2026-08-29
[DTM] SOFT-4201 [5/6]: Submit the add and rename modals with Enter

pauloiankoski:
## Summary

Typing a name in any add or rename modal and pressing Enter did nothing — you had to leave the
keyboard and click. All six modals rendered `TextControl` and the primary `Button` as siblings with
no `<form>` between them, so the browser had nothing to submit.

Each modal's body is now wrapped in `<form onSubmit>` with `type="submit"` on the primary button,
which turns on the browser's own implicit submission. No key handler of our own.

Covers: Create Library, Rename Library, Create Color Palette, Rename Color Palette, Add Color
Group, Rename Color Group.

## Three details that decide whether this is correct

- **`Button` accepts a `type`.** Core hardcodes `type: "button"` but spreads the caller's props
  after it, so `type="submit"` wins. Checked in the built component, not assumed.
- **Cancel is now explicitly `type="button"`.** It already rendered as one by core's default, but
  inside a form the difference between Cancel and submit is the difference between dismissing and
  saving, so it is stated in the markup rather than left to a library default.
- **Each submit handler repeats its primary button's disabled expression**, named once per modal
  (`canCreate` / `canSave` / `canAdd`) and used by both. A disabled default button already blocks
  implicit submission per the HTML spec, so this is belt and braces — it is there so the guard
  survives if that button's disabled state is ever loosened.

## Reviewing the diff

Most of the line count is re-indentation from the `<form>` wrap. `git diff -w` shows the real
change: **117 insertions, 23 deletions** across the six files.

## Screenshot

modal-add-color-group

## Test plan

- `npm run test -- src/style-library/__tests__/modal-submit.test.js` — submitting creates the
  library and renames the group with the trimmed value; an empty title and an unchanged label both
  submit to nothing; Cancel is a `type="button"` that dismisses without creating.
- Checked in the browser: the Add Color Group modal renders a `<form>`, Add is `type="submit"` and
  disabled until the field has a value, Cancel is `type="button"`.

## Note on scope

This is not in SOFT-4201 as written — it is a usability fix folded in at the ticket owner's
request. It touches nothing the other slices touch, so it lifts out cleanly if you would rather
review it on its own.

## Stack

4. `SOFT-4201/slice-4-action-tooltips` — #1457
5. **This PR** — Enter submits the modals

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved modal form submission across library, palette, and color-group creation and renaming workflows.
  * Prevented invalid submissions when fields are empty, unchanged, duplicated, or while an operation is in progress.
  * Ensured Cancel actions do not accidentally submit forms.
  * Preserved inline validation errors and loading states.

* **Tests**
  * Added coverage for submit behavior, validation rules, trimming, cancellation, and successful creation or renaming.

### DIFF
--- a/src/style-library/__tests__/modal-submit.test.js
+++ b/src/style-library/__tests__/modal-submit.test.js
@@ -1,0 +1,205 @@
+/* eslint-env jest */
+/**
+ * External dependencies
+ */
+import { act } from 'react';
+import { createRoot } from 'react-dom/client';
+
+/**
+ * Internal dependencies
+ */
+import { CreateLibraryModal } from '../components/organisms/CreateLibraryModal';
+import { RenameColorGroupModal } from '../components/organisms/RenameColorGroupModal';
+
+// Stand-ins for the nested `@wordpress/components` copy (see any other screen test for why).
+// `Modal` renders its children inline and `TextControl` renders a real input, so the form and its
+// implicit submission behave the way they do in the browser.
+jest.mock('@wordpress/components', () => ({
+	Modal: ({ children }) => <div>{children}</div>,
+	Notice: ({ children }) => <div>{children}</div>,
+	Button: ({ children, ...props }) => <button {...props}>{children}</button>,
+	TextControl: ({ label, value, onChange, help, ...props }) => (
+		<input aria-label={label} value={value} onChange={(event) => onChange(event.target.value)} {...props} />
+	),
+}));
+
+let container;
+let root;
+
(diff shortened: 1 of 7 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):
THANK YOU FOR THIS

I have been meaning to say something every other time I use this interface

PR #1459, opened 2026-08-29
[DTM] SOFT-4201 [6/6]: Give every screen header the same height

pauloiankoski:
## Summary

The screen header was a different height on every kind of screen, so the content below moved every
time you switched tabs. Two separate causes, both fixed here.

## 1. The row had no floor

Its height was whatever the tallest thing inside it happened to be:

| Screen | Tallest thing in the row | Height |
| --- | --- | --- |
| Typography | the title alone — its Add Size lives in the toolbar below | 28px |
| Border Radius, Border Width, Spacing, Icon Sizes, Shadow, all six preset screens | the "+ Add …" button | 36px |
| Color Palette | the palette `SelectDropdown` | 40px |

`min-height: var(--kb-sl-space-50)` on `.screen-header-row` — the same primitive the
`SelectDropdown` uses for its own fixed 40px height, so the two stay visibly in agreement. A floor
rather than a fixed height, so Color Palette's five-item lead can still grow instead of
overflowing. Nothing shrinks: the 28px and 36px rows grow to meet the 40px one.

40 rather than 36, because the dropdown's height is deliberate and shared with the app header's
library selector and several settings fields. A 36px floor would not standardise Color Palette
anyway — `min-height` shrinks nothing, so that screen would stay 40 and still jump.

## 2. Color Palette paid for its gap twice

`.color-palette-screen` is the only screen root that is a flex column with its own
`gap: var(--kb-sl-space-30)`. Every other screen is a plain block where the shared header's
`margin-bottom` is the single separator. On that one screen the two stacked: **48px against
everyone else's 24px**. The margin now steps aside there and the column's gap is the one owner.

The alternative was removing `margin-bottom` from `ScreenHeader` entirely and making all thirteen
screen roots flex columns. That is arguably the cleaner ownership model, but it would set a 24px
gap between every pair of children on every screen, changing the spacing around inline `Notice`s
that use their own margins today. That is a spacing redesign, not a bug fix, so it is deliberately
not in here.

## Result

Measured across all thirteen screens after the change: row **40px**, header **68px**, gap to
content **24px**, content top **209px** — identical everywhere.

**Before** — 28px, 36px and 40px rows

before-border-radius-36px
before-color-palette-40px
before-typography-28px

**After** — the same three screens

after-border-radius
after-color-palette
after-typography

## Known limit

The description has no `max-width`, so the longest sentence still wraps once the content column
drops below ~764px — roughly a laptop under 1376px wide with the settings panel open. At that width
the Button preset screen goes back to a taller header. Nothing short of shortening that sentence
removes it entirely.

## Test plan

CSS only; no test changes. The suite stays green (1701 tests). Verified in the browser by walking
all thirteen screens and measuring the row, the header block, and the top of the first content
element on each.

## Stack

4. `SOFT-4201/slice-4-action-tooltips` — #1457
5. `SOFT-4201/slice-5-modal-enter-submit` — #1458
6. **This PR** — one header height everywhere

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved screen header sizing to maintain a consistent minimum height while accommodating longer content.
  * Corrected spacing on the color palette screen to prevent excessive gaps below the header.

### DIFF
--- a/src/style-library/components/organisms/ScreenHeader.scss
+++ b/src/style-library/components/organisms/ScreenHeader.scss
@@ -4,11 +4,15 @@
 	margin-bottom: var(--kb-sl-space-30);
 }
 
+// The floor matches the tallest control the row can hold — the palette `SelectDropdown`, which
+// sets the same `var(--kb-sl-space-50)` height — so the row does not change height from screen to
+// screen. A floor rather than a fixed height: a crowded lead still grows instead of overflowing.
 .kadence-blocks-style-library__screen-header-row {
 	display: flex;
 	align-items: center;
 	justify-content: space-between;
 	gap: var(--kb-sl-space-15);
+	min-height: var(--kb-sl-space-50);
 }
 
 .kadence-blocks-style-library__screen-header-lead {
--- a/src/style-library/components/pages/ColorPaletteScreen.scss
+++ b/src/style-library/components/pages/ColorPaletteScreen.scss
@@ -4,6 +4,12 @@
 	@include flex-column(var(--kb-sl-space-30));
 }
 
+// The only screen root that is a flex column with its own gap, which would otherwise stack with
+// the shared header's bottom margin: 48px here against 24px everywhere else.
+.kadence-blocks-style-library__color-palette-screen .kadence-blocks-style-library__screen-header {
+	margin-bottom: 0;
+}
+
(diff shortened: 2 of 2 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1462, opened 2026-08-30
Token-pickable block controls [8/11]: Advanced Text size, border, radius, padding and colors

d4mation:
🎫  https://linear.app/nexcess/issue/SOFT-4263/token-pickable-block-controls-for-the-remaining-alpha-blocks-row
:video_camera: https://www.loom.com/share/9363726b1a30466a9d6e8fb429f82f2d

The Advanced Text preset screen offers all of these and each binds a token, but the block's controls could only take a literal.

Font Size is the shape this stack had not met: the heading packs all three devices into ONE array attribute rather than naming a sibling per device, so the breakpoint-to-attribute mapping cannot address it — the index is resolved in the caller instead. Border width uses the by-key pool lookup, since width, style and color share one control attribute. Text and Background become `ColorControl`, each clearing its legacy `*Class` attribute so a stale palette class cannot beat the pick. Font Family passes its binding state through the seam.

Also lifts `renderBorderColor` out of the Button to sit beside `renderShadowColor`. The block toolbar's color control is deliberately left alone — it routes through the same swatch filter, so a palette pick there already writes the same alias.

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added design-token support to Advanced Heading styling controls, including colors, borders, typography, spacing, and corner radius.
  * Added visibility for token-bound and overridden settings, with reset actions.
  * Preserved responsive value inheritance across supported styling options.

* **Bug Fixes**
  * Improved unit-aware rendering for border-radius values.
  * Standardized border-color handling in button styling.

### DIFF
--- a/src/blocks/advancedheading/edit.js
+++ b/src/blocks/advancedheading/edit.js
@@ -78,7 +78,27 @@ import Typed from 'typed.js';
  * Import Css
  */
 import './editor.scss';
-import { presetPropertyValueForDevice } from '../../extension/token-indicators';
+import {
+	presetPropertyValueForDevice,
+	resetAttr,
+	useLinkedMeasureState,
+	usePresetBinding,
+} from '../../extension/token-indicators';
+import {
+	anyCornerInherited,
+	inheritedMeasureSlots,
+	measureAttrsForDevice,
+	presetValueForDevice,
+} from '../../extension/token-indicators/normalize';
+import { EditorBorderControl } from '../../extension/design-tokens/components/EditorBorderControl';
+import { EditorBoxControl } from '../../extension/design-tokens/components/EditorBoxControl';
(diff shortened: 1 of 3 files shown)


pauloiankoski (approved):
After moving `renderBorderColor` out of `src/blocks/singlebtn/edit.js`, `upperFirst` and `sprintf` are not used there anymore. Can you remove both imports? The linter does not catch this one, so it is easy to miss.

d4mation:
@pauloiankoski Fixed that in [e1955ba](https://github.com/stellarwp/kadence-blocks/pull/1462/commits/e1955ba757f86279777053c567845e7dceb4b5bd) and fixed a similar issue in #1451 with [92a69e2](https://github.com/stellarwp/kadence-blocks/pull/1451/commits/92a69e296d2ee77c02228e784292239ebdaf8b8e)

PR #1467, opened 2026-08-30
[DTM] SOFT-3908 [1/4]: palette inheritance helpers

pauloiankoski:
Adds the two pure rules the Color Palette screen's inheritance affordances read. No UI yet — the pills and the notice that consume these land in the slices stacked on top of this one.

## What this adds

`src/style-library/helpers/palettes.js`, the screen's home for pure data rules (no REST, no JSX):

- `paletteShowsInheritance( listing, editingId )` — whether the palette being edited shows inheritance affordances at all. Only a non-default palette does: the default palette is where a swatch's value is defined, so it has no source to name and no delta of its own to drop. It composes the existing `isDefaultPalette()`, so it inherits that helper's fail-closed behavior — a listing with no `$default` pointer reports no inheritance rather than naming a source it cannot verify.
- `inheritedSwatchCount( groups )` — how many swatches in a mapped grid still take their value from the default palette. Groups and swatches that are optimistically deleted are skipped: they are on their way out of the grid, and counting them would state a number the grid is about to contradict.

`groups` is the shape `mapPaletteToSwatchGroups()` already produces, so the count describes exactly the cards on screen — optimistic additions included, pending deletions excluded.

## Test plan

`npm test -- src/style-library/__tests__/palettes.test.js` — 123 passing.

Eight new cases: non-default palette, the default palette itself, a listing with no default pointer, no palette open; multi-group counting, pending-delete groups and swatches skipped, and an empty list.

## Summary by CodeRabbit

* **New Features**
  * Added inheritance indicators for palettes when applicable.
  * Added counts for inherited swatches that remain active and have not been overridden.

* **Bug Fixes**
  * Improved handling of default, missing-default, empty, and pending-deletion palette selections when displaying inheritance information.

## Screenshots

_Nothing to show: this slice is two pure functions with no UI surface. The screens that render them are in the slices stacked on top._

### DIFF
--- a/src/style-library/__tests__/palettes.test.js
+++ b/src/style-library/__tests__/palettes.test.js
@@ -5,6 +5,7 @@ import {
 	applyOptimisticOverlay,
 	customColorTokenId,
 	findSwatch,
+	inheritedSwatchCount,
 	isCustomColorToken,
 	isDefaultPalette,
 	isDuplicatePaletteLabel,
@@ -13,6 +14,7 @@ import {
 	newSwatchValue,
 	nextCustomColorSlug,
 	paletteDisplayLabel,
+	paletteShowsInheritance,
 	paletteSuccessorOptions,
 	removeGroupFromGroups,
 	removeSwatchFromGroups,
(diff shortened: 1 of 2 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1468, opened 2026-08-30
[DTM] SOFT-3908 [2/4]: swatch card pill slot

pauloiankoski:
Gives `SwatchCard` a slot under its value line for a caller-supplied node, and makes room for it without changing how any card looks today. Nothing fills the slot yet — the inheritance pill that does lands in the slice stacked on this one.

Stacked on #1467.

## The card's click target moved

The bordered box was itself a `<button>`. The slot has to be able to hold a button of its own, and a button inside a button is invalid markup that swallows the inner one's clicks. So the box is now a plain `<div>` and a new inner `.swatch-card-select` button covers exactly what it covered before: the preview, the name, and the sub-line.

The box model did not move with it. Border, radius, background, width and padding all stay on `.swatch-card-main`, so every existing hover, selected, drag-placeholder and pending-delete rule keeps describing the same box it always did. The inner button carries only the layout its own children need, plus the pointer cursor.

## The slot is reserved per row, not per card

A row where at least one card has a node reserves the slot on **every** card in that row, so a mixed row stays on one baseline. A row where no card has one renders no slot at all, so a grid with nothing to show — the default color palette, and the component gallery in `PlaceholderScreen` — keeps its cards exactly the height they are today, with no empty strip at the bottom.

`SwatchGrid` owns that decision because the row is where it belongs: the group derives whether any of its items carries a node and passes `reservePillSlot` down to each card. The `DragOverlay` copy gets the reservation too but never the node itself — a second live control following the pointer would be a stray tab stop, while the reserved height keeps the floating copy matching the placeholder it stands in for.

## Loading skeleton

`SwatchGridSkeleton` gains the `.swatch-card-select` wrapper so its bars keep hugging the left edge now that the alignment lives on that element. It deliberately does not reserve the slot: the placeholder cannot know whether the palette about to load has anything to show, and the palette most sites use has nothing.

## Test plan

`npm test -- src/style-library/__tests__/swatch-grid.test.js`

Covers: the node renders inside the bordered box but outside the selecting button (the nested-button guarantee); a group with no nodes renders no slot anywhere; a group with at least one renders the slot on every card including the ones without; selecting still works through the inner button and is disabled while a swatch is pending delete.

The whole app suite is green, and `npx eslint src/style-library` is clean.

## Screenshots
default-palette-accent-row
default-palette-no-pills

## Summary by CodeRabbit

- **Improvements**
  - Swatch cards now support optional action pills, such as Reset, without disrupting card selection.
  - Swatch groups reserve consistent space for pills so cards remain aligned, including during sorting and dragging.
  - Pending-deletion cards continue to display the correct disabled state.
  - Loading placeholders now more closely match the final swatch card layout.

### DIFF
--- a/src/style-library/components/molecules/SwatchCard.js
+++ b/src/style-library/components/molecules/SwatchCard.js
@@ -40,6 +40,10 @@ import './SwatchCard.scss';
  *                                              this is never derived from `subLine`.
  * @param {string}       props.name             The card's name.
  * @param {string}       [props.subLine]        The sub-line under the name (e.g. a hex value).
+ * @param {?JSX.Element} [props.pill]    The node for the slot under the sub-line — the caller's
+ *                                       inheritance pill, or null. Caller-supplied for the same
+ *                                       reason `preview` is: this card is used by screens that
+ *                                       have no notion of palette inheritance at all.
  * @param {boolean}      [props.isSelected]     Whether the card shows the selected treatment.
  * @param {Function}     props.onSelect         Called with the card id on click.
  * @param {boolean}      [props.isDraggable]    Whether the drag handle renders.
@@ -68,6 +72,7 @@ export function SwatchCard({
 	previewStyle,
 	name,
 	subLine,
+	pill = null,
 	isSelected = false,
 	onSelect,
 	isDraggable = false,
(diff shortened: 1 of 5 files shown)


pauloiankoski:
@coderabbitai review

src/style-library/components/pages/ColorPaletteScreen.js
```diff
@@ -81,15 +81,18 @@ function SwatchGridSkeleton({ label }) {
...
-								<Skeleton className="kadence-blocks-style-library__swatch-card-preview" />
-								{/* `.swatch-card-name` only declares `max-width: 100%`, never a `width` — a real
-								 * swatch name gets its width from its own text, but this shape has none, and its
-								 * `align-items: flex-start` parent collapses an unsized block to 0 width without
-								 * one. Same fix as the group heading bar above: pin a plausible literal width. */}
-								<Skeleton
-									className="kadence-blocks-style-library__swatch-card-name kadence-blocks-style-library__skeleton--bar"
-									style={{ width: '70%' }}
-								/>
+								<div className="kadence-blocks-style-library__swatch-card-select">
```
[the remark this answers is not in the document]
pauloiankoski: Fixed in d1dca3758. `cursor: pointer` moved off the shared `.swatch-card-select` class onto a `button.`-qualified rule, so the pointer belongs to the real control and the skeleton's inert `div` inherits only the layout.

src/style-library/components/molecules/SwatchCard.js
```diff
@@ -87,19 +97,30 @@ export function SwatchCard({
...
-					{preview}
-				</span>
-				<span className="kadence-blocks-style-library__swatch-card-name">{name}</span>
-				{subLine && <span className="kadence-blocks-style-library__swatch-card-sub-line">{subLine}</span>}
-			</button>
+			{/* The bordered box is a plain element, not the click target it used to be: the pill slot
+			 * below can hold a button, and a button inside a button is invalid markup. Selecting the
+			 * card moved to the inner button, which covers exactly what it covered before. */}
+			<div className="kadence-blocks-style-library__swatch-card-main">
+				<button
```
d4mation, line 111: This ended up slightly breaking styling. If you try to click the edges of the container, that is now the `<div>` itself (due to the padding) rather than the `<button>` so clicking along the edges no longer does anything. Moving the padding from the `<div>` onto the `<button>` like it used to be should solve this.

pauloiankoski: Fixed across 0fa027a77 and 9a7ecd35a. The padding moved off the `div` as you suggested, split between the button (`8px 8px 0`) and the pill slot (`0 8px 8px`) so the gap between the hex and the pill stays what the design has — a single `padding: 8px` on the button would have stacked with the slot's own spacing.

That fixed the top and side edges but not the pill row, which is also an edge, so the button's hit area is now stretched over the whole card with an `::after`. Only an interactive pill is lifted above it, so a Reset button still takes its own click while the static `From Default` pill and the space around it select the card.

Verified in the browser on all four: padding ring selects, the band under the pill selects, the static pill selects, and Reset still resets. No unit test for this — jsdom does not apply the stylesheet, so a hit-area assertion there would prove nothing.

d4mation: When the pill doesn't exist, this results in no padding on the bottom (Easiest to see when viewing the Default palette).

We may need a BEM modifier for when the pill is present so that we can have it default to having that bottom padding and then remove it when a pill is there.

image

pauloiankoski: Fixed in b04fddab7, with the modifier you suggested. `.swatch-card-select` is back to a full `padding: var(--kb-sl-space-10)`, and a new `--with-pill-slot` modifier on the card drops its `padding-bottom` only when the slot actually renders and takes the lower inset over. So the default palette gets its bottom inset back, and a card with a pill keeps the value line and the pill the distance apart the design has them.

The modifier is set from one `hasPillSlot` value in `SwatchCard`, the same one that decides whether to render the slot, so the class and the element can't disagree. Covered by a new test asserting the flag appears only on cards whose slot renders.

Verified in the browser on both palettes, and the whole-card click target still works on the default palette where the slot is absent.

d4mation (approved):

PR #1470, opened 2026-08-30
[DTM] SOFT-3908 [4/4]: palette inheritance notice

pauloiankoski:
States the inheritance rule once, above the grid, instead of leaving it to be inferred from the pills.

Stacked on #1469.

## What it says

> 17 colors in this palette still follow Default. Until you customize them in this palette, editing them in Default will also update them here.

Rendered only while the palette being edited has a source to inherit from, and only while something still inherits — customize the last one and the notice goes away rather than announcing zero. Singular and plural are separate strings, so one color reads "1 color … still follows … until you customize it".

Like the pills, the palette is named from the listing rather than the literal word "Default", so renaming the default palette renames it here too.

The count comes from the mapped grid groups — the cards actually on screen — so an optimistically added color counts immediately and one being deleted stops counting before the write settles, instead of the number contradicting the grid under it.

`role="status"` on the notice means the count change after a reset is announced, which is the only feedback a screen-reader user gets that the reset landed once the pill they activated has been replaced.

## Test plan

`npm test -- src/style-library/__tests__/palette-inheritance.test.js`

Covers the exact sentence for the current count, absence when every color is customized, and absence on the default palette. The reset round-trip test in the previous slice asserts the count grows when a color goes back to inherited.

The whole app suite is green, and `npx eslint src/style-library` is clean.

## Screenshots

shows the notice above the grid with the pills underneath it.
secondary-palette-pills-and-notice

## Summary by CodeRabbit

* **New Features**
  * Added an accessible notice to color palette screens when colors are inherited from another palette.
  * Notices now show the number of inherited colors and explain how updates to the source palette affect them.
  * Added singular and plural messaging, with no notice shown for fully customized or default palettes.

* **Bug Fixes**
  * Corrected inheritance counts after palette colors are reset to inherited values.

### DIFF
--- a/src/style-library/components/molecules/PaletteInheritanceNotice.js
+++ b/src/style-library/components/molecules/PaletteInheritanceNotice.js
@@ -38,8 +38,8 @@ export function PaletteInheritanceNotice({ count, sourceLabel }) {
 			{sprintf(
 				/* translators: 1: number of colors, 2: the source palette's name. */
 				_n(
-					'%1$d color in this palette still follows %2$s. Editing it in %2$s updates it here too, until you customize it.',
-					'%1$d colors in this palette still follow %2$s. Editing them in %2$s updates them here too, until you customize them.',
+					'%1$d color in this palette still follows %2$s. Until you customize it in this palette, editing it in %2$s will also update it here.',
+					'%1$d colors in this palette still follow %2$s. Until you customize them in this palette, editing them in %2$s will also update them here.',
 					count,
 					'kadence-blocks'
 				),
--- a/src/style-library/components/molecules/PaletteInheritanceNotice.js
+++ b/src/style-library/components/molecules/PaletteInheritanceNotice.js
@@ -1,0 +1,51 @@
+/**
+ * The notice above the swatch grid, stating how many of this palette's colors still take their
+ * value from the palette they were built on, and what that means when that palette is edited.
+ *
+ * Stated once, in place, rather than behind a tooltip on each card: inheritance is the state of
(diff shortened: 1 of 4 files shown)


pauloiankoski:
@coderabbitai review

src/style-library/components/molecules/PaletteInheritanceNotice.js
```diff
@@ -0,0 +1,51 @@
...
+		return null;
+	}
+
+	return (
+		<p className="kadence-blocks-style-library__palette-inheritance-notice" role="status">
+			{sprintf(
+				/* translators: 1: number of colors, 2: the source palette's name. */
+				_n(
+					'%1$d color in this palette still follows %2$s. Editing it in %2$s updates it here too, until you customize it.',
+					'%1$d colors in this palette still follow %2$s. Editing them in %2$s updates them here too, until you customize them.',
```
d4mation: Minor adjustment. This will read a little nicer.

```suggestion
					'%1$d color in this palette still follows %2$s. Until you customize it in this palette, editing it in %2$s will also update it here.',
					'%1$d colors in this palette still follow %2$s. Until you customize them in this palette, editing them in %2$s will also update them here.',
```

pauloiankoski: Applied verbatim in 106cf86d3 (now 06d8171f3 after the rebase), both the singular and plural strings, plus the two screen tests that pin the exact sentence.

d4mation (approved):

PR #1476, opened 2026-08-30
[DTM] SOFT-4266 [1/6]: bind a picked shadow token instead of freezing it to a literal

pauloiankoski:
Picking a shadow token in the block editor wrote the token's resolved value into the block's plain numeric fields (`hOffset`, `vOffset`, `blur`, `spread`, `color`, `opacity`). The control then read that back as "Custom", and the value stopped tracking the token — editing the token later did not move the button.

This slice adds the data model and the control behavior. Nothing renders differently yet; the two renderers follow in later slices of this stack.

## What changed

**A new `shadowToken` key on the shadow item.** `src/extension/design-tokens/shadow-token.js` holds the key and `boundShadowToken()`, the one function that reads it. It is a module of its own, depending only on `./alias`, because the editor-canvas style builder needs the key too and must not pull a React component, its SCSS, and the `token-controls` barrel into a module that only builds strings.

**The write path.** `toNativeShadow()` records the picked `{dot.alias}` on `shadowToken` **and** still splays the token's resolved value across the numeric legs. Keeping both is deliberate:

- the item stays a valid literal shadow for every reader that does not know the new key, and
- they seed the control's Custom tab, so switching a bound shadow to Custom starts from real values.

The two are not a second opinion: when the binding is present and backed by the active library it wins outright and the legs are never read, and when the token has been deleted the renderers fall back to the block's default CSS rather than to the legs. The legs are a frozen snapshot the inspector reads, not a rendering fallback.

Only a real `{dot.alias}` binds. The Style Library's fixed "None" sentinel arrives as the literal string `0px 0px 0px 0px transparent`, which is not alias-shaped, so it keeps resolving to a plain zero item with no binding. Any Custom-tab edit reports a composite object, which carries no alias, so touching a leg breaks the binding — the value is then the user's, not the token's.

**The read path.** `EditorShadowControl` hands `BoxShadowControl` the bare alias string when a binding is present, which is the shape it already recognizes: the trigger names the token, the popover opens on the Style Library tab, and the preview square shows the token's own shadow. `isUnsetShadow()` treats a bound item as set even when every leg is zero, since some registered shadow tokens genuinely resolve to an all-zero transparent shadow and reading those as "unset" would drop the token's name from the trigger.

**Two fixes in `BoxShadowControl`.** With an alias value, the Custom tab previously seeded from the all-zero `DEFAULT_COMPOSITE`, so switching to it on a bound token showed zeros and editing one leg discarded the token's real value. It now seeds from the bound token's own resolved shorthand, resolving from `tokens` because the token's current value is what is actually rendering. When the alias matches no entry — a token deleted after saving — an optional `fallbackShadow` prop seeds it from the host's stored legs instead, and the preview square above the tab falls through to the same composite so the two agree. This also fixes the Style Library host, which uses the same control and passes no `fallbackShadow`.

The three hand-rolled `entry.alias === value` lookups this area had are now single calls to the existing `findTokenEntry()`.

## No deprecation needed

`kadence/singlebtn` is fully dynamic — `save()` returns `null`, so Gutenberg's markup validation trivially passes and never reaches `deprecated`. `shadowToken` is a new optional key: absent on all existing content, and absent means "no token bound", which is exactly today's behavior. There is no schema change either — `shadow` is declared `{ "type": "array" }` with a default and no `items` schema.

## Test plan

- `npm test -- src/extension/design-tokens src/token-controls`
- The write path is covered for: a pick recording the alias, the resolved legs surviving alongside it, a Custom-tab edit dropping the binding, the fixed "None" sentinel *not* binding, and an alias matching no pickable token still recording the binding while falling the legs back to the default shape.
- The read path is covered for: the bound alias being handed down instead of the composite, a bound item with all-zero legs reading as set, and an unbound item still bridging to the composite.
- `BoxShadowControl`'s existing alias-label tests prove the half of the bridge this slice does not write, and are unchanged.

## Summary by CodeRabbit

- **New Features**
  - Shadow selections retain their design-token bindings while preserving resolved shadow values.
  - Edit and preview shadows when token aliases are stale or unavailable, using saved fallback values.
  - Live token values take priority when available.

- **Bug Fixes**
  - Improved handling of zero-value shadows, unknown aliases, missing tokens, and the fixed “None” option.
  - Composite shadow edits now remove outdated token bindings and preserve accurate visual previews.

### DIFF
--- a/src/extension/design-tokens/components/EditorShadowControl.js
+++ b/src/extension/design-tokens/components/EditorShadowControl.js
@@ -46,7 +46,7 @@
 /**
  * Internal dependencies
  */
-import { BoxShadowControl, DEFAULT_COMPOSITE, parseResolvedShadow } from '../../../token-controls';
+import { BoxShadowControl, DEFAULT_COMPOSITE, findTokenEntry, parseResolvedShadow } from '../../../token-controls';
 import { TokenControlRow } from '../../token-indicators/components/TokenControlRow';
 import { isTokenAlias } from '../alias';
 import { SHADOW_TOKEN_KEY, boundShadowToken } from '../shadow-token';
@@ -223,6 +223,13 @@ function axisToNative(slot) {
  * Resolve a picked token alias to its literal composite value, via the same `tokens` list
  * `BoxShadowControl` already offers for its trigger label.
  *
+ * Uses `findTokenEntry()` rather than a plain `entry.alias === alias` scan. The only string this
+ * function ever receives that is not a real `{dot.alias}` is `toNativeShadow()`'s fixed "None"
+ * sentinel, `0px 0px 0px 0px transparent` — that matches a `fixed: true` entry on alias equality,
+ * which `findTokenEntry()`'s guard (`entry.fixed || isTokenAlias(value)`) also accepts. Its
+ * legacy-slug fallback is inert here: it only matches a `primitive.dimension.*` entry, and a
+ * shadow-role pool holds none.
+ *
  * @param {string} alias  The picked token alias.
  * @param {Array}  tokens Pickable `shadow`-type tokens, `[{id, label, value, alias}]`.
  *
@@ -231,7 +238,7 @@ function axisToNative(slot) {
  * @return {?Object} The resolved composite, or `null` when the alias matches no pickable token.
  */
 function resolveShadowAlias(alias, tokens) {
-	const entry = (tokens || []).find((token) => token.alias === alias);
+	const entry = findTokenEntry(tokens || [], alias);
 
 	return entry ? parseResolvedShadow(entry.value) : null;
 }
--- a/src/extension/design-tokens/components/EditorShadowControl.js
+++ b/src/extension/design-tokens/components/EditorShadowControl.js
@@ -49,7 +49,7 @@
 import { BoxShadowControl, DEFAULT_COMPOSITE, parseResolvedShadow } from '../../../token-controls';
 import { TokenControlRow } from '../../token-indicators/components/TokenControlRow';
 import { isTokenAlias } from '../alias';
-import { SHADOW_TOKEN_KEY } from '../shadow-token';
+import { SHADOW_TOKEN_KEY, boundShadowToken } from '../shadow-token';
 
 /**
  * A CSS `rgba(r, g, b, a)` string, matching `hexToRGBA`'s own output format exactly (comma-space
@@ -237,6 +237,19 @@ function resolveShadowAlias(alias, tokens) {
 }
 
 /**
+ * The shadow token a stored native value is bound to.
+ *
+ * @param {?Array} native The stored native shadow attribute value.
+ *
+ * @since TBD
+ *
+ * @return {?string} The bound `{dot.alias}`, or null when the value carries no binding.
+ */
+export function shadowTokenOf(native) {
+	return boundShadowToken(native?.[0]);
+}
+
+/**
  * Convert the native `[{ color, opacity, hOffset, vOffset, blur, spread, inset }]` attribute value to
  * the composite `BoxShadowControl` edits.
  *
@@ -357,6 +370,12 @@ export function isUnsetShadow(native, defaultValue) {
 		return true;
 	}
 
+	// A binding, not the geometry, decides here: a token that resolves to a subtle or zero-offset
+	// shadow is still a deliberate pick, and reading it as unset would drop its name from the trigger.
+	if (shadowTokenOf(native)) {
+		return false;
+	}
+
 	// A preset shadow behind it makes an invisible shadow a deliberate suppression, not an absence.
 	if (defaultValue !== undefined && defaultValue !== null && defaultValue !== '') {
 		return false;
@@ -375,7 +394,8 @@ export function isUnsetShadow(native, defaultValue) {
  *
  * @param {Object}    props               The component props.
  * @param {string}    props.label         The control's label.
- * @param {?Array}    props.value         The native shadow attribute value.
+ * @param {?Array}    props.value         The native shadow attribute value, optionally carrying a
+ *                                        `shadowToken` binding.
  * @param {Function}  props.onChange      Called with the next native shadow attribute value.
  * @param {Array}     [props.tokens]      Pickable `shadow`-type tokens, `[{id, label, value, alias}]`.
  * @param {*}         [props.defaultValue] The active preset's own resolved shadow, or nothing when it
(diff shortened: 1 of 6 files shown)


pauloiankoski:
@coderabbitai review

src/extension/design-tokens/components/EditorShadowControl.js
```diff
@@ -362,11 +386,16 @@ export function EditorShadowControl({
...
+	// recognizes as a token: it names the token on the trigger, opens on the Style Library tab, and
+	// previews the token's own shadow. The stored legs stay untouched behind it.
+	const bound = shadowTokenOf(value);
+
 	return (
 		<TokenControlRow stacked>
 			<BoxShadowControl
 				label={label}
-				value={isUnsetShadow(value, defaultValue) ? '' : fromNativeShadow(value)}
+				value={isUnsetShadow(value, defaultValue) ? '' : bound || fromNativeShadow(value)}
```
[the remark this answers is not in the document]
pauloiankoski: Addressed in 0787bfc1e. `BoxShadowControl` gained an optional `fallbackShadow` prop, and `EditorShadowControl` passes `fromNativeShadow(value)` into it whenever a binding is present. Precedence is now: the alias's own token entry, then the host's fallback, then `DEFAULT_COMPOSITE`.

This closes a genuine inconsistency — a stale binding already falls back to the stored legs in both renderers, and the Custom tab was the one place it did not.

The prop is optional, so the Style Library host, which passes nothing, keeps the previous behavior. Four tests cover it: a stale alias with a fallback, a stale alias without one, a resolving alias where the fallback must not override the token entry, and the prop being passed down for a bound value.

pauloiankoski:
Rebased onto current `feature/design-tokens-module` (it had moved 22 commits since this branch was cut). @coderabbitai review

src/token-controls/controls/BoxShadowControl.js
```diff
@@ -231,19 +231,45 @@ function ShadowCustomTab({ shadow, onChange, renderColor, disabled = false }) {
...
 	// Only a real object is spread. A shadow can also arrive as a rendered STRING — the editor's capture
 	// flow writes one — and spreading a string splays it into indexed character keys, which the Custom
 	// tab would then write back as a shadow whose sub-fields are `0`, `p`, `x`. Degrading to the default
 	// shape reads as unset, which is wrong but harmless where the alternative cannot be saved at all.
 	const custom = typeof value === 'object' && value !== null && !Array.isArray(value) ? value : {};
-	const shadow = { ...DEFAULT_COMPOSITE, ...(aliased ? {} : custom) };
+	// An alias carries no legs of its own, so the Custom tab has nothing to seed from unless it borrows
+	// the bound token's own resolved shorthand here — without this, switching to Custom tab reads as
+	// an all-zero, transparent shadow and editing one leg discards the token's real value.
+	const aliasedEntry = aliased ? (tokens || []).find((entry) => entry.alias === value) : null;
```
d4mation: `findTokenEntry()` already does this lookup, and it is exported both from `helpers/token-summary` -- the module this file imports `defaultSummary`, `fieldSummary`, `hasValue`, `isTokenAlias` and `resolveDefaultValue` from at line 61 -- and from the package barrel at `src/token-controls/index.js:52`, so every call site below can reach it without a new dependency.

This is the third hand-rolled copy of `entry.alias === value`: here, `resolvePreviewCss()` at line 131, and `resolveShadowAlias()` in `EditorShadowControl.js:234`.

The two inside this file are exact swaps. Both are already behind an `isTokenAlias` guard -- `aliased` on this line, the `if (isTokenAlias(value))` branch at line 130 -- so `findTokenEntry()`'s `entry.fixed || isTokenAlias(value)` condition always passes, and its legacy-slug fallback cannot fire because `isLegacySlugFor()` matches an id's last segment, which never carries braces. Line 131 also gains the `tokens || []` guard it currently lacks.

`resolveShadowAlias()` is worth a second look rather than a blind swap. It is the one site NOT guarded by `isTokenAlias`: `toNativeShadow()` hands it any string, which is how the fixed "None" sentinel resolves -- it arrives as the literal `0px 0px 0px 0px transparent` and matches a `fixed: true` entry on `alias` equality. `findTokenEntry()` handles that identically (`entry.fixed` satisfies its guard), but it would additionally consult the legacy-slug branch for an unmatched non-alias string. That is inert against a shadow-role pool, which holds no `primitive.dimension.*` entries, so the swap is safe today -- just not safe for the same reason as the other two.

`matchFixedEntry()` at line 112 is deliberately left off this list: it compares `entry.value` against a rendered shorthand, not `entry.alias` against a value, so it is a different lookup and `findTokenEntry()` cannot stand in for it.

src/token-controls/controls/BoxShadowControl.js
```diff
@@ -231,19 +231,45 @@ function ShadowCustomTab({ shadow, onChange, renderColor, disabled = false }) {
...
+	// A stale alias (the token was deleted after the binding was saved) has no entry to resolve, but the
+	// host still stores the legs it was resolved to at save time — that snapshot is what the other
+	// renderers already fall back to for a stale binding, so the Custom tab seeds from it here too,
+	// rather than from the all-zero default.
+	const resolvedAliasShadow = aliasedEntry ? parseResolvedShadow(aliasedEntry.value) : fallbackShadow;
```
d4mation, line 286: The new `fallbackShadow` seam covers the Custom tab but not the preview square, so a stale binding is now inconsistent inside a single popover. `resolvePreviewCss()` (line 129) returns `''` when the alias matches no entry, so the preview renders shadow-less while `shadow` on this line has just been seeded from `fallbackShadow`, and the Custom tab shows those legs. `resolvePreviewCss()` already receives `shadow` as its third argument and ignores it on the alias branch; falling through to the composite shorthand it builds for a literal value when `entry` is null would make the preview agree with the tab it sits above.

src/token-controls/controls/BoxShadowControl.js
pauloiankoski: Done in 2ec4fbdff — all three swapped to `findTokenEntry()`, `matchFixedEntry()` left alone. Recorded your reasoning for the `resolveShadowAlias()` case in its docblock so it does not have to be re-derived.

src/token-controls/controls/BoxShadowControl.js
pauloiankoski: Fixed in 2ec4fbdff. `resolvePreviewCss()` now falls through to the composite shorthand when the alias resolves to no entry, so the preview matches the tab. `resolveHoverPreviewCss()` delegates to it and needed nothing.

pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai full review

d4mation (approved):

PR #1478, opened 2026-08-30
[DTM] SOFT-4269 [1/4]: forget a library's cached store state

pauloiankoski:
## Summary

Deleting or resetting a token library leaves the Style Library app holding that library's cached state: nothing in the store ever removes a keyed cache entry, so the palette and preset screens keep serving pre-delete data. This slice adds the store action that drops it. Nothing calls it yet — the wiring lands further up the stack.

`forgetLibrary(slug)` removes every entry addressed to a library across the six slug-keyed slices:

| Slice | Key shape | Slug position |
| --- | --- | --- |
| `presets` | `namespace::block::slug` | segment 2 |
| `paletteListings` | `namespace::slug` | segment 1 |
| `optimisticSwatchEdits` | `namespace::slug::paletteId` | segment 1 |
| `paletteBusy` | `namespace::slug` | segment 1 |
| `optimisticScaleEdits` | `slug` | whole key |
| `scaleBusy` | `slug` | whole key |

Two predicates read the slug back out by segment rather than by prefix or suffix, so a library whose slug is a prefix of another's can't be mistaken for it. The three key builders now join on a shared `KEY_SEPARATOR` so the builders and the predicates can't drift apart; the strings they produce are unchanged.

`state.feeds` is deliberately left alone. The delete flow replaces the feed it lands on moments later, and blanking it first would drop `feed.isReady` to false for a tick, which swaps the whole app for its loading skeleton and unmounts the delete modal mid-request. A dead entry for a removed slug can never be served: `useDesignTokensFeed` only advances its `slug` after a fresh fetch resolves.

`omitLibraryEntries` returns the original slice by reference when nothing matched, so forgetting a library with no entry in a given slice doesn't make `useSelect` see a changed result and re-render its consumers for nothing.

## Test plan

- `npm test -- src/style-library/store/reducer.test.js` — 22 passing, including four new cases: sibling libraries survive, `feeds` is untouched, unmatched slices come back by reference, and a block name is never mistaken for a slug.
- Full JS suite green (116 suites, 1730 tests).

## Stack

1. **This PR** — the store action
2. `library-reset/slice-2-workspace-reset-helper` — the workspace reset helper
3. `library-reset/slice-3-wire-library-flow-resets` — wiring both into the library flows

## Summary by CodeRabbit

- **New Features**
  - Added the ability to clear cached style-library data for a specific library.
  - Related feed data remains available after clearing the library cache.

- **Bug Fixes**
  - Improved cleanup of library-specific palettes, presets, editing states, and loading indicators.
  - Ensured data for other libraries remains unaffected.

- **Tests**
  - Added coverage for cache removal, sibling-library preservation, unchanged feeds, and library-name edge cases.

### DIFF
--- a/src/style-library/store/actions.js
+++ b/src/style-library/store/actions.js
@@ -59,6 +59,25 @@ export function receiveDesignTokensFeed(slug, feed) {
 }
 
 /**
+ * Forget every cached entry addressed to one library: its block presets, its palette listing, its
+ * pending optimistic overlays, and its busy flags.
+ *
+ * Dispatched after a library is deleted or reset, so the screens stop rendering values the server
+ * no longer holds. The feeds slice is deliberately untouched — the flow that dispatches this is
+ * about to replace the feed it lands on, and blanking it first would drop the whole app to its
+ * loading skeleton for a tick.
+ *
+ * @param {string} slug Token library slug.
+ *
+ * @since TBD
+ *
+ * @return {Object} The action object.
+ */
+export function forgetLibrary(slug) {
+	return { type: 'FORGET_LIBRARY', slug };
+}
+
+/**
  * Patch a swatch's label and/or value in the optimistic overlay, ahead of its write's response.
  *
  * @param {string} key   The palette listing key (`paletteListingKey(namespace, slug)`).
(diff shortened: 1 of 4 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1479, opened 2026-08-30
[DTM] SOFT-4266 [2/6]: resolve a bound shadow token in the editor canvas

pauloiankoski:
Second slice of the shadow-token binding stack. The slice below it records a picked shadow token on a `shadowToken` key and names it on the control; this one teaches the **editor canvas** to resolve that binding, so the button on the canvas tracks the token instead of a frozen literal.

Based on `SOFT-4266/slice-1-shadow-token-binding-model`.

## What changed

**One `shadowCss()` helper replaces nine hand-rolled shorthand builders.** `backend-styles/index.js` built its `box-shadow` value inline at nine sites — the base, transparent and sticky states, plus an inset/non-inset pair for each of their hover variants. Each was a nine-line string concatenation. They are now one call.

`shadowCss( shadowItem, blurFallback )` returns, in order:

1. `var(--kb-token--<id>)` when the item carries a `shadowToken` binding **backed by the active library** — the stored legs are not read at all, which is what makes the value track the token;
2. `''` when the binding is no longer backed, i.e. the token was deleted after the item was saved. The block then falls back to its own default CSS, matching what every other deleted-token reference already does. The stored legs are deliberately not used: they hold the value the token resolved to at pick time, not the current one. They still earn their place elsewhere — they keep the item a valid literal shadow for readers that do not know the binding key, and they seed the control's Custom tab;
3. otherwise the literal shorthand built from those legs, for an item that carries no binding at all.

**`hasVisibleShadow()` agrees with that rule.** A **backed** binding counts as visible — the gate cannot read a token's value, so reading it as invisible would let the caller's `box-shadow: none` reset erase a shadow the token does paint. An **unbacked** binding counts as invisible, so a stale binding takes exactly the path an item with no shadow already takes. Without that second half a stale binding would still enter the shadow branch and skip the reset, leaving the fallback in the wrong place.

**A pre-existing bug fixed along the way.** The six hover builders did not use `shadowAxisPx()`, so a `{dot.alias}` on a per-leg value emitted invalid CSS like `{semantic.shadow.card}px`. Routing them through `shadowCss()` closes that; it was never reachable from the three base-state builders, which already used the helper.

The binding reader is imported from `extension/design-tokens/shadow-token`, not from `EditorShadowControl` — that module is deliberately dependency-free so this style builder does not pull React, SCSS and the `token-controls` barrel into a module that only builds strings.

## Behavior preserved

`shadowCss()` is byte-identical to each builder it replaces for any item with no binding. Two things it deliberately does **not** do, because their callers own them:

- it never returns `none`. The literal `box-shadow: none` a "None" pick renders comes from the callers' existing `hasExplicitShadow ? … : 'none'` ternary, which is untouched. `shadowCss()` only ever builds a shorthand or a `var()`;
- it does not decide whether to emit at all. The surrounding `if` conditions, including the one that lets a preset's `var(--kb-btn-shadow)` survive, are unchanged.

## Test plan

- `npm test -- src/blocks/singlebtn`
- `shadowCss()` is covered for: a backed binding emitting the token var, an unbacked binding returning nothing, an unbound item building the literal shorthand with its `inset` prefix, the historic per-leg defaults filling missing axes, and a missing item producing no declaration.
- `hasVisibleShadow()` is covered for a backed binding with all-zero legs reading as visible, and an unbacked one reading as invisible.
- Also adds `pathOfAlias()` to `extension/design-tokens/alias.js`, the JS peer of PHP's `Alias::path_of()`, replacing five hand-rolled `slice(1, -1)` copies. It is pinned against the shared `token-alias-conformance.json` fixture so the two peers cannot drift. `color-literal.js` is deliberately left alone — it is guarded only by a truthy check, and `entry.alias` there can be a non-brace value, so the swap would not be equivalent.
- The pre-existing canvas tests are unchanged and still pass, which is what pins the replacement as behavior-preserving — a changed expected string would mean `shadowCss()` is not byte-identical to the builder it replaced.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved button shadow rendering for token-based and custom shadows, including inset shadows, transparency, hover states, and sticky styles.
  - Removed deleted or unavailable shadow bindings from rendered output while preserving valid design-token shadows.
  - Added reliable handling for incomplete shadow settings, including missing axes and shadow items.
  - Improved design-token alias recognition and resolution across dimensions, literals, and style settings.

### DIFF
--- a/src/blocks/singlebtn/components/backend-styles/index.js
+++ b/src/blocks/singlebtn/components/backend-styles/index.js
@@ -112,6 +112,12 @@ export function shadowAxisPx(raw, fallback) {
  * render nothing regardless of color, matching the value the "None" pick now writes and mirroring
  * the PHP renderer's `has_visible_shadow()`.
  *
+ * A `shadowToken` binding backed by the active library is visible outright — its real value lives
+ * in the token, which this gate cannot read, so reading it as invisible would let the base rule's
+ * `box-shadow: none` reset erase a shadow the token does paint. A binding the library no longer
+ * backs takes the same path as an item with no shadow at all: the stored legs are the value frozen
+ * at pick time, not the current one, so they are not consulted.
+ *
  * @param {?Object} shadowItem One `shadow[0]`-shaped item.
  *
  * @since TBD
@@ -123,11 +129,10 @@ export function hasVisibleShadow(shadowItem) {
 		return false;
 	}
 
-	// A bound item's real value lives in the token, which this gate cannot read. Counting it as visible
-	// keeps the base rule's `box-shadow: none` reset from erasing a shadow the token does paint —
-	// the same reasoning the per-leg alias branch below uses. Mirrors the PHP gate.
-	if (boundShadowToken(shadowItem)) {
-		return true;
+	const bound = boundShadowToken(shadowItem);
+
+	if (bound) {
+		return isBackedToken(pathOfAlias(bound));
 	}
 
 	return ['hOffset', 'vOffset', 'blur', 'spread'].some((axis) => {
@@ -151,9 +156,11 @@ export function hasVisibleShadow(shadowItem) {
  *
  * A `shadowToken` binding backed by the active library wins outright and the stored legs are never
  * read — that is what keeps the value tracking the token. A binding the library no longer backs (a
- * token deleted after the post was saved) falls back to those legs, which still hold the value the
- * token resolved to when it was picked. That differs on purpose from an unbacked PER-LEG alias, which
- * `shadowAxisPx()` cannot fall back for because the alias replaced the number outright.
+ * token deleted after the post was saved) renders nothing: the block falls back to its default CSS
+ * the same way every other block does when a token disappears, rather than the legs that still hold
+ * the value the token resolved to when it was picked — those legs stay stored for readers that do
+ * not know the binding key and to seed the Custom tab, but the renderer no longer reads them. An
+ * unbound item still builds its literal shorthand from the legs below.
  *
  * @param {?Object} shadowItem   One `shadow[0]`-shaped item.
  * @param {number}  blurFallback What `blur` defaults to when unset — 14 on every current caller,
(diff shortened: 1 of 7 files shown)


pauloiankoski:
@coderabbitai review

src/blocks/singlebtn/components/backend-styles/index.js
```diff
@@ -136,6 +146,52 @@ export function hasVisibleShadow(shadowItem) {
...
+ * @return {string} The `box-shadow` value, or '' when there is no item to render.
+ */
+export function shadowCss(shadowItem, blurFallback) {
+	if (!shadowItem) {
+		return '';
+	}
+
+	const bound = boundShadowToken(shadowItem);
+
+	if (bound && isBackedToken(bound.slice(1, -1))) {
```
d4mation: `bound.slice(1, -1)` is the JS side re-deriving what PHP exposes as `Alias::path_of()`, and `alias.js` has no peer for it. We should create a centralized helper for this instead of hand-rolling it as-needed.

d4mation: Related to https://github.com/stellarwp/kadence-blocks/pull/1480#discussion_r3895697354

Two changes here, mirroring the PHP:

1. This function: return `''` for a bound-but-unbacked item instead of dropping through to the leg builder, and have the call sites treat that the way they already treat "no shadow".
2. `hasVisibleShadow()` at line 129: it returns `true` for any binding regardless of backing, so a stale one still reaches the builder and still suppresses the base rule's `box-shadow: none` reset. It has to agree with the new rule.

Both need to land here rather than in #1480, so the canvas and the front end change together -- a fix on only one side would put them out of parity for exactly the case being fixed.

pauloiankoski: Added `pathOfAlias()` to `alias.js` in 76a245614, pinned against the shared conformance fixture. Converted five of the six copies; left `color-literal.js`, whose `entry.alias` can be a non-brace value, so the swap is not equivalent.

pauloiankoski: Both done in f05786574. `shadowCss()` returns `''` for an unbacked binding and `hasVisibleShadow()` now reports it invisible, so a stale binding takes the no-shadow path and no longer skips the reset. Call sites needed nothing — all nine were already gated.

pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai full review

d4mation (approved):

PR #1480, opened 2026-08-30
[DTM] SOFT-4266 [3/6]: resolve a whole-shadow token binding on the front end

pauloiankoski:
Third slice of the shadow-token binding stack. The slices below record a picked shadow token on a `shadowToken` key and resolve it in the editor canvas; this one makes the **front end** agree, so the rendered page tracks the token rather than a literal frozen at pick time.

Based on `SOFT-4266/slice-2-editor-canvas-shadow-css`.

## What changed

**`Kadence_Blocks_CSS::render_shadow()` resolves a whole-shadow binding.** A shadow item may carry a `shadowToken` key holding a `{dot.alias}`. When that token is backed by the active library the method returns its bare `var(--kb-token--<id>)` and reads no other key.

When the binding is **unbacked** — the token was deleted after the item was saved — it returns `false`, so the block falls back to its own default CSS. That matches what this method already does for an unbacked per-leg alias, and what every other deleted-token reference does. The stored legs are deliberately not used as a fallback: they hold the value the token resolved to at pick time, not the current one. No call site needs changing, because `add_rule()` guards with `if ( $value && ! empty( $value ) )` and drops the declaration rather than emitting `box-shadow:;`.

The key is exposed through `Kadence_Blocks_CSS::get_shadow_token_key()` over a `private const`, since it is read from another class.

**`has_visible_shadow()` agrees with that rule.** It now lives on `Kadence_Blocks_Abstract_Block`, so the guard lands there and every shadow-carrying block gets it. A **backed** binding counts as visible — the gate cannot read a token's value, and reading it as invisible would let a caller's `box-shadow: none` reset erase a shadow the token does paint. An **unbacked** binding counts as invisible. That second half is load-bearing: left visible, a stale binding would still enter the shadow branch and skip its `elseif ( ! $has_preset_shadow )` reset, putting the fallback in the wrong place.

Backing is established through a new public `Kadence_Blocks_CSS::is_token_reference_backed()`, which wraps the existing private `get_backed_token_reference()` rather than duplicating the check.

## Two ordering details that are load-bearing

**The binding is read before `normalize_shadow_defaults()` runs.** That helper does not filter — it returns a fresh seven-key array (`hOffset`, `vOffset`, `blur`, `spread`, `color`, `opacity`, `inset`), so it silently drops `shadowToken`. Since `render_button_shadow()` always passes a non-empty `$defaults`, reading the binding after normalizing would have made the whole feature dead on every button on the front end while still working in the editor. The lookup therefore sits above the normalize call, on the original array.

**The `empty( $shadow )` guard stays below the normalize call.** Splitting it out to the top looks like a tidy but is not behavior-neutral: `render_shadow( [], $non_empty_defaults )` has always normalized the empty array into a complete default item and rendered it, and the image block reaches that path whenever `boxShadow` is `[[]]`. The guard order is preserved, and a test pins it so the two are not merged back together later.

## Purely additive for the other blocks

`render_shadow()` is shared by nine blocks. The new branch fires only on a `shadowToken` key that no other block writes, and `normalize_shadow_defaults()` / `render_legacy_shadow()` are untouched, so every other caller's output is byte-identical.

## Test plan

- `slic run wpunit KadenceBlocksCssTokenEmissionTest`
- `slic run wpunit Blocks/SinglebtnTest`
- `slic run wpunit KadenceBlocksCssBcSnapshotTest` — the guard that the shared `render_shadow()` still renders every other block's shadows unchanged.
- Covered: a backed binding emitting the bare `var()` both with and without a `$defaults` array (the with-defaults case is the one that exercises the real button path); an unbacked binding returning `false` in both shapes; an empty item with caller defaults still rendering those defaults; a backed binding with all-zero legs counting as visible and an unbacked one not; the end-to-end button render emitting `box-shadow:var(--kb-token--…)`; and a stale binding reaching the `box-shadow: none` fallback.
- The existing "None" test still asserts `box-shadow: none`, unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shadow token handling so active, supported tokens render correctly.
  * Added support for whole-shadow token references.
  * Prevented stale or unresolved token references from producing invalid shadow styles.
  * Preserved visibility for valid zero-offset shadows and applied appropriate fallbacks when bindings are unavailable.

* **Tests**
  * Added coverage for supported, unresolved, and stale shadow token bindings.
  * Verified fallback rendering for empty shadow settings and shadows with defaults.

### DIFF
--- a/includes/class-kadence-blocks-css.php
+++ b/includes/class-kadence-blocks-css.php
@@ -25,6 +25,18 @@ use KadenceWP\KadenceBlocks\Psr\Log\LoggerInterface;
 class Kadence_Blocks_CSS {
 
 	/**
+	 * The shadow item's optional binding key: the {dot.alias} of the shadow token the value follows.
+	 * Kept alongside the numeric legs rather than replacing them, so an item stays a valid literal
+	 * shadow for readers that do not know this key and has something to fall back to when the bound
+	 * token is no longer in the active library.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const SHADOW_TOKEN_KEY = 'shadowToken';
+
+	/**
 	 * CSS to enqueue
 	 *
 	 * @var array
@@ -253,6 +265,17 @@ class Kadence_Blocks_CSS {
 		'md' => 'var(--global-kb-gap-md, 2rem)',
 		'lg' => 'var(--global-kb-gap-lg, 4rem)',
 	);
+
+	/**
+	 * The shadow item key that carries a whole-shadow token binding.
+	 *
+	 * @since TBD
+	 *
+	 * @return string The binding key.
+	 */
+	public static function get_shadow_token_key(): string {
+		return self::SHADOW_TOKEN_KEY;
+	}
 
 	/**
 	 * Instance Control
@@ -1646,6 +1669,10 @@ class Kadence_Blocks_CSS {
 	/**
 	 * Generates the shadow output.
 	 *
+	 * A `shadowToken` key holding a {dot.alias} binds the WHOLE shadow to a token: when that token is
+	 * backed by the active library the method returns its bare var() and reads no other key. Everything
+	 * else — including an unbacked binding — renders from the numeric legs as before.
+	 *
 	 * @param array                     $shadow   an array of shadow settings.
 	 * @param array<string, string|float> $defaults optional per-caller fallback literals used to fill
 	 *                                            any empty/missing leg before rendering. When omitted
@@ -1662,6 +1689,14 @@ class Kadence_Blocks_CSS {
 		}
 		if ( ! is_array( $shadow ) ) {
 			return false;
+		}
+		// A backed whole-shadow binding replaces the entire shorthand, so the legs below are never read.
+		// An unbacked one (a token deleted after the post was saved) falls through to them instead: unlike
+		// a per-leg alias, which displaced its number and leaves nothing to fall back to, the legs here
+		// still hold the value the token resolved to when it was picked.
+		$shadow_token_reference = $this->get_backed_token_reference( $shadow[ self::SHADOW_TOKEN_KEY ] ?? null );
+		if ( null !== $shadow_token_reference ) {
+			return $shadow_token_reference;
 		}
 		if ( ! isset( $shadow['color'] ) ) {
 			return false;
--- a/includes/class-kadence-blocks-css.php
+++ b/includes/class-kadence-blocks-css.php
@@ -1681,7 +1681,7 @@ class Kadence_Blocks_CSS {
 	 * @return string|false
 	 */
 	public function render_shadow( $shadow, array $defaults = [] ) {
-		if ( ! is_array( $shadow ) || empty( $shadow ) ) {
+		if ( ! is_array( $shadow ) ) {
 			return false;
 		}
 
@@ -1696,6 +1696,10 @@ class Kadence_Blocks_CSS {
 
 		if ( ! empty( $defaults ) ) {
 			$shadow = $this->normalize_shadow_defaults( $shadow, $defaults );
+		}
+
+		if ( empty( $shadow ) ) {
+			return false;
 		}
 
 		if ( ! isset( $shadow['color'] ) ) {
--- a/includes/class-kadence-blocks-css.php
+++ b/includes/class-kadence-blocks-css.php
@@ -1105,6 +1105,22 @@ class Kadence_Blocks_CSS {
 	}
 
 	/**
+	 * Whether a value is a `{dot.alias}` reference backed by the active library — the same backing check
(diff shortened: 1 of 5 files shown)


pauloiankoski:
@coderabbitai review

tests/wpunit/Blocks/SinglebtnTest.php
```diff
@@ -216,6 +216,33 @@ public function testExplicitShadowOverridesPresetShadow(): void {
...
+	/**
+	 * A button whose shadow is bound to a token emits that token's custom property as its box-shadow,
+	 * so editing the token in the Style Library moves every button that follows it.
+	 *
+	 * @return void
+	 */
+	public function testBoundShadowEmitsTokenVar(): void {
+		$output = $this->render_button(
+			[
+				'shadow' => [
```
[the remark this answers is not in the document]
pauloiankoski: Fixed in e36f4a2e1 — the fixture now sets `'displayShadow' => true`.

The cause is worth recording: this test predates the enable-flag gate, and this branch was rebased onto a `feature/design-tokens-module` that had moved 22 commits, which brought that gate underneath it. Without the flag the gate closes and the assertion could never have held.

I checked the rest of the file for the same gap after the rebase. Of the ten tests that pass a shadow value, this was the only one missing its flag — the others arrived alongside the gate and already set it.

includes/class-kadence-blocks-css.php
```diff
@@ -1654,15 +1681,27 @@ public function render_row_gap_property( $attributes, $name = array( 'gap', 'tab
...
-		if ( ! empty( $defaults ) && is_array( $shadow ) ) {
+		if ( ! is_array( $shadow ) ) {
+			return false;
+		}
+
+		// A backed whole-shadow binding replaces the entire shorthand, so the legs below are never read.
+		// An unbacked one (a token deleted after the post was saved) falls through to them instead: unlike
+		// a per-leg alias, which displaced its number and leaves nothing to fall back to, the legs here
+		// still hold the value the token resolved to when it was picked.
+		$shadow_token_reference = $this->get_backed_token_reference( $shadow[ self::SHADOW_TOKEN_KEY ] ?? null );
```
d4mation: This branch makes the whole-shadow binding live for every caller of `render_shadow()`, but only `kadence/singlebtn` has an editor-canvas half (#1479). `kadence/image` is the other block whose shadow control is `EditorShadowControl` (`src/blocks/image/image.js:1330`), so #1476's `toNativeShadow()` will write `shadowToken` onto its `boxShadow` item and `class-kadence-blocks-image-block.php:248` routes that item through this method, so the front end emits `var(--kb-token--<id>)` from here on. The image block's canvas does not: `src/blocks/image/image.js:1959-1974` still builds the shorthand inline from the frozen legs and never reads the binding key, so it paints the pick-time snapshot forever. The two agree at pick time and diverge the moment the token is edited which is the entire behavior this stack exists to deliver. The Image Block will need to be updated to support this as well.

d4mation: Also: There is an issue here if the Token Reference is deleted.

This doesn't match the behavior of other Blocks where instead the default Block CSS is used: https://github.com/stellarwp/kadence-blocks/pull/1266

To be clear about what should NOT change: the stored legs keep earning their place. They keep the item a valid literal shadow for readers that do not know the binding key, and they seed the Custom tab through the `fallbackShadow` prop added in #1476. Only the renderers should stop reading them for a stale binding. The logging half is already consistent, since `token_alias_is_backed()` logs under `WP_DEBUG` either way.

Two changes on this PR:

1. Here: return `false` for an unbacked binding instead of falling through to the legs. Nothing is needed at the call sites -- `add_rule()` guards with `if ( $value && ! empty( $value ) )`, so the empty declaration is dropped rather than emitted as `box-shadow:;`.
2. `class-kadence-blocks-abstract-block.php:738`: `has_visible_shadow()` returns `true` for ANY alias-shaped `shadowToken`, backed or not. Left that way, a stale binding still enters the shadow branch at `class-kadence-blocks-singlebtn-block.php:136` and therefore skips its `elseif ( ! $has_preset_shadow )` reset, so the fallback lands in the wrong place. It needs to stop counting an unbacked binding as visible.

pauloiankoski: Both done in 58c206c0f. `render_shadow()` returns `false` for an unbacked binding, and `has_visible_shadow()` now checks backing via a new public `is_token_reference_backed()` wrapping the existing private helper rather than duplicating it. JS half landed on #1479.

pauloiankoski: Addressed in the new #1489. `shadowAxisPx`/`shadowCss` are lifted into `extension/design-tokens/shadow-css.js` so neither block imports the other, and the image canvas now reads the binding.

pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai full review

d4mation (approved):

PR #1481, opened 2026-08-30
[DTM] SOFT-4269 [2/4]: clear a stale draft when the feed is swapped

pauloiankoski:
## Summary

Adds `resetWorkspace()`, the one function that returns the Style Library app to a clean, item-less view. Nothing calls it yet — the wiring lands in the next slice.

```js
resetWorkspace({ clearPublication, replace });
```

It clears the draft channel, then empties the route's `scope` and `item`.

**Why emptying the route is what discards the draft.** The settings panel is mounted only while `route.item` is non-empty, so clearing it unmounts the panel and the draft dies with it. The panel cannot simply be told to reseed: `useSettingsPanel` seeds its draft once per open item and deliberately ignores later value changes, which is what stops a save's feed refresh from clobbering a sibling panel's in-flight edit. Unmounting is the available lever.

**`screen` is deliberately left alone.** Every screen id exists in every library, so there is no reason to move the user off the screen they were reading. `scope` goes with `item` because it is that screen's own sub-selection (a palette id on the Color Palette screen) and it belongs to the library that is changing.

**`replace`, not `navigate`** — a route the user is being moved off must not enter browser history, or Back walks straight into the item that no longer exists.

**The channel is cleared before the route.** The unmounting panel's own effect cleanup also clears the channel, so doing it here first means the guard is empty regardless of how React orders those effects. A null `clearPublication` is tolerated, which is what `useDraftChannel()` returns when no provider is mounted (the dev gallery relies on this).

The function takes its collaborators as arguments rather than being a hook, for the reason `helpers/library-flows.js` gives for the same shape: it can then be exercised directly in a test without rendering the app.

## Test plan

- `npm test -- src/style-library/__tests__/workspace.test.js` — 4 passing: the exact `replace` payload, the absence of a `screen` key, the call ordering, and the null-channel path.
- Full JS suite green (117 suites, 1734 tests).

## Stack

2. **This PR** — the workspace reset helper
3. `library-reset/slice-3-wire-library-flow-resets` — wiring both into the library flows

## Summary by CodeRabbit

* **New Features**
  * Added workspace reset functionality that clears publication state and route details while preserving the current screen.

* **Tests**
  * Added coverage for workspace reset behavior, including route and channel clearing, operation order, and missing publication channels.

### DIFF
--- a/src/style-library/__tests__/workspace.test.js
+++ b/src/style-library/__tests__/workspace.test.js
@@ -20,7 +20,7 @@ describe('resetWorkspace', () => {
 		expect(replace.mock.calls[0][0]).not.toHaveProperty('screen');
 	});
 
-	it('clears the channel before the route, so an unmounting panel cannot republish a dirty draft', () => {
+	it('rewrites the route before clearing the channel, so a failed rewrite leaves the guard armed', () => {
 		const order = [];
 
 		resetWorkspace({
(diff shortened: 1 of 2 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1483, opened 2026-08-30
[DTM] SOFT-4269 [3/4]: reset the workspace when a library is deleted or switched

pauloiankoski:
## Summary

Fixes the reported bug: deleting a library did not reset the interface, kept every edit in memory, and then complained about unsaved changes on navigation.

### Root cause

`useSettingsPanel` seeds a panel's draft **once per open item** and deliberately never reseeds it when the persisted values change underneath. That one-shot rule is load-bearing — it stops a save's `refreshFeed` from clobbering a sibling panel's in-flight edit.

`deleteLibraryFlow` ended with `refreshFeed(...)` and nothing else. So the open panel's `initialValues` became the post-delete baseline while its draft still held the pre-delete edits, `computeIsDirty` went permanently true, and every guarded navigation opened the unsaved-changes modal about a draft with nothing left to save.

The existing stale-item self-heal only rescues an item that *vanished*. A token that still exists in the baseline (`semantic.color.primary`, say) kept its orphaned dirty draft, which is why the bug reproduced reliably rather than intermittently.

### The fix

`openLibraryFlow` and `deleteLibraryFlow` now take the two resets as injected callbacks, matching how they already take `refreshFeed` and `loadLibraries`:

- **`resetWorkspace()`** clears the draft channel and empties `scope`/`item`, unmounting the panel and taking the stale draft with it. Both flows call it.
- **`forgetLibrary(slug)`** drops the deleted library's cached store state. Delete only — switching destroys no data, and both libraries' caches stay valid.

Ordering is the point: both run **after** the delete resolves and **before** `refreshFeed`. Resetting earlier would throw away a draft for a request that might still fail; resetting later would let a screen render the fresh feed while still holding the deleted library's cached presets, palettes and pending overlays. `forgetLibrary` is called with the **deleted** slug, never the one the app lands on.

`useLibraries` gains a required third argument and builds the `forgetLibrary` callback. The store action empties the cached values; the two `invalidateResolutionForStoreSelector` calls are what make them refetch, since `@wordpress/data` otherwise still has each tuple marked resolved. `StyleLibraryApp` builds the reset and moves `useDraftChannelState()` above `useLibraries` so it exists in time.

Neither reset can report a failure as a failed operation: in `openLibraryFlow` the call sits inside the promise chain so the existing `.catch()` covers it, and in `deleteLibraryFlow` each is caught separately — the workspace reset still runs if forgetting the cached state throws first, because the reset is what actually stops the unsaved-changes complaint.

Neither parameter is defaulted. A default would let a caller silently skip the wiring, which is the bug class being fixed.

## Test plan

- `npm test -- src/style-library/__tests__/library-flows.test.js` — 32 passing, covering call ordering (`forget → reset → refresh`), the deleted-slug identity, and the three paths where neither reset may run (delete rejects, missing successor, activation fails).
- `npm test -- src/style-library/__tests__/use-libraries.test.js` — asserts real store state after a delete rather than spying on dispatch.
- Full JS suite green (117 suites, 1744 tests).

**Manual:** open a Spacing token, edit it without saving, delete the library being edited. The panel closes on its own, the screen shows baseline values, and navigating away prompts nothing.

## Stack

2. `library-reset/slice-2-workspace-reset-helper` — the workspace reset helper (#1481)
3. **This PR** — wiring both into the library flows

## Summary by CodeRabbit

* **Bug Fixes**
  * Opening or deleting a style library now resets the active workspace and draft state.
  * Deleted libraries no longer leave cached palettes or presets behind.
  * Cleanup errors no longer interrupt otherwise successful library deletions.
  * Unsaved drafts correctly block navigation until the workspace is reset.

### DIFF
--- a/src/style-library/helpers/library-flows.js
+++ b/src/style-library/helpers/library-flows.js
@@ -48,18 +48,25 @@ export function errorMessage(error) {
  * until someone explicitly activates a different one through `activateLibraryFlow`.
  *
  * @param {Object}   args
- * @param {string}   args.slug        The token library slug to open for editing.
- * @param {Function} args.refreshFeed Replaces the feed with a fresh REST read for a slug.
- * @param {Function} args.onBusy      Called with a boolean as the request starts and settles.
- * @param {Function} args.onError     Called with `{ message }` on failure.
+ * @param {string}   args.slug           The token library slug to open for editing.
+ * @param {Function} args.refreshFeed    Replaces the feed with a fresh REST read for a slug.
+ * @param {Function} args.resetWorkspace Clears the draft channel and the open route item, so the
+ *                                       library being left behind cannot strand a draft.
+ * @param {Function} args.onBusy         Called with a boolean as the request starts and settles.
+ * @param {Function} args.onError        Called with `{ message }` on failure.
  *
  * @since TBD
  *
  * @return {Promise<void>} Resolves once the feed refresh completes; rejects on failure, after
  *                          `onError`/`onBusy` have already run.
  */
-export function openLibraryFlow({ slug, refreshFeed, onBusy, onError }) {
+export function openLibraryFlow({ slug, refreshFeed, resetWorkspace, onBusy, onError }) {
 	onBusy(true);
+
+	// Before the read, not after: the open panel's draft belongs to the library being left, and
+	// the panel cannot reseed itself when the values change underneath it. Unmounting it first is
+	// what stops it reporting unsaved changes about a draft with nowhere left to go.
+	resetWorkspace();
 
 	return refreshFeed(slug)
 		.then(() => onBusy(false))
@@ -243,6 +250,8 @@ export function renameLibraryFlow({ slug, title, libraries, loadLibraries, onBus
  *                                        active non-default library.
  * @param {Function} args.refreshFeed     Replaces the feed with a fresh REST read for a slug.
  * @param {Function} args.loadLibraries   Refreshes the libraries list.
+ * @param {Function} args.forgetLibrary   Drops every cached entry addressed to a library slug.
+ * @param {Function} args.resetWorkspace  Clears the draft channel and the open route item.
  * @param {Function} args.onBusy          Called with a boolean as the request starts and settles.
  * @param {Function} args.onError         Called with `{ message }` on failure.
  * @param {Function} args.onActiveChanged Called with the slug that ends up active, when it moves.
@@ -258,6 +267,8 @@ export function deleteLibraryFlow({
 	successorSlug,
 	refreshFeed,
 	loadLibraries,
+	forgetLibrary,
+	resetWorkspace,
 	onBusy,
 	onError,
 	onActiveChanged,
@@ -293,6 +304,14 @@ export function deleteLibraryFlow({
 
 	return activation
 		.then(() => deleteLibrary(slug))
+		.then(() => {
+			// Both run only once the delete has actually landed, and both run before the feed is
+			// re-read. Resetting earlier would throw away an open draft for a request that might
+			// still fail; resetting later would let a screen render the fresh feed while still
+			// holding the deleted library's cached presets, palettes and pending overlays.
+			forgetLibrary(slug);
+			resetWorkspace();
+		})
 		.then(() => refreshFeed(nextSlug))
 		.then(() =>
 			// A failed refetch here must not undo a delete that already succeeded — by this point
--- a/src/style-library/helpers/library-flows.js
+++ b/src/style-library/helpers/library-flows.js
@@ -63,12 +63,18 @@ export function errorMessage(error) {
 export function openLibraryFlow({ slug, refreshFeed, resetWorkspace, onBusy, onError }) {
 	onBusy(true);
 
-	// Before the read, not after: the open panel's draft belongs to the library being left, and
-	// the panel cannot reseed itself when the values change underneath it. Unmounting it first is
-	// what stops it reporting unsaved changes about a draft with nowhere left to go.
-	resetWorkspace();
+	return Promise.resolve()
+		.then(() => {
+			// Before the read, not after: the open panel's draft belongs to the library being left,
+			// and the panel cannot reseed itself when the values change underneath it. Unmounting it
+			// first is what stops it reporting unsaved changes about a draft with nowhere left to go.
+			// Running it inside the chain (rather than before it starts) keeps a throw here inside
+			// this flow's own `.catch()`, instead of escaping synchronously and leaving `onBusy(false)`
+			// never called.
+			resetWorkspace();
 
-	return refreshFeed(slug)
+			return refreshFeed(slug);
+		})
 		.then(() => onBusy(false))
 		.catch((err) => {
 			onError({ message: errorMessage(err) });
@@ -309,8 +315,17 @@ export function deleteLibraryFlow({
 			// re-read. Resetting earlier would throw away an open draft for a request that might
 			// still fail; resetting later would let a screen render the fresh feed while still
 			// holding the deleted library's cached presets, palettes and pending overlays.
-			forgetLibrary(slug);
-			resetWorkspace();
+			//
+			// Wrapped in its own try/catch, not folded into the flow's outer `.catch()`, for the same
+			// reason the `loadLibraries()` step below swallows its own failure: by this point
+			// `deleteLibrary(slug)` has already resolved, so a throw from either callback (e.g.
+			// `resetWorkspace()`'s `history.replaceState`) must not be reported as a failed delete.
+			try {
+				forgetLibrary(slug);
+				resetWorkspace();
+			} catch {
+				// Intentionally swallowed — see above.
+			}
 		})
 		.then(() => refreshFeed(nextSlug))
 		.then(() =>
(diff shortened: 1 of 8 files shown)


pauloiankoski:
@coderabbitai review

src/style-library/helpers/library-flows.js
```diff
@@ -48,20 +48,33 @@ export function errorMessage(error) {
...
-	return refreshFeed(slug)
+	return Promise.resolve()
+		.then(() => {
+			// Before the read, not after: the open panel's draft belongs to the library being left,
+			// and the panel cannot reseed itself when the values change underneath it. Unmounting it
+			// first is what stops it reporting unsaved changes about a draft with nowhere left to go.
+			// Running it inside the chain (rather than before it starts) keeps a throw here inside
+			// this flow's own `.catch()`, instead of escaping synchronously and leaving `onBusy(false)`
+			// never called.
+			resetWorkspace();
```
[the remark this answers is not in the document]
pauloiankoski: The failure mode is real for this PR read on its own, but the proposed reordering would reintroduce the bug this branch exists to fix, and the stack closes the gap one slice up. Leaving the order as is.

**Why the reset cannot move after the refresh.** `useSettingsPanel` seeds a panel's draft once per open item and deliberately never reseeds it when the persisted values change underneath — that rule is what stops a save's `refreshFeed` from clobbering a sibling panel's in-flight edit. So any window where the feed has swapped while the panel is still mounted is a window where the panel holds a draft that can never converge, reports `isDirty` forever, and guards navigation about a library that is no longer open. That is precisely the reported bug. Validating the destination first and clearing afterwards makes that window the normal path rather than an edge case.

**Why the draft is not actually lost in the merged result.** `library-reset/slice-4-guard-library-switch`, the next PR in this stack, routes the dropdown through `channel.guard()`, so a dirty draft produces a Save/Discard prompt *before* this flow runs. By the time `openLibraryFlow` executes, the user has already decided: saved, or deliberately discarded. A failed `refreshFeed` after that point cannot lose anything the user still wanted.

**What a failed open does leave behind:** the route's `scope`/`item` cleared while the app stays on the current library — the panel closes and the screen shows its list view. `openError` surfaces the failure. That state is coherent rather than broken, and reopening the item is one click.

The genuinely atomic version — fetch and validate the destination feed, then reset, then commit the feed — needs `refreshFeed` split into fetch and commit phases. That is the heavy lift this finding names, it touches a hook with its own documented race handling around `latestRequestRef`, and it buys nothing once the guard above is in place.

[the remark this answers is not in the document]
pauloiankoski: Confirming the resolution for whoever reads this thread later: the gap is real for this PR in isolation and is closed by #1484, the next slice. These four PRs are registered as GitHub stack #1485 and land as one unit, so the failure path never reaches a released state. Reviewing #1483 on its own, treat the discarded draft on a failed open as a known, one-slice-long window rather than shipped behavior.

src/style-library/hooks/use-libraries.js
```diff
@@ -112,6 +114,41 @@ export function useLibraries(feed, refreshFeed) {
...
+	 * only one library is open at a time, so no other slug's tuples are mounted to re-resolve.
+	 *
+	 * Called while the app is still pointed at the library being deleted, so a hard delete
+	 * re-arms these resolvers against a slug the server no longer has, and the browser fires one
+	 * guaranteed-404 request as a result. That is expected, not a bug: its error lands on a
+	 * `getPaletteListing`/`getBlockPresets` tuple keyed to the deleted slug, which nothing reads
+	 * again once the feed moves to the next library moments later. Moving this call to after the
+	 * feed re-read would avoid the request, but would let a screen render the fresh feed while
+	 * still serving the deleted library's cached presets and palettes until the invalidation
+	 * caught up — worse than a 404 nothing acts on.
```
d4mation: The docblock claims the delete-time 404 "lands on a tuple keyed to the deleted slug, which nothing reads again", but `forgetLibrary()` dispatches `invalidateResolutionForStoreSelector('getPaletteListing')` while `useDesignTokensFeed`'s `slug` is still the deleted library (it only advances once `refreshFeed(nextSlug)` resolves), and `AppShell`'s `isBlocked` renders a scrim rather than unmounting, so `ColorPaletteScreen` stays mounted. The resolver refires, 404s, and `use-palettes.js:190-194` copies the error into a `useState` slot via `setOpenError(...)`. `ColorPaletteScreen` is the same element type in the same position across the swap, so React does not remount it and that state survives. The user lands on the next library screen with a stale error Notice rendered at `ColorPaletteScreen.js:208` / `:343`.

src/style-library/helpers/library-flows.js
```diff
@@ -48,20 +48,33 @@ export function errorMessage(error) {
...
+
+			return refreshFeed(slug);
+		})
 		.then(() => onBusy(false))
 		.catch((err) => {
```
d4mation, line 79: A throw from `resetWorkspace()` is routed into `openLibraryFlow`'s `.catch()` and reported as a failed open. This is the exact opposite of `deleteLibraryFlow` (lines 326-336), which deliberately swallows the same throw. If `replace()`'s `history.replaceState` throws (example: Safari's rate-limit SecurityError) after `clearPublication()` has already succeeded, the open is aborted AND the still-mounted panel's dirty draft is invisible to `guard()`. `ScaleSettings`'s publish effect (deps `[publish, clearPublication, id, token, panel.draft, panel.isDirty]`) will not re-fire until the next keystroke, so the next navigation discards it with no prompt.

src/style-library/hooks/use-libraries.js
pauloiankoski: Confirmed, and the first fix I wrote for it did not actually work — thanks for the precise trace, it is what made that visible.

**What I tried first.** I split `forgetLibrary` into the dispatch alone and a new `revalidateLibraryCaches`, and moved the invalidation to after `refreshFeed(nextSlug)` resolves, on the theory that the screen would be on the next library by then.

That theory is wrong. `refreshFeed` advances the visible library with `setSlug(targetSlug)` *inside* its fetch `.then` (`use-design-tokens-feed.js:144-155`) — a React state update, not a synchronous value. The flow's next `.then` runs on the following microtask while that render is still pending, and `use-palettes.js:106` reads its slug from `feed?.slug`, which is still the deleted library until the render commits. Since `useSelect` re-runs `mapSelect` synchronously when a dispatch notifies the store, the resolver can still refire against the deleted slug. Reordering narrows the window; it cannot close it.

**What actually fixes it** is the mechanism you identified — the effect only ever sets:

```js
// An error belongs to the library it came from.
useEffect(() => {
	setOpenError(null);
}, [namespace, slug]);
```

Declared immediately above the failure effect, so on a swap that genuinely fails the clear runs first and the new library's own error still lands. Not folded into the existing effect, because clearing there on `listingFailure === null` would also wipe errors the palette write flows put into the same slot. This also fixes the general case: any listing failure previously followed you onto every library you opened afterwards, for the life of the mount.

I kept the `revalidateLibraryCaches` split, but on honest grounds rather than the ones I first claimed: re-arming resolvers for a library the app is leaving is wasted work and after the feed read is the right place for it. The docblock now says that, and says explicitly that it is ordering hygiene rather than a guarantee, with a pointer to where the real protection lives. The paragraph asserting the error "lands on a tuple nothing reads again" is gone — it was the wrong claim that hid this.

Added `use-palettes.test.js` coverage for the swap, using the existing probe harness.

src/style-library/helpers/library-flows.js
pauloiankoski: Fixed, though not by making `openLibraryFlow` swallow the throw — the asymmetry you spotted is real, but swallowing would have left the worse half of the problem in place. Even with the open reported as successful, the panel would still be mounted holding a draft that `guard()` can no longer see, and as you note its publish effect will not re-fire until the draft changes again.

The order inside `resetWorkspace` was the actual defect. It cleared the channel first and rewrote the route second, so the only step that can fail ran *after* the state it protects had already been discarded. Swapped:

```js
export function resetWorkspace({ clearPublication, replace }) {
	// The route rewrite goes first because it is the step that can fail. `replace()` reaches
	// `history.replaceState`, which a browser may refuse (Safari rate-limits it), and the panel only
	// unmounts if the route actually changed. Clearing the channel first would, on that failure,
	// leave a mounted panel whose dirty draft is invisible to `guard()` — the next navigation would
	// discard it silently, and the panel's own publish effect will not re-fire until its draft
	// changes again. Failing before the channel is touched leaves the workspace exactly as it was.
	replace({ scope: '', item: '' });

	if (clearPublication) {
		clearPublication();
	}
}
```

The invariant is now: the publication is never cleared unless the route actually changed. Both calls still run in the same synchronous tick, so React batches them into one render exactly as before — nothing changes on the success path.

That also settles the asymmetry in the direction I think you want. With the reorder, a throwing `replace` in `openLibraryFlow` means nothing was touched and `refreshFeed` never ran, so the open genuinely did fail and reporting it is correct. `deleteLibraryFlow` still swallows, because there the delete has already landed and must not be reported as failed.

Pinned by a test that fails against the old order:

```js
expect(clearPublication).not.toHaveBeenCalled();
```

This lands in `library-reset/slice-2-workspace-reset-helper` (#1481), since that is where `resetWorkspace` is defined; slices 3 and 4 were rebased onto it, so all three branches were force-pushed.

d4mation (approved):

PR #1484, opened 2026-08-30
[DTM] SOFT-4269 [4/4]: ask before discarding a draft when switching libraries

pauloiankoski:
## Summary

Switching to another library — or creating one, which switches to it — discarded an open settings panel's unsaved draft without asking. Switching *screens* prompts. This makes the two consistent.

`LibrarySelector` now consumes the draft channel and routes both library-switching entry points through `channel.guard()`, using the idiom already established in `ScaleScreen.js`:

```js
const open = () => onOpen(slug).catch(() => {});

channel ? channel.guard(open) : open();
```

The `channel ?` fallback is not defensive padding — `useDraftChannel()` returns `null` when no provider is mounted, which the dev gallery relies on.

**Two entry points, because creating a library also switches to it.** The create flow ends by opening the new library, so it swaps the feed like any other switch. The guard sits on the trailing action that opens the create modal rather than around `onCreate`, so the unsaved-changes prompt never stacks on top of the create modal.

**Deleting stays deliberately unguarded.** Destroying a library makes its draft moot, so prompting there would be nonsense — the same reasoning `ScaleSettings.js` gives for its own unguarded delete. This PR does not touch that path.

**Why the prompt is the right call here.** Unlike a delete, the token being edited still exists in the library being left, so the edit is worth offering to save. This also closes the failure mode raised on the previous PR: with the guard in place, a `refreshFeed` that fails after the workspace reset cannot lose anything, because the user has already chosen save or discard before the flow runs.

## Test plan

- `npm test -- src/style-library/__tests__/library-selector.test.js` — 5 passing, covering: no channel mounted (opens directly); channel mounted (asserts `onOpen` is *not* called until the guarded callback runs, then that it fires with the right slug); a rejected open still swallowed; create guarded; create with no channel.
- Full JS suite green (118 suites, 1749 tests).

**Manual:** edit a Spacing token without saving, then pick another library from the header dropdown. You are asked to save or discard instead of losing the edit silently.

## Note for the reviewer

The local pre-push PHP suites were skipped for this push: the shared slic test database is currently wedged (deadlock, then missing tables) and fails 249 unrelated `Optimizer` tests on branches that pushed green earlier today with identical PHP. This PR changes no PHP at all. CI's `tests.yml` is the authoritative run.

## Stack

2. `library-reset/slice-2-workspace-reset-helper` — the workspace reset helper (#1481)
3. `library-reset/slice-3-wire-library-flow-resets` — wiring both into the library flows (#1483)
4. **This PR** — prompt before discarding a draft on switch

## Summary by CodeRabbit

- **Bug Fixes**
  - Added protection against losing unsaved drafts when switching libraries or opening the create-library dialog.
  - Actions proceed immediately when no draft channel is active and are safely deferred when unsaved changes may be affected.
  - Preserved graceful handling when a guarded library switch is declined or fails.

### DIFF
--- a/src/style-library/__tests__/library-selector.test.js
+++ b/src/style-library/__tests__/library-selector.test.js
@@ -1,0 +1,205 @@
+/* eslint-env jest */
+/**
+ * External dependencies
+ */
+import { act, createElement } from 'react';
+import { createRoot } from 'react-dom/client';
+
+/**
+ * Internal dependencies
+ */
+import { LibrarySelector } from '../components/organisms/LibrarySelector';
+import { useDraftChannel } from '../hooks/use-draft-channel';
+
+// Same convention `scale-settings.test.js` uses for this hook: a bare `jest.fn()` stand-in whose
+// return value each test sets, rather than automocking the module.
+jest.mock('../hooks/use-draft-channel', () => ({
+	useDraftChannel: jest.fn(),
+}));
+
+// `SelectDropdown` and `CreateLibraryModal` both render `@wordpress/components` controls, which
+// ship their own nested `react` copy and trip React's "Invalid hook call" guard under this test's
+// top-level renderer (see `scale-settings.test.js`'s identical note). Stand-ins that expose only
+// what these tests need — a change trigger and the trailing action's click — keep the tests about
+// guard routing rather than dropdown or modal internals.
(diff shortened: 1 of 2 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1486, opened 2026-08-30
[DTM] SOFT-4266 [5/6]: restore the shipped shadow defaults and drop the shadow deprecation

pauloiankoski:
Final slice of the shadow-token binding stack, and the one piece of it that is purely backwards compatibility. It restores the block's shipped shadow value defaults, teaches the inspector to read a lowered enable flag as unset, and removes a deprecation that has become actively harmful.

Based on `SOFT-4266/slice-4-flag-follows-binding`.

## A legacy button silently lost its shadow

The six shadow value attributes had been changed from their shipped defaults to an all-zero transparent shadow. Compare against `master`:

```
shadow / shadowTransparent / shadowSticky
  master: [{"color":"#000000","opacity":0.2,"spread":0,"blur":2,"hOffset":1,"vOffset":1,"inset":false}]
shadowHover / shadowTransparentHover / shadowStickyHover
  master: [{"color":"#000000","opacity":0.4,"spread":0,"blur":3,"hOffset":2,"vOffset":2,"inset":false}]
```

WordPress omits an attribute equal to its registered default, so a button whose shadow was enabled but never customized saved its flag and **no value key at all**. `save()` returns `null`, so PHP renders straight from stored attributes and an editor-only migration never runs for a post nobody reopens. Such a button merged the all-zero default server-side and rendered nothing, where the shipped plugin renders `1px 1px 2px 0px rgba(0, 0, 0, 0.2)`.

All twelve shadow-related attributes — the six enable flags and the six values — now deep-equal `master`'s schema again. A test pins that, so the defaults cannot drift back unnoticed.

The enable flag is what keeps a visible default safe: an untouched button has its flag `false`, so nothing renders despite the default. That is covered on both the front end and the editor canvas.

## The inspector reads a lowered flag as unset

With a visible default restored, an untouched button's stored value has visible geometry, so the control would have read "Custom" on a brand-new button. `isUnsetShadow()` now returns true when the flag is lowered — a lowered flag genuinely means "no shadow set here". The ordering inside it is load-bearing and covered by tests: no item at all is unset; a lowered flag is unset; a binding present with the flag raised is **set**; otherwise the existing preset/geometry logic decides.

`enabled` defaults to `true`, so `kadence/image`, which uses the same control and has no such flag, is unaffected.

## The shadow deprecation is removed, because it had turned destructive

`deprecated.js` existed to supply historical value defaults to old content. With the value defaults restored and the flags present, the current schema **is** the schema that deprecation reconstructs — `deprecatedAttributes` and `metadata.attributes` were measured as identical in keys and values, so old content parses natively and there is no shape left to migrate.

Leaving it in place would not have been neutral. `migrate()` deletes every `display*` flag and rewrites the value of any pair whose flag is not exactly `true` to an all-zero item, and `isEligible` fires whenever any `display*` key is present. Now that the flags are part of the schema and the inspector writes them, a saved button with a raised flag would hit that on parse: all six flags dropped and five other states' stored shadows zeroed. It is deleted along with its test and its registration in `block.js`.

What it was protecting is now covered directly rather than by migration: a legacy button with its flag raised and no stored value key still renders the shipped shadow, pinned on the front end, on the canvas, and at the schema itself.

## Test plan

- `npm test -- src/blocks/singlebtn src/extension/design-tokens`
- `slic run wpunit Blocks/SinglebtnTest`
- Covered: an untouched button emitting no shadow despite the visible default; a legacy button with its flag raised and no stored value rendering the shipped shadow; the four `isUnsetShadow()` ordering cases including bound-plus-raised and bound-plus-lowered; and the six value defaults matching the shipped schema.
- Nothing references the removed module: `singlebtn/deprecated`, `NONE_SHADOW_ITEM` and `LEGACY_SHADOW_DEFAULTS` have no remaining hits in `src/` or `tests/`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable shadow enablement for normal, hover, transparent, and sticky button states.
  * New buttons include polished default shadow settings while keeping shadows disabled until enabled.

* **Bug Fixes**
  * Fixed shadow rendering for disabled, token-backed, and stale shadow values.
  * Preserved expected shadows for existing buttons with legacy enabled-shadow settings.

* **Tests**
  * Expanded coverage for shadow defaults, token bindings, and enabled/disabled rendering behavior.


pauloiankoski:
@coderabbitai review

src/extension/design-tokens/components/EditorShadowControl.js
```diff
@@ -424,6 +437,7 @@ export function EditorShadowControl({
 	defaultValue,
 	renderColor,
 	disabled = false,
+	enabled = true,
```
d4mation, line 452: The new `enabled` prop is wired into all six `kadence/singlebtn` call sites but not into `kadence/image`, which has the identical shape and the identical problem. `src/blocks/image/image.js:1330` passes no `enabled`, yet both of its renderers gate on the flag. The canvas at `image.js:1960` (`displayBoxShadow && hasVisibleShadow(...)`) and the front end at `class-kadence-blocks-image-block.php:241`. So a legacy image saved with the old toggle switched OFF but real values still behind it reads as "Custom" in the inspector while both renderers paint nothing. That is exactly the inspector/renderer disagreement this prop was introduced to close; `enabled={displayBoxShadow}` there finishes the job.

src/blocks/singlebtn/block.json
```diff
@@ -342,7 +342,15 @@
...
+					"color": "#000000",
+					"opacity": 0.2,
+					"spread": 0,
+					"blur": 2,
+					"hOffset": 1,
+					"vOffset": 1,
+					"inset": false
+				}
 			]
 		},
```
d4mation, line 355: Worth flagging while this reasoning is fresh, though it is outside this PR's diff: `kadence/image` still has the regression this PR reverses for the button, and it is half-fixed in a way that hides it.

`3c6512be6` ("refactor(image): remove the Enable Box Shadow toggle, decide emission from the value") did two things to `src/blocks/image/block.json` in one move: it dropped the `displayBoxShadow` attribute, and it lowered `boxShadow`'s registered default from the shipped `{ "color": "#000000", "opacity": 0.2, "spread": 0, "blur": 14, "hOffset": 0, "vOffset": 0, "inset": false }` to the all-zero transparent composite. Its reasoning was that "a fresh image renders exactly as it did when a `false` boolean suppressed the old visible default" -- true for a fresh image, but it is the saved-content case that breaks. WordPress omits any attribute equal to its registered default, so an image switched ON but never customized stored `displayBoxShadow: true` and NO `boxShadow` key at all. Parsed against the lowered default it now resolves to the transparent composite, `has_visible_shadow()` reads it as invisible, and the shadow is gone. Exactly the shape this PR's `1d0f8c9c5` restores for the button.

`20c6b89f6` ("fix(image): keep a shadow switched off before the toggle was removed switched off") then walked back the flag half -- it re-added the `displayBoxShadow` declaration and restored the gate -- but left `boxShadow`'s value default lowered. So the pair is inconsistent: the flag is back, the value it gates is not. That is why the bug survived a commit aimed at this very problem.

Nothing downstream catches it, either. `kadence/singlebtn` is fully dynamic, which is what let this PR drop its deprecation outright; `kadence/image` has a real `save()`, so `src/blocks/image/deprecated.js` only runs on a markup mismatch -- it has no `isEligible`, and it carries no shadow migration. An affected image is only repaired if someone reopens and re-edits it.

The fix is the same one-line revert this PR makes: restore `boxShadow`'s default to the shipped visible value and let the already-restored `displayBoxShadow` flag keep a fresh image clean, which is the two-halves-together contract `src/blocks/singlebtn/__tests__/shadow-defaults.test.js` pins here.

src/extension/design-tokens/components/EditorShadowControl.js
pauloiankoski: Wired up in #1489 — the image control now passes `enabled={displayBoxShadow}`.

src/blocks/singlebtn/block.json
pauloiankoski: Confirmed against `master` and fixed in #1489: `displayBoxShadow` already matched, `boxShadow`'s default was still lowered, `dropShadow` untouched. Restored it. Also checked `image/deprecated.js` — save-only, no `isEligible`, no shadow migration, as you said.

pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai full review

src/extension/design-tokens/components/EditorShadowControl.js
```diff
@@ -441,7 +455,7 @@ export function EditorShadowControl({
 		<TokenControlRow stacked>
 			<BoxShadowControl
 				label={label}
-				value={isUnsetShadow(value, defaultValue) ? '' : bound || fromNativeShadow(value)}
+				value={isUnsetShadow(value, defaultValue, enabled) ? '' : bound || fromNativeShadow(value)}
```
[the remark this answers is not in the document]
pauloiankoski: Not applying this one. Passing an empty value would stop `fallbackShadow` being consulted — it only applies on the aliased branch — so the Custom tab would show zeros again for a deleted token, which is the regression #1476 fixed.

That was an explicit request on review: keep the legs seeding the Custom tab, and have only the renderers stop reading them. The panel-vs-page difference is that rule's intended consequence.

The related half was real and is fixed in 74f1699a1: the derived enable flag no longer stays raised for an unbacked binding.

pauloiankoski:
On the Prettier nitpick for `src/blocks/singlebtn/__tests__/shadow-defaults.test.js` lines 23-24 — `npx prettier --check` on that file reports `All matched files use Prettier code style!`, so there is nothing to reformat. Leaving it as is.

d4mation (approved):

PR #1489, opened 2026-08-31
[DTM] SOFT-4266 [6/6]: bring the image block's shadow up to the same binding and defaults

pauloiankoski:
Final slice of the shadow-token binding stack. The slices below build the binding for `kadence/singlebtn`; this one brings `kadence/image` — the other block whose shadow control is `EditorShadowControl` — to the same state, and repairs a shadow regression it is carrying today.

Based on `SOFT-4266/slice-5-shipped-shadow-defaults-v2`.

## The image canvas ignored the binding while its front end honored it

Because both blocks share `EditorShadowControl` and `Kadence_Blocks_CSS::render_shadow()`, the slices below already made half of this live for images: `toNativeShadow()` writes `shadowToken` onto the `boxShadow` item, and the front end emits `var(--kb-token--<id>)` for it. The canvas did not — it built the shorthand inline from the frozen legs and never read the binding key, so the two agreed at pick time and diverged the moment the token was edited.

`shadowAxisPx()` and `shadowCss()` are lifted out of `src/blocks/singlebtn/components/backend-styles/index.js` into a new `src/extension/design-tokens/shadow-css.js`, beside `shadow-token.js`. Neither block reaches into the other, and the module stays a plain string builder — no React, no control barrel. `shadowCss()` gains an `opacityFallback` argument because the image's historic opacity default is `0.2` where the button's is `1`; existing callers are unchanged.

The image canvas now goes through `src/blocks/image/box-shadow.js`, which keeps its gate, its blur-14 / opacity-0.2 leg defaults and its `inset` prefix, and is testable without rendering the editor.

One behavior change falls out of this: a per-leg `{dot.alias}` on an image shadow now resolves through `tokenPx()` instead of being concatenated raw. The old output was `{semantic.shadow.card}px` — invalid CSS — so this fixes a latent bug and matches what the front end and the button already do.

## The `enabled` prop is wired into the image control

`EditorShadowControl` gained an `enabled` prop in the slice below so a lowered enable flag reads as unset rather than "Custom". `kadence/image` has the identical shape and both of its renderers gate on `displayBoxShadow`, but the control was passing no `enabled` — so a legacy image saved with the toggle switched off but real values behind it read as "Custom" in the inspector while both renderers painted nothing. It now passes `enabled={displayBoxShadow}`.

## `boxShadow`'s registered default was left lowered while its flag was restored

This is a live regression on the feature branch, not something this stack introduced.

`3c6512be6` removed the `displayBoxShadow` attribute **and** lowered `boxShadow`'s registered default from the shipped visible value to an all-zero transparent composite. `20c6b89f6` later walked back the flag half — it re-added the attribute and restored the gate — but left the value default lowered. The two halves have been inconsistent since.

That breaks saved content. WordPress omits an attribute equal to its registered default, so an image switched **on** but never customized stored `displayBoxShadow: true` and no `boxShadow` key at all. Parsed against the lowered default it resolves to the transparent composite, `has_visible_shadow()` reads it as invisible, and the shadow is gone.

Measured against `master`:

| attribute | vs shipped |
|---|---|
| `displayBoxShadow` | already restored |
| `boxShadow` | **still lowered** |
| `displayDropShadow` / `dropShadow` | untouched |

`boxShadow`'s default is restored to the shipped value. The already-restored `displayBoxShadow` flag is what keeps a fresh image clean — both renderers gate on it, so a visible default cannot make a new image sprout a shadow. This is the same two-halves-together fix the slice below makes for the button.

Nothing downstream would have caught it: `kadence/image` has a real `save()`, and `src/blocks/image/deprecated.js` is save-only with no `isEligible`, no `migrate` and no shadow handling, so an affected image is repaired only if someone reopens and re-edits it.

## Test plan

- `npm test -- src/blocks/image src/blocks/singlebtn src/extension/design-tokens`
- `slic run wpunit Blocks/ImageTest`
- Covered: a fresh image with the flag lowered emitting no shadow despite the restored visible default, on the canvas and the front end; a legacy image with the flag raised and no stored value rendering the shipped shadow; the canvas emitting the token `var()` for a backed binding, nothing for an unbacked one, and the stored legs for an unbound item; and `boxShadow`'s default matching the shipped schema so it cannot drift again.
- The button's canvas tests move to the lifted module's import path with no expectation changes, which is what pins the lift as behavior-preserving.

## Summary by CodeRabbit

* **New Features**
  * Added support for image box shadows with configurable visibility, opacity, blur, color, and inset styling.
  * Updated the default image shadow appearance to a subtle black shadow.

* **Bug Fixes**
  * Improved handling of legacy and missing shadow settings.
  * Prevented disabled or unresolved shadows from being rendered.
  * Preserved compatibility with existing shadow token values.

* **Tests**
  * Added comprehensive coverage for image shadow defaults, visibility, token resolution, and legacy behavior.

### DIFF
--- a/src/extension/design-tokens/shadow-css.js
+++ b/src/extension/design-tokens/shadow-css.js
@@ -1,0 +1,99 @@
+/**
+ * A stored shadow item as a `box-shadow` declaration value.
+ *
+ * This lives beside `shadow-token.js` rather than inside either block that needs it: `kadence/image`
+ * and `kadence/singlebtn` both build a canvas `box-shadow` from the same `shadow[0]` shape and must
+ * honor a binding identically, and neither block should have to reach into the other for the builder.
+ * Like its neighbor it only builds strings, so it pulls in no React and no control barrel.
+ */
+
+/**
+ * External dependencies
+ */
+import { KadenceColorOutput } from '@kadence/helpers';
+
+/**
+ * Internal dependencies
+ */
+import { tokenPx } from './token-px';
+import { pathOfAlias, resolveTokenAlias } from './alias';
+import { isBackedToken } from './backed-tokens';
+import { boundShadowToken } from './shadow-token';
+
+/**
+ * One shadow axis as a bare pixel number.
+ *
+ * A {dot.alias} leg is resolved through the token pool the way the PHP renderer's `render_shadow()`
+ * does. Concatenated raw it would emit `{alias}px`, which is not valid CSS — and `hasVisibleShadow()`
+ * deliberately counts such a leg as visible, so it does reach here.
+ *
+ * @param {*} raw      The stored axis value.
+ * @param {*} fallback What this axis defaults to when unset.
+ *
+ * @since TBD
+ *
+ * @return {*} The axis value to serialize.
+ */
+export function shadowAxisPx(raw, fallback) {
+	if (typeof raw === 'string' && raw.trim() !== '' && !Number.isFinite(Number(raw))) {
+		const resolved = tokenPx(raw);
+
+		return resolved === null || resolved === undefined ? fallback : resolved;
+	}
+
+	return undefined !== raw && null !== raw ? raw : fallback;
+}
+
+/**
+ * One shadow item as a `box-shadow` declaration value.
+ *
+ * A `shadowToken` binding backed by the active library wins outright and the stored legs are never
+ * read — that is what keeps the value tracking the token. A binding the library no longer backs (a
+ * token deleted after the post was saved) falls back to those legs, which still hold the value the
+ * token resolved to when it was picked. That differs on purpose from an unbacked PER-LEG alias, which
+ * `shadowAxisPx()` cannot fall back for because the alias replaced the number outright.
+ *
+ * @param {?Object} shadowItem      One `shadow[0]`-shaped item.
+ * @param {number}  blurFallback    What `blur` defaults to when unset — 14 on every current caller,
+ *                                  taken as an argument rather than hard-coded so the historic
+ *                                  per-state default stays with the call site that owns it.
+ * @param {number}  opacityFallback What `opacity` defaults to when unset. Blocks disagree on this
+ *                                  one — `kadence/singlebtn` has always used 1 and `kadence/image`
+ *                                  0.2 — so it too stays with the call site.
+ *
+ * @since TBD
+ *
+ * @return {string} The `box-shadow` value, or '' when there is no item to render.
+ */
+export function shadowCss(shadowItem, blurFallback, opacityFallback = 1) {
+	if (!shadowItem) {
+		return '';
+	}
+
+	const bound = boundShadowToken(shadowItem);
+
+	// A bound item renders from its token or not at all. When the token is no longer in the active
+	// library the stored legs are deliberately NOT used as a fallback: the value falls back to the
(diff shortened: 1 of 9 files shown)


pauloiankoski:
@coderabbitai review

tests/wpunit/Blocks/ImageTest.php
```diff
@@ -190,6 +190,66 @@ public function testBorderRadiusTokenAliasRendersAsAVariable(): void {
...
+	 * in and must render the shadow it has always rendered — the regression a lowered value default
+	 * causes, which nothing downstream repairs.
+	 *
+	 * @return void
+	 */
+	public function testLegacyImageWithNoStoredShadowValueStillRendersItsShippedShadow(): void {
+		$output = $this->render_image(
+			[
+				'displayBoxShadow' => true,
+				'boxShadow'        => [ $this->shipped_shadow_default() ],
```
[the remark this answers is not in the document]
pauloiankoski: Fixed in e88e8f1a8, and the same defect was in `SinglebtnTest` — fixed there too (4e28520f4). Both now read the default from `block.json` instead of supplying a literal.

Verified by lowering each default to transparent and confirming the test fails, then restoring.

src/extension/design-tokens/shadow-css.js
```diff
@@ -0,0 +1,99 @@
...
+ *                                  per-state default stays with the call site that owns it.
+ * @param {number}  opacityFallback What `opacity` defaults to when unset. Blocks disagree on this
+ *                                  one — `kadence/singlebtn` has always used 1 and `kadence/image`
+ *                                  0.2 — so it too stays with the call site.
+ *
+ * @since TBD
+ *
+ * @return {string} The `box-shadow` value, or '' when there is no item to render.
+ */
+export function shadowCss(shadowItem, blurFallback, opacityFallback = 1) {
```
d4mation, line 71: This docblock is the pre-#1266-policy text, and it now describes the exact behavior the three slices below spent their last commits removing. It says an unbacked binding "falls back to those legs, which still hold the value the token resolved to when it was picked" -- but the body returns `''`, and the inline comment eight lines down says the opposite in as many words: "the stored legs are deliberately NOT used as a fallback."

Slice 2 updated this docblock in `5dacaa749` ("renders nothing: the block falls back to its default CSS the same way every other block does when a token disappears"). The lift into this module picked up the version from before that commit, so the correction was silently rolled back. The `@return` at position 68 lost the same half -- it reads "or '' when there is no item to render", dropping "or the item's binding is no longer backed by the active library".

Behavior is right and the tests cover it, so this is documentation only -- but it is the single most load-bearing comment in the stack. This is now the shared module both blocks build their canvas shadow through, and this paragraph is the only place the deleted-token policy is written down for a reader who lands here first. Someone reading it would reimplement the bug. Worth restoring the slice 2 wording verbatim, including the trailing sentence about the legs still earning their place for the Custom tab.

pauloiankoski: Restored in fd9f33f01 — slice 2's wording verbatim, plus the `@return` half about an unbacked binding.

Same cause as a bug you'd have seen the fallout from otherwise: the lift picked up the pre-fix body, so the code silently kept the old behavior too. I caught that half when the cascade conflicted and fixed the code, but not the docblock above it.

Grepped the three shadow builders for any other fall-back-to-legs wording the lift carried over — none left.

d4mation (approved):

PR #1490, opened 2026-08-31
Ship one Default preset for the Button, matching every other block

d4mation:
The Button was the only block whose baseline presets were named **Primary** and **Secondary**.
Secondary is dropped and Primary becomes **Default**, so all six blocks now ship a single preset
slugged `default`.

The four `semantic.color.button-secondary-*` tokens existed only to feed the Secondary map, so they
go with it rather than shipping as dead entries in the token picker. The four that remain lose their
now-meaningless `primary` segment — `semantic.color.button-primary-bg` becomes
`semantic.color.button-bg`, and likewise for text, bg-hover and text-hover — matching the sibling
naming already used by `column-bg`, `heading-bg` and `image-bg`.

One behavior follows from the rename. `default` is reserved against the *creation* of a preset the
per-preset item route could never address, not against writing a slug the block already ships. The
Button was the sole block shipping no `default`, so it was the only place that write was refused; it
now merges onto the shipped preset like the other five. `Presets_Controller::guard_reserved_slugs()`
is unchanged apart from its docblock; the two tests that depended on the old asymmetry are reworked,
and the `order` slug still covers the refusal branch.

Three test groups needed real rework rather than a slug rename, because the Button was the only block
shipping two presets: the preset-ordering tests in `PresetsControllerTest` and the scoped-rule tests
in `Projection/Preset/Css_BuilderTest` now create a user `accent` preset and order or scope against
it, so the coverage survives.

**Gates:** `slic run wpunit Resources/Design_Tokens` 1666 tests green, `Blocks/SinglebtnTest` green,
Jest 1748 green, PHPStan no errors, `partial-phpcs.sh diff` clean, cspell clean.

### Checklist

### DIFF
--- a/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
+++ b/includes/resources/Design_Tokens/Registry/Baseline/baseline.json
@@ -386,35 +386,19 @@
 				"$type": "color",
 				"$value": "{primitive.color.brand.primary}"
 			},
-			"button-primary-bg": {
+			"button-bg": {
 				"$type": "color",
 				"$value": "{primitive.color.brand.button}"
 			},
-			"button-primary-text": {
+			"button-text": {
 				"$type": "color",
 				"$value": "{primitive.color.neutral.0}"
 			},
-			"button-primary-bg-hover": {
+			"button-bg-hover": {
 				"$type": "color",
 				"$value": "{primitive.color.brand.button-hover}"
 			},
-			"button-primary-text-hover": {
-				"$type": "color",
-				"$value": "{primitive.color.neutral.0}"
-			},
-			"button-secondary-bg": {
-				"$type": "color",
-				"$value": "{primitive.color.neutral.900}"
-			},
-			"button-secondary-text": {
-				"$type": "color",
-				"$value": "{primitive.color.neutral.0}"
-			},
-			"button-secondary-bg-hover": {
-				"$type": "color",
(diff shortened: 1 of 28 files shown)


includes/resources/Design_Tokens/Registry/Baseline/baseline.json
d4mation, line 1: `baseline.json` and `declarations.php` are the main files to look at here. Everything else is mostly test updates due to the names changing.

tests/wpunit/Resources/Design_Tokens/Rest/V1/PresetsControllerTest.php
```diff
@@ -1329,29 +1336,6 @@ public function testADanglingPresetAliasIsRejected(): void {
...
-					'preset' => 'default',
-					'tokens' => $this->button_tokens(),
-				]
-			)
-		);
-
-		$this->assertInstanceOf( WP_Error::class, $result );
-		$this->assertSame( 'rest_design_tokens_reserved_slug', $result->get_error_code() );
-		$this->assertSame( WP_Http::UNPROCESSABLE_ENTITY, $result->get_error_data()['status'] );
-	}
```
d4mation, line 1353: Expected. Writing to it would be considered a merge now since every Block has an identical slug for `default`.

pauloiankoski (approved):

PR #1491, opened 2026-08-31
[DTM] SOFT-4261 [1/4]: Extract the shared color popover

pauloiankoski:
## Summary

`ColorControl` held its popover body inline: a `TokenPopover` composed with `ColorGroupList` for the grouped Style Library tab, the relocated `ColorPicker` for the Custom tab, and a `Clear` row. Nothing else in the library could open that popover.

This slice lifts that body into a `ColorPopover` molecule, and the state a trigger and its popover both derive from `groups` + `value` into a pure `colorSelection()` helper. `ColorControl` keeps its own full trigger row and now composes `Dropdown` with `ColorPopover`.

The helper is a plain function rather than a hook because this repo ships no `@testing-library/react` / `react-test-renderer`, so a component's real logic has to live somewhere a test can call directly — the same split `colorSwatchStyle` already uses.

## Legacy CSS-variable colors

`colorSelection()` also routes a `var(--…)` value away from the Custom tab. The editor's older per-row color widget commonly stored a Kadence global-palette color that way. It is truthy and not a bracket alias, so it would otherwise open on Custom, where `react-color` cannot parse it, renders black, and the first interaction overwrites a working palette color with a hex. Such a value now opens on Style Library and is never handed to the picker as a seed; the swatch still paints it, since CSS resolves the variable in a `background`.

The pattern deliberately tolerates whitespace (`var( --x )`, `var(--x , #fff)`) — missing a spelling would route a real color down the destructive path this check exists to prevent.

## Behavior

No behavior change to `ColorControl`. `ColorControl.test.js` is untouched and still passes, which is the regression proof for the extraction.

## Test plan

- `npx wp-scripts test-unit-js src/token-controls` — 24 suites, 363 tests green
- `ColorControl.test.js` passes unmodified
- `color-selection.test.js` covers alias matching, the `Default` fallback for an out-of-group alias, literal passthrough, initial-tab selection, and the CSS-variable cases

Part of a 4-PR chain. Next: the compact swatch control that opens this same popover.

## Summary by CodeRabbit

* **New Features**
  * Added a reusable color-selection popover with Style Library tokens and a Custom color picker.
  * Added support for selecting, clearing, and displaying color tokens, aliases, raw values, and CSS-variable references.
  * Improved initial tab selection based on the current color value.

* **Bug Fixes**
  * Improved handling of unresolved aliases, whitespace, and fallback values in CSS variables.

* **Tests**
  * Added coverage for color swatches, aliases, labels, tabs, raw values, and CSS-variable formats.

### DIFF
--- a/src/token-controls/helpers/token-summary.js
+++ b/src/token-controls/helpers/token-summary.js
@@ -35,6 +35,40 @@ export function isTokenAlias(value) {
 }
 
 /**
+ * A CSS custom-property reference, e.g. `var(--global-palette1)`.
+ *
+ * The old editor's per-row color widget commonly stored a Kadence global-palette color this way.
+ * It is a real, working color — the browser resolves it at paint time — but neither a bracket alias
+ * nor a hex/rgba literal `react-color` can parse, so it needs its own shape check wherever a value
+ * is routed by kind.
+ *
+ * Deliberately permissive, because the two ways of being wrong are not equally costly: failing to
+ * recognize a real reference routes a working color to the Custom tab, where it is shown as black
+ * and overwritten on the first interaction, while over-matching only opens the Style Library tab on
+ * something that was not a variable after all. So this accepts every spelling CSS itself allows —
+ * `var` is ASCII case-insensitive like every CSS function name, whitespace may surround the
+ * custom-property name, the name may contain non-ASCII or backslash-escaped characters, and a
+ * fallback may follow a comma and span lines.
+ *
+ * @since TBD
+ */
+const CSS_VAR_PATTERN = /^var\(\s*--(?:\\.|[^\s,()\\])+\s*(?:,[\s\S]*)?\)$/i;
+
+/**
+ * Whether a slot value is a CSS custom-property reference rather than a token alias or a plain
+ * literal.
+ *
+ * @param {*} value The slot value.
+ *
+ * @since TBD
+ *
+ * @return {boolean} True when the value is a `var(--...)` reference.
+ */
+export function isCssVariableReference(value) {
+	return typeof value === 'string' && CSS_VAR_PATTERN.test(value.trim());
+}
+
+/**
  * Whether a slot holds anything at all.
  *
  * @param {*} value The slot value.
--- a/src/token-controls/molecules/ColorPopover.js
+++ b/src/token-controls/molecules/ColorPopover.js
@@ -39,13 +39,28 @@ import { isCssVariableReference } from '../helpers/token-summary';
  * @param {?Function} [props.resolveLiteral] `(entry) => string` — the host's hook for seeding the
  *                                            Custom tab from a currently-bound token entry.
  * @param {Function}  props.onClose          Closes the popover after a choice.
+ * @param {?Object}   [props.selection]      The already-derived `colorSelection(groups, value)`.
+ *                                            A control's trigger derives it anyway to render its own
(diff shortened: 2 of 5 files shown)


pauloiankoski:
@coderabbitai review

src/token-controls/helpers/token-summary.js
```diff
@@ -34,6 +34,37 @@ export function isTokenAlias(value) {
...
+ * is routed by kind.
+ *
+ * CSS permits whitespace around the custom-property name and before the fallback's comma, and a
+ * fallback may itself span lines. Missing one of those spellings would route a real color down the
+ * literal path, which is the destructive case this predicate exists to prevent — so the pattern is
+ * deliberately permissive about whitespace rather than matching only the canonical form.
+ *
+ * @since TBD
+ */
+const CSS_VAR_PATTERN = /^var\(\s*--[\w-]+\s*(?:,[\s\S]*)?\)$/;
```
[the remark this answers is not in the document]
pauloiankoski: Addressed. `CSS_VAR_PATTERN` now accepts every spelling CSS allows — case-insensitive `var`, whitespace around the custom-property name, non-ASCII and backslash-escaped identifiers, and a fallback after the comma. Regression cases added for `VAR(--global-palette1)`, `var(--café)` and `var(--escaped\ name)`.

Kept as a permissive pattern rather than a full CSS-value parser, because the two failure directions are not equally costly: missing a real reference sends a working color to the Custom tab, where it renders black and is overwritten on the first interaction, whereas over-matching only opens the Style Library tab on something that turned out not to be a variable.

src/token-controls/molecules/ColorPopover.js
```diff
@@ -0,0 +1,93 @@
...
+ * @param {?Function} [props.resolveLiteral] `(entry) => string` — the host's hook for seeding the
+ *                                            Custom tab from a currently-bound token entry.
+ * @param {Function}  props.onClose          Closes the popover after a choice.
+ *
+ * @since TBD
+ *
+ * @return {JSX.Element} The rendered popover body.
+ */
+export function ColorPopover({ value, groups, onClear = null, onPick, onCustom, resolveLiteral, onClose }) {
+	const { allSwatches, entry, initialTab } = colorSelection(groups, value);
```
d4mation: `colorSelection(groups, value)` is computed twice for the same control: `ColorControl.js:69` already calls it for `entry`/`selectedLabel`, and this line recomputes the identical `groups.flatMap(...)` + `findTokenEntry(...)` to build the popover body. `ColorSwatchControl` repeats the pair in slice 2. Neither result is memoized, so every render of an open control flattens the palette twice, and the helper's stated role as the one place trigger and popover derive from stops being literally true. Accepting the computed selection as a prop and falling back to `colorSelection(groups, value)` when it is omitted (`function ColorPopover({ selection, value, groups, ... }) { const { allSwatches, entry, initialTab } = selection ?? colorSelection(groups, value); }`) keeps both call sites honest and drops the second pass.

d4mation (approved):

src/token-controls/molecules/ColorPopover.js
pauloiankoski: Fixed in 7dec4cb14. `ColorPopover` now takes an optional `selection` prop and falls back to `colorSelection(groups, value)` when omitted; `ColorControl` derives it once and passes it down, so the palette is flattened a single time per render.

PR #1492, opened 2026-08-31
[DTM] SOFT-4261 [2/4]: Add the compact swatch color control

pauloiankoski:
## Summary

Adds `ColorSwatchControl`, the trigger that lets a border row open the popover the previous slice extracted.

`BorderControl`'s row is one composite box — swatch, style preview, width field — with no room for `ColorControl`'s full row (a static attribute label plus the selected token's label, and a `BindingIndicator` beside it). This control is the same popover behind a bare 20px swatch: no visible text at all, its name carried on `aria-label`, and no binding indicator, since a border color carries no preset binding of its own and has no Reset to render.

It reuses `kb-color-control__popover` as its `contentClassName` deliberately — it is the same popover, and a second class would let the two sizings drift.

Also adds `sideLabel()`, which maps a border side key (`top`, `right`, …) to a translated name. `BorderControl` is what names a row's side and hands it to `renderColor`, so the vocabulary belongs in the library; both hosts consume it in the next two slices. The four sides are spelled out rather than derived so translators get whole phrases to work with.

## Wired to nothing yet

Neither host renders this control until the following slices. This one adds it, exports it, styles it, and tests it in isolation.

## Test plan

- `npx wp-scripts test-unit-js src/token-controls` — 25 suites, 370 tests green
- `ColorSwatchControl.test.js` covers the bare trigger (asserting no visible text, not merely that a node exists), `aria-label`, the disabled state, a grouped pick returning the entry's alias, a Custom-tab write going to `onCustom` rather than `onPick`, the Custom tab seeding through `resolveLiteral`, and the Clear row's presence and disabled state

Part of a 4-PR chain, on top of the shared popover extraction.

## Summary by CodeRabbit

* **New Features**
  * Added an accessible color swatch control for selecting design tokens or custom colors.
  * Supports clearing selections, disabled states, literal color resolution, and keyboard focus styling.
  * Added localized labels for border sides.
* **Bug Fixes**
  * Added comprehensive automated coverage for color selection, custom colors, CSS variables, disabled behavior, and clearing.

### DIFF
--- a/src/token-controls/controls/ColorSwatchControl.js
+++ b/src/token-controls/controls/ColorSwatchControl.js
@@ -1,0 +1,92 @@
+/**
+ * The compact color token control: a bare round swatch that opens the same popover `ColorControl`
+ * opens, with no label, no value text, and no binding indicator beside it.
+ *
+ * `BorderControl`'s row is one composite box — swatch, style preview, width field — with no room for
+ * `ColorControl`'s full trigger row, and a border's color carries no preset binding of its own, so
+ * there is no Reset glyph to render. The field's name therefore rides on `aria-label` rather than on
+ * visible text, the way the Style Library's own swatch toggles already name themselves.
+ *
+ * The popover body itself is `ColorPopover`, shared with `ColorControl` — the two controls differ
+ * only in what opens the popover.
+ */
+
+/**
+ * WordPress dependencies
+ */
+import { Dropdown } from '@wordpress/components';
+
+/**
+ * Internal dependencies
+ */
+import { ColorPopover } from '../molecules/ColorPopover';
+import { ColorSwatch } from '../atoms/ColorSwatch';
+import { colorSelection } from '../helpers/color-selection';
+import { isTokenAlias } from '../helpers/token-summary';
+import '../styles/token-controls.scss';
+
+/**
+ * Render a compact color token control.
+ *
+ * @param {Object}    props                  The component props.
+ * @param {string}    props.label            The field's accessible name (e.g. "Top Border Color").
+ *                                            Never rendered as visible text.
+ * @param {*}         props.value            The current slot value: a bracket alias or a raw literal.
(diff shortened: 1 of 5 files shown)


pauloiankoski:
@coderabbitai review

src/token-controls/controls/ColorSwatchControl.js
```diff
@@ -0,0 +1,92 @@
...
+			className="kb-color-swatch-control"
+			// Deliberately the same class `ColorControl` gives its own popover: it IS the same popover,
+			// and a second class would let the two sizings drift apart.
+			contentClassName="kb-color-control__popover"
+			popoverProps={{ placement: 'left-start' }}
+			renderToggle={({ isOpen, onToggle }) => (
+				<button
+					type="button"
+					className="kb-color-swatch-control__button"
+					aria-label={label}
```
d4mation: The accessible name is the field name alone, so a screen reader announces "Top Border Color, button" with no indication of what is currently selected. `ColorControl` conveys that visually through `selectedLabel` (`ColorControl.js:88`); this control drops the visible text by design but does not replace the information anywhere, so the selection becomes unreadable rather than just unwritten. `colorSelection()` already returns `selectedLabel` -- destructuring it here and composing the name restores parity: `aria-label={selectedLabel ? sprintf(/* translators: 1: the field name, 2: the selected color's name. */ __('%1$s: %2$s', 'kadence-blocks'), label, selectedLabel) : label}`.

d4mation (approved):

src/token-controls/controls/ColorSwatchControl.js
pauloiankoski: Fixed in dc0502509. The trigger now composes `%1$s: %2$s` from the field name and `selectedLabel`, so a bound color is announced as "Top Border Color: Main". It also passes the derived selection to `ColorPopover`. Test added.

PR #1493, opened 2026-08-31
[DTM] SOFT-4261 [3/4]: Open the shared color popover from a border row in the block editor

pauloiankoski:
## Summary

Switches the block editor's border color from `PopColorControl` to the shared `ColorSwatchControl`, so a border's color opens the same grouped popover as every other color control in the product.

`border-color.js` exported a plain `renderBorderColor` function with no access to palette data. It now exports a `BorderColorField` component taking `groups`. Both blocks that use it already call `useColorGroups(clientId)`, so each defines the render prop once:

```js
const renderBorderColor = useCallback((row) => <BorderColorField {...row} groups={colorGroups} />, [colorGroups]);
```

That local shadows what the import used to provide, so all seven `renderColor={renderBorderColor}` sites are untouched — six in `singlebtn`, one in `advancedheading`.

The file's docblock previously explained that it was kept out of `EditorBorderControl.js` because `@kadence/components`' `dist/cjs` build ships bare `import` statements Jest cannot parse. Dropping `PopColorControl` removes that constraint, so the docblock is rewritten: the file stays where it is because it is the editor's color adapter, beside `shadow-color.js`.

## Stored values are unchanged

`register-color-control-filters.js` already rewrote token-backed palette swatches to their `{alias}`, so `PopColorControl` was writing bracket aliases into `borderStyle.color` — exactly what the new control speaks. Colors stored as raw literals keep working: they match no entry and the swatch paints them directly.

## One narrowing worth a reviewer's attention

The border color loses `PopColorControl`'s wider swatch set — custom colors, recents, and the native picker grid — in exchange for the palette's four groups plus a Custom tab. That is the point of the change, but it is a shorter list than what ships today.

A legacy color stored as `var(--global-palette1)` is handled by the guard added in the first PR of this chain: it opens on Style Library rather than the Custom tab, and the swatch still paints it.

## CSS

Deletes the rules that existed only to strip `PopColorControl`'s chrome inside `.kb-border-control__swatch`, including the standalone `(0,4,0)` specificity override of its 28×28 swatch. Box shadow still renders `PopColorControl`, and its rules are untouched.

screenshot-1788205187480-22

## Test plan

- `npx wp-scripts test-unit-js src/token-controls src/extension src/blocks` — 60 suites, 929 tests green
- Verified in the editor on `kadence/singlebtn` and `kadence/advancedheading`: the grouped popover opens, a pick writes the token alias to all four sides and paints on the canvas, unlinked mode gives four swatches with distinct accessible names, per-side writes stay isolated, the Custom tab seeds from the bound token, and Clear affects only its own side.

## Summary by CodeRabbit

* **Improvements**
  * Updated border color controls across heading and button settings with a more consistent color selection experience.
  * Added support for color palettes, custom colors, clearing selections, and disabled states.
  * Improved border color labels and handling for individual sides.

* **Style**
  * Refined color swatch styling for a cleaner, more consistent appearance.

### DIFF
--- a/src/extension/design-tokens/components/border-color.js
+++ b/src/extension/design-tokens/components/border-color.js
@@ -1,81 +1,65 @@
 /**
- * The default `renderColor` for `EditorBorderControl`.
+ * The block editor's color field for one `BorderControl` row.
  *
- * Kept out of `EditorBorderControl.js` for the reason `shadow-color.js` is kept out of
- * `EditorShadowControl.js`: it is the one piece that reaches for `@kadence/components`, whose
- * `dist/cjs` build ships bare `import` statements Jest cannot parse, so importing it into the
- * component would make that component untestable in isolation.
+ * Sits beside `color-literal.js` and `use-color-groups.js`, the two halves of a `ColorControl`'s
+ * data in this host, because it is the editor's own adapter rather than part of the host-agnostic
+ * control library: it knows how a border side is named in this editor and how a token resolves
+ * against this document.
+ *
+ * Border color is bundled with width and style inside one nested attribute, so it has no
+ * `ColorControl` row of its own; this renders the compact `ColorSwatchControl` into the slot
+ * `BorderControl` leaves for it, opening the same grouped popover every other color control opens.
  */
 
 /**
  * WordPress dependencies
  */
-import { PopColorControl } from '@kadence/components';
 import { __, sprintf } from '@wordpress/i18n';
 
 /**
- * The translated name of one border side.
- *
- * `BorderControl` names its rows in English (`top`, `right`, …) because those are the keys of the
- * value it edits, not display text. Capitalizing the key leaves it in English, so the side is spelled
- * out here per side rather than derived — the four are a closed set, and a translator needs the whole
- * phrase in front of them to render "Top Border Color" naturally in a language that orders it
- * differently.
- *
- * @param {string} side The side key: `top`, `right`, `bottom` or `left`.
- *
- * @since TBD
- *
- * @return {string} The translated side name, or the key itself when it is not one of the four.
+ * Internal dependencies
  */
-function sideLabel(side) {
-	const names = {
-		top: __('Top', 'kadence-blocks'),
-		right: __('Right', 'kadence-blocks'),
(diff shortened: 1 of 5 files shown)


pauloiankoski:
@coderabbitai review

src/extension/design-tokens/components/border-color.js
```diff
@@ -1,81 +1,65 @@
...
+ *                                      linked.
+ * @param {boolean}   [props.disabled] Whether the row is read-only.
+ * @param {Array}     props.groups     The block's effective palette groups, from `useColorGroups()`.
  *
  * @since TBD
  *
  * @return {JSX.Element} The rendered color field.
  */
-export function renderBorderColor({ value, onChange, label }) {
+export function BorderColorField({ value, onChange, label, disabled = false, groups }) {
```
d4mation, line 40: This is the only slice in the stack with no test. `BorderColorField` is now a plain component over `ColorSwatchControl` with no `@kadence/components` import -- the exact reason the module docblock used to give for why it could not be tested in isolation is gone, and `ColorSwatchControl.test.js` in slice 2 already establishes the mock set (`@wordpress/components`, `@wordpress/icons`, `../molecules/ColorPicker`, the virtual SCSS) that a sibling `__tests__/border-color.test.js` could reuse verbatim. Four assertions would cover the whole wiring and none of them exist right now: `onPick` writes the bracket alias straight through `onChange`; `onCustom` writes a raw literal through the same `onChange`; `onClear` writes `''` (a behavior change from `PopColorControl`'s `hideClear={true}`, so it is new surface, not preserved surface); and `label` produces "Top Border Color" per side but "Border Color" when `label` is `null`. The Style Library host gets equivalent coverage in `ColorSelectField.test.js`, so this is the asymmetry rather than a new standard.

d4mation (approved):

src/extension/design-tokens/components/border-color.js
pauloiankoski: Fixed in e97ae788e. Added `__tests__/border-color.test.js` reusing slice 2 mock set, covering all four: alias write, literal write, `onClear` writing `""`, and per-side vs linked naming. It also now uses the shared `borderColorLabel()`.

PR #1494, opened 2026-08-31
[DTM] SOFT-4261 [4/4]: Open the shared color popover from a border row in the Style Library

pauloiankoski:
## Summary

Switches the Style Library's border color from `TokenColorSelectField` — a flat, ungrouped menu with no Custom tab — to the shared `ColorSwatchControl`, completing the chain. Both hosts now open the same popover from a border row.

## The value shape differs between hosts

The block editor stores a bracket alias (`{semantic.color.accent.main}`). This host stores a **bare token id** (`semantic.color.accent.main`) — what `TokenColorSelectField` wrote, and what the preset's `${field.path}-color` key holds. `ColorSwatchControl` only speaks bracket aliases, so the value is bridged in both directions, and a Custom-tab literal passes through untouched (running a raw hex through `toStoredValue` would slice off its first and last characters).

Those three functions previously lived in `ColorSelectField.js` and were imported by `BorderField` from its sibling. They move to `src/style-library/helpers/color-values.js`, matching what this chain already did for the palette read, so neither field depends on the other's file.

## Shared palette read

`useActivePaletteGroups()` extracts the active-palette lookup out of `ColorSelectField`. `BorderField` needs it in its own component body, since a hook cannot be called from inside the `renderColor` render prop. The comment explaining why `STORE_NAME` comes from `store/constants` rather than `store` moves with it — that reasoning is the whole point of the odd-looking import.

## Translated side names

`BorderField` used `upperFirst(side)` on the raw English key, so its four unlinked swatches read "Top/Right/Bottom/Left" in every locale. Both hosts now use the shared `sideLabel()` and the same Title Case wording.

## Scope

`TokenColorSelectField` stays — `BoxShadowField` still renders it. Box shadow keeps its existing widgets in both hosts, deliberately: its color slot packs opacity into the same value, which needs its own decision.

`BorderControl.js`'s docblock is corrected here (prose only, no code): it said color "is being reworked separately", which contradicted the same docblock's later paragraph now that the rework has landed. The `.kb-border-control__swatch` comment is updated for the same reason, and the last chrome-stripping rules are deleted.

screenshot-1788203677209-21

## Test plan

- `npx wp-scripts test-unit-js src` — 125 suites, 1856 tests green
- `border-shadow-field-rendering.test.js`'s case pinning `TokenColorSelectField` now asserts `ColorSwatchControl`; `ColorSelectField.test.js` passes unmodified, since the extraction moved code rather than behavior
- Verified on the Button preset screen: the same popover opens, and a pick survives Save and a full reload — the popover reopens on the Style Library tab with the chosen swatch checked, confirming the bare-token-id round trip. Box shadow's color row still renders `TokenColorSelectField`.

## Summary by CodeRabbit

* **Improvements**
  * Updated border color controls to use a consistent color swatch interface.
  * Improved color selection across style fields, including palette tokens, custom values, clearing, and read-only states.
  * Ensured selected colors display correctly whether they come from a palette or a literal value.
  * Added clearer, side-specific labels to improve accessibility when editing border colors.
* **Documentation**
  * Updated BorderControl guidance to reflect the current color swatch behavior.

### DIFF
--- a/src/style-library/components/molecules/fields/BorderField.js
+++ b/src/style-library/components/molecules/fields/BorderField.js
@@ -22,10 +22,13 @@
  * `BoxControl` — so this adapter fixes the unit at `px` (the only unit the `border-width` scale's
  * tokens use) rather than tracking one the way `BoxTokenField` tracks radius/spacing's unit.
  *
- * Color is out of this plan's scope (see `BorderControl`'s own docblock): `renderColor` wraps the
- * same `TokenColorSelectField` the Button screen's Color panel already renders for text/background,
- * rather than building or importing anything new. Color's own path carries no breakpoint envelope —
- * a border color has never varied by breakpoint here, so its path always stores the plain value.
+ * Color's own sub-field is the shared `ColorSwatchControl` — `renderColor` wraps it here the same
+ * way the block editor's own `BorderControl` host does, so a border's color opens the same grouped
+ * Style Library / Custom popover `ColorSelectField` opens, bridged through the
+ * `toControlValue`/`toStoredValue`/`resolveLiteral` pair `helpers/color-values.js` shares with that
+ * field, since this host stores a bare token id, not a bracket alias. Color's own path carries no
+ * breakpoint envelope — a border color has never varied by breakpoint here, so its path always
+ * stores the plain value.
  *
  * Link state is owned here, controlled, exactly the way `BoxTokenField` owns it for radius/spacing
  * — not derived from whether the stored value happens to be a scalar or a four-slot list.
@@ -39,15 +42,10 @@
  */
 
 /**
- * External dependencies
- */
-import { upperFirst } from 'lodash';
-
-/**
  * WordPress dependencies
  */
 import { useState } from '@wordpress/element';
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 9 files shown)


pauloiankoski:
@coderabbitai review

src/style-library/components/molecules/fields/BorderField.js
```diff
@@ -331,19 +333,29 @@ export function BorderField({ field, values, originalValues, onValueChange }) {
...
+					// visible text at all — do not read as four copies of the same field.
+					label={
+						side
+							? sprintf(
+									/* translators: %s: the border side, already translated — Top, Right, Bottom or Left. */
+									__('%s Border Color', 'kadence-blocks'),
+									sideLabel(side)
+								)
+							: __('Border Color', 'kadence-blocks')
+					}
```
d4mation: Lines 341-349 are a verbatim copy of `src/extension/design-tokens/components/border-color.js:47-56`, translator comment included. Slice 2 lifted `sideLabel()` into `token-controls/helpers/side-labels.js` specifically so the two hosts could not drift on how a border side is named, and its docblock argues the case in terms of the whole phrase a translator sees -- but the phrase itself is the half that stayed behind, in two copies that now have to be kept in sync by hand. Exporting the composed label beside `sideLabel` (`borderColorLabel(side)`, returning the `%s Border Color` sprintf for a side and `Border Color` for `null`) finishes the extraction slice 2 started and leaves one string for translators instead of two identical ones.

src/style-library/components/molecules/fields/ColorSelectField.js
d4mation, line 1: Stale file docblock as of this PR: `TokenColorSelectField.js` is no longer "`BorderField`/`BoxShadowField`'s own color sub-field" -- it is `BoxShadowField`'s only. Two more say the same thing elsewhere and need the same pass: `src/style-library/components/molecules/fields/BoxShadowField.js:18-19` ("Color is out of this plan's scope, exactly as in `BorderField`: `renderColor` wraps the same `TokenColorSelectField`"), where the "exactly as in `BorderField`" comparison is now backwards, and `src/style-library/__tests__/ColorSelectField.test.js:227` ("`field.readOnly` disables the control the same way it disables `TokenColorSelectField`"), which points at a field this one no longer sits beside.

src/style-library/components/molecules/fields/BorderField.js
pauloiankoski: Fixed in 2bffd1676. Added `borderColorLabel(side)` beside `sideLabel` in `side-labels.js`, returning the composed phrase for a side and the bare name for `null`. Both hosts call it, so translators see one string.

src/style-library/components/molecules/fields/ColorSelectField.js
pauloiankoski: Fixed in 2bffd1676. All three updated: `ColorSelectField.js` now says `BoxShadowField`s alone, `BoxShadowField.js` states it is the only remaining renderer instead of comparing to `BorderField`, and the test comment no longer cites that neighbour.

d4mation (approved):

PR #1496, opened 2026-08-31
[DTM] Give the editor's color controls the same bottom spacing as their neighbors

pauloiankoski:
## Summary

In a block's inspector, the color controls have no bottom gap to whatever follows them — "Background Color" sits flush against the "Border" label beneath it, while Border, Border Radius and Box Shadow each keep a clear margin.

Every other editor token control gets that gap from wrapping itself in `TokenControlRow`, whose `.kb-token-control-row:not(:last-child) { margin-bottom: 16px; }` supplies it:

- `EditorBoxControl.js` — `<TokenControlRow stacked>`
- `EditorBorderControl.js` — `<TokenControlRow stacked>`
- `EditorShadowControl.js` — `<TokenControlRow stacked>`

`ColorControl` is never wrapped in one, so it has no margin at all. It cannot wrap itself: it lives in the host-agnostic `src/token-controls/`, and importing from `src/extension/` would invert the library/host dependency direction the codebase deliberately maintains.

## Fix

Adds the missing sibling adapter, `EditorColorControl`, which wraps `ColorControl` in `TokenControlRow` and forwards every prop untouched — the pattern `TokenControlRow`'s own docblock describes:

> `attr`/`binding` can be omitted entirely … for a control with no bound attribute to track at all — used this way purely for the `.kb-token-control-row` spacing/`stacked` layout, e.g. by the `Editor*` design-token controls that already carry their own indicator internally.

That fits `ColorControl`, which renders its own `BindingIndicator` inline. With no `heading`, `TokenControlRow`'s header block renders nothing, so no second indicator appears.

`EditorColorControlGroup` does the same for `ColorControlGroup`. The group stacks its children flush with shared borders, so the wrapper goes around the **group**, never its children — a separate component rather than an inline wrap at each call site, so a future call site cannot wrap the children by mistake.

Call sites: `singlebtn` has 12 standalone controls plus one group; `advancedheading` has one group holding both of its controls.

## No CSS

An earlier attempt did this with two rules in `token-controls.scss`, including one scoped to a Style Library class to cancel the margin where `SettingsForm`'s flex `gap` already spaces fields. That put host-specific knowledge in the shared stylesheet. This approach needs none of it — `TokenControlRow` lives in the editor extension, so the Style Library never sees it. `token-controls.scss` has zero diff.

## Follow-up, not included

Every call site already computes a `tokenBinding.<key>` object, which is the shape `TokenControlRow`'s `attr`/`binding` props expect, so these controls could also get the "highlight edits" tint the other `Editor*` controls have. It needs a `heading` too, so it is its own change.

## Test plan

- `npx wp-scripts test-unit-js src/token-controls src/extension src/blocks` — 58 suites, 915 tests green
- Verified no double indicator: `TokenControlRow`'s header is `{heading ? (...) : null}`, and no `heading` is passed
- `stacked` matches the three sibling adapters; without it the base `.kb-token-control-row` is a centered row-direction flex, which does not let the full-width color trigger fill correctly

## Summary by CodeRabbit

- **Enhancements**
  - Updated heading and button color settings with a more consistent editor experience.
  - Improved spacing and presentation for color controls without adding redundant labels or indicators.
  - Preserved existing color options and behavior across normal, hover, transparent, and sticky button styles.
  - Maintained separate icon color controls for precise customization.

### DIFF
--- a/src/blocks/advancedheading/edit.js
+++ b/src/blocks/advancedheading/edit.js
@@ -98,7 +98,8 @@ import { resolveColorLiteral } from '../../extension/design-tokens/color-literal
 import { renderBorderColor } from '../../extension/design-tokens/components/border-color';
 import { tokenDimension } from '../../extension/design-tokens/token-dimension';
 import { pickableTokensForControl, pickableTokensForKey } from '../../extension/token-picker';
-import { ColorControl, ColorControlGroup } from '../../token-controls';
+import { ColorControl } from '../../token-controls';
+import { EditorColorControlGroup } from '../../extension/design-tokens/components/EditorColorControlGroup';
 import { presetFontVariant } from './preset-font-variant';
 import metadata from './block.json';
 /**
@@ -1871,7 +1872,7 @@ function KadenceAdvancedHeading(props) {
 								{showSettings('colorSettings', 'kadence/advancedheading') && (
 									<>
 										{!enableTextGradient && (
-											<ColorControlGroup>
+											<EditorColorControlGroup>
 												{/* Every write clears the matching `*Class` attribute. Those carry
 												    legacy global-palette CLASSES the heading emits on its own
 												    element, so they beat the color these controls write and a stale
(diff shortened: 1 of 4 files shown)


pauloiankoski:
@coderabbitai review

d4mation (approved):

PR #1497, opened 2026-08-31
[DTM] SOFT-4270 [1/2]: Reset the Default palette when it is also the active one

pauloiankoski:
## Summary

The Default palette could not be reset. Clicking **Reset** on the Default palette while it was also the active one issued no request at all: the modal stayed open showing "Choose which palette your site should use instead." — a demand for a successor palette, with no control rendered to pick one in. The only way out was Cancel.

The modal and the flow behind it disagreed about when a successor is needed. `DeletePaletteModal` had it right — a reset leaves the palette in place and still active, so it renders no successor dropdown and confirms with an empty successor:

```js
const needsSuccessor = isActive && isUserCreated;
```

`deletePaletteFlow` compared ids only, with no user-created dimension, so for the Default palette while it was active it rejected before `deletePalette()` was ever called. `usePalettes`'s `removePalette` passed `currentId` but never the user-created flag, so the flow had no way to tell a reset from a delete.

The server was already correct: `Palettes_Controller::delete_item()` resets a palette the baseline defines and only reassigns `$current` when the palette actually goes away. The request the UI refused to send would have succeeded. **No PHP changes here** — this is the client half.

## What changed

* `deletePaletteFlow` takes `isUserCreated` and gates on `id === currentId && isUserCreated`, the same line the modal draws and the same one `deleteLibraryFlow` already draws with `!isDefaultLibrary(slug)`.
* `removePalette` derives that flag with `isUserCreatedPalette(listing, id)` and announces **"Palette reset."** rather than "Palette deleted." when the palette is one the baseline defines.
* Four comments citing a `delete_item()` default-id guard that no longer exists are corrected — `deletePalette`'s doc in `api/client.js`, two in `ColorPaletteScreen`, and one in `renamePaletteFlow`.

## Tests

* `palette-flows.test.js` — a new case for resetting the live baseline palette (the shipped state, and the one that was broken); the stale `surfaces the default-palette 400 message` case replaced with a plain request-failure case, since the server no longer returns that 400; `isUserCreated` added to the three existing delete cases so they keep exercising the successor path.
* `use-palettes.test.js` — a reset case asserting the DELETE goes out with no successor activation and the notice reads "Palette reset.". The shared fixture's Sunset row is now `user_created: true`, which is what it always meant; without that its sibling's "Palette deleted." assertion would silently flip.
* `palette-reset-modal.test.js` (new) — at the screen level: confirming Reset on the live Default palette reaches the hook with no successor and renders no successor control, while deleting a live user-created palette still renders the dropdown and keeps the confirm button disabled until it is answered.

Restoring the old `id === currentId` gate fails four of these, so the coverage does catch the bug.

## Test plan

* [x] `npx wp-scripts test-unit-js --testPathPattern "src/style-library"` — 66 suites, 928 tests pass.
* [x] eslint and cspell clean on every changed file.
* [x] Manual, on the shipped state (Default palette open and active): clicking Reset now issues `DELETE /kb-design-tokens/v1/palettes/default` (200), the stored override is dropped, the notice reads "Palette reset.", and after a reload the palette is still in the listing and still marked Active.

## Summary by CodeRabbit

* **New Features**
  * Resetting a shipped palette now removes its customizations while keeping the palette available.
  * Default palettes can be reset without selecting a replacement.
  * User-created palette deletion continues to require a replacement when deleting the active palette.

* **Bug Fixes**
  * Improved palette reset and deletion messaging.
  * Updated reset and deletion behavior for more consistent palette listings.

### DIFF
--- a/src/style-library/hooks/use-palettes.js
+++ b/src/style-library/hooks/use-palettes.js
@@ -13,6 +13,7 @@ import {
 	applyOptimisticOverlay,
 	customColorTokenId,
 	isCustomColorToken,
+	isUserCreatedPalette,
 	newSwatchValue,
 	nextCustomColorSlug,
 	reorderGroupSwatches,
@@ -379,23 +380,30 @@ export function usePalettes(feed, refreshFeed, route, navigate) {
 				return busy;
 			}
 
+			// Decides both halves of what this call means: only a user-created palette is removed and
+			// can need a successor, while a baseline one is reset in place and stays in the listing.
+			const isUserCreated = isUserCreatedPalette(listing, id);
+
 			setDeleteError(null);
 			return deletePaletteFlow({
 				namespace,
 				slug,
 				id,
 				currentId: listing.currentId,
+				isUserCreated,
 				successorId,
 				onReceive,
 				refreshFeed,
 				onBusy: setIsBusy,
 				onError: setDeleteError,
 			}).then((result) => {
-				notifySuccess(__('Palette deleted.', 'kadence-blocks'));
+				notifySuccess(
+					isUserCreated ? __('Palette deleted.', 'kadence-blocks') : __('Palette reset.', 'kadence-blocks')
+				);
 				return result;
 			});
 		},
-		[namespace, slug, listing.currentId, onReceive, refreshFeed]
+		[namespace, slug, listing, onReceive, refreshFeed]
 	);
 
 	const saveSwatchEdits = useCallback(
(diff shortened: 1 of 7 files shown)


pauloiankoski:
@coderabbitai review

d4mation (dismissed):

src/style-library/hooks/use-palettes.js
d4mation, line 1: When deleting a non-Default Palette, you see the "Reset" modal briefly after the non-Default Palette Deletion Modal disappears.

This seems to happen because we switch to the Default Palette _just before_ dismissing the modal, so the state there changes to show the incorrect modal.

image

pauloiankoski: Fixed in 644809b87: the modal now snapshots the palette it was opened for when it opens, so the mid-delete switch back to the Default palette no longer flips its copy. Regression test added.

d4mation (approved):

Issue #1175, opened 2026-07-27
Block CSS extension HOC calls hooks conditionally, triggering React "Expected static flag was missing" warning on block selection


mauteri:
## Summary

`BlockCSSComponent` in `src/extension/block-css/index.js` violates the Rules of Hooks: it returns early for deselected blocks **before** calling `useSelect` and `useState`, so the same component instance renders with zero hooks while deselected and two hooks once selected. React's development build flags this the moment any block is selected:

```
Warning: Internal React error: Expected static flag was missing. Please notify the React team.
```

## Where

`src/extension/block-css/index.js` (current default branch):

```js
const BlockCSSComponent = createHigherOrderComponent((BlockEdit) => {
	return (props) => {
		if (!props.isSelected) {
			return <BlockEdit {...props} />;      // line 46: early return...
		}
		const { hasBlockCSS, openTab } = useSelect((select) => {   // line 49: ...then hooks
			...
		const [isOpen, setOpen] = useState(false);                 // line 67
```

Because the HOC is attached via `editor.BlockEdit`, it wraps every block, so the warning fires for any block type.

## Steps to reproduce

1. WordPress with `SCRIPT_DEBUG` enabled (so the unminified React development build loads) and Kadence Blocks active.
2. Open the post or site editor with any content.
3. Click any block to select it.
4. Console shows the "Expected static flag was missing" warning, with `extension-block-css.js` at the top of the component stack.

Production builds of React tolerate the mismatch silently, which is why the warning only appears on debug environments. Observed on Kadence Blocks 3.7.8, WordPress 6.8.x, React 18.3.1.

## Suggested fix

Keep the `isSelected` performance gate but move it to the component level instead of the hook level, since conditional *components* are fine while conditional *hooks* are not:

```js
const BlockCSSComponent = createHigherOrderComponent((BlockEdit) => {
	return (props) => {
		if (!props.isSelected) {
			return <BlockEdit {...props} />;
		}
		return (
			<>
				<BlockEdit {...props} />
				<BlockCSSControls {...props} />
			</>
		);
	};
}, 'BlockCSSComponent');

// All hooks (useSelect, useState, the panel UI) live in BlockCSSControls,
// which mounts only while the block is selected.
```

The deselected path stays as cheap as today, and hook counts are consistent for every render of each component instance.

PR #1501, opened 2026-09-02
fix: Call hooks unconditionally in BlockCSSComponent

Saggre:
`BlockCSSComponent` returned early with `<BlockEdit {...props} />` before calling `useSelect` and `useState` whenever `props.isSelected` was `false`. Since the number of hooks a component calls must stay identical across renders, selecting or deselecting the block toggles this early return and changes the hook count between renders, which violates the Rules of Hooks.

React 18 catches this and logs:

```
Warning: Internal React error: Expected static flag was missing...
```

repeatedly in the block editor console.

The fix moves the `isSelected` early-return check to after the hook calls, so `useSelect` and `useState` always run regardless of selection state. Behavior is unchanged, only the order shifted.

## Testing

1. Open a page containing a Kadence block in the block editor (e.g. `wp-admin/post.php?post=<id>&action=edit`).
2. Open the browser console.
3. Click through several blocks to select/deselect them.
4. Before the fix: "Expected static flag was missing" warnings appear on selection changes. After the fix: no such warnings.

## Validation

Reproduced and verified locally against a live WordPress install (Kadence Blocks 3.7.9.1): baseline showed 38 occurrences of the warning after clicking through 46 blocks in the editor canvas; with the equivalent patch applied to the installed plugin bundle, the warning was completely eliminated across the same test.

### Checklist

Not applicable

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved block editing stability by ensuring selection-related behavior remains consistent when block selection changes.
  * Updated internal hook execution order to comply with React’s Rules of Hooks, with no intended user-facing behavior changes.

### DIFF
--- a/src/extension/block-css/index.js
+++ b/src/extension/block-css/index.js
@@ -40,12 +40,11 @@ addFilter('blocks.registerBlockType', 'kadence/blockCSS', blockCSSAttribute);
 
 /**
  * Build the block css control
+ *
+ * Hooks run before the isSelected check to satisfy the Rules of Hooks.
  */
 const BlockCSSComponent = createHigherOrderComponent((BlockEdit) => {
 	return (props) => {
-		if (!props.isSelected) {
-			return <BlockEdit {...props} />;
-		}
 		const { hasBlockCSS, openTab } = useSelect((select) => {
 			const hasSupport = select(blocksStore).hasBlockSupport(props.name, 'kbcss');
 			let openTab = 'general';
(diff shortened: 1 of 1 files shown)


Saggre:
@d4mation @pauloiankoski IDK who to assign this to, but this closes #1175

d4mation:
@Saggre Thanks for the contribution! I've made sure that a ticket has been created internally for this and we have it targeted for an upcoming maintenance release. We'll follow up once a developer has been assigned to it. :)

PR #1505, opened 2026-09-04
[DTM] SOFT-4268 [3/3]: Keep a preset's row chip filled while its panel is open

pauloiankoski:
🎥 [Walkthrough video](https://www.loom.com/share/5d2e88ee3bee41e4a845dcccf314cf9a)

Third of three. Based on #1502.

## What

Opening a preset's settings panel blanked its row chip. The draft seeds every non-overridden
property empty (see `presetInitialValues`), and the overlay previewed that draft alone — so the
moment the panel opened, the chip showed a button the preset never renders.

The overlay now merges the draft's *set* properties over the row's stored, baseline-merged token
map. A property the draft leaves unset keeps previewing the stored value, which is also what a save
produces: `presetSaveTokens` omits unset properties and the reloaded payload re-merges the baseline.

`presetRows` now carries `tokens` on each row so the overlay has that stored map to merge onto.

This applies to every preset screen sharing the overlay helper, not only the Button.

## Known limitation

Clearing a property that was previously overridden keeps previewing the old override until save. The
value it will fall back to is a server-side baseline merge this client cannot reproduce, so the
choice is between a stale preview and a blank one — stale is the smaller lie, and it corrects itself
on save. Documented on `overlayPresetRows`.

## Testing

- `src/style-library/__tests__/presets.test.js` — the merge over stored tokens, unset properties
  falling through to the stored value, and the reference-identity contract (same array reference for
  a null draft or an unmatched item id; non-matching rows keep object identity).
- Full JS suite green on this branch alone: 129 suites / 1903 tests.
- CodeRabbit CLI against #1502's branch: one finding, which is the limitation documented above.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preset previews now preserve stored values when draft properties are unset.
  - Explicitly edited properties correctly override stored preset values.
  - Preview rendering now accurately reflects background, border radius, hover background, text color, and spacing settings.

### DIFF
--- a/src/style-library/__tests__/presets.test.js
+++ b/src/style-library/__tests__/presets.test.js
@@ -200,6 +200,11 @@ describe('presetRows', () => {
 			id: 'primary',
 			label: 'Primary',
 			userCreated: false,
+			tokens: {
+				'button-bg': '{semantic.color.action-primary}',
+				'button-text': '{semantic.color.on-primary}',
+				'button-radius': '0.5rem',
+			},
 			preview: {
 				background: '#3633e1',
 				color: '#ffffff',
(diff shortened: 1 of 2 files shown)


pauloiankoski:
@coderabbitai review

pauloiankoski:
@coderabbitai review

d4mation (approved):
Nice!

PR #1508, opened 2026-09-09
[DTM] SOFT-4320 [3/8]: answer the theme's palette reads with token colors

pauloiankoski:

## Summary

Puts token colors back into what the theme renders, without writing to the theme's stored palette.

The theme emits `--global-palette1…15`, and every other palette read, through
`kadence()->palette_option()`, which ends in the `kadence_palette_option` filter. Hooking that
filter makes tokens win wherever the theme renders a palette color while the stored Style Guide is
left alone — so resetting a token gives the user their own color back, and whichever palette set the
theme has active is honored, because the filter runs after the set is picked.

That last point fixes the set-2/set-3 bug for free: the removed write only ever touched
`['palette']`.

## What's in it

- `Projection/Kadence_Option/Palette_Filter.php` — returns the resolved token color for a slot a
  token claims, and the theme's own value otherwise. Gated on `Token_Registry::is_active()` and
  fails open on a corrupt stored document.
- Provider wiring: the filter at priority 10 with 2 args, plus a `Token_Store` changed-action
  listener at priority 5 — ahead of the projector's at 10, so a palette read later in the request
  that wrote a token cannot be answered from a stale memo.

A literal color is returned rather than a `var(--kb-token--…)` reference: the theme passes this
value through `hex2rgb()` in places and into the Customizer preview JS, where a `var()` would break.

## The bug this slice nearly shipped with

The first version memoized an empty map when the filter ran before the token declarations register
on `init:0` — and the theme reads the palette earlier than that. Every later read in the request
then returned the theme's value, so **a token override silently did nothing**. Found by running it
on a real site, not by the suite. The guard and its regression test are in this slice; the test
fails if the guard is removed.

## Test plan

`slic run wpunit Resources/Design_Tokens/Projection/Kadence_Option/Palette_FilterTest`

11 tests: claimed slots, pass-through for unclaimed and non-string keys, the deactivated-registry
guard, override-wins-after-write, independence from the active palette set, fail-open on a dangling
alias, the early-read guard, and the two hook wirings.

Verified on a live site: an override renders, deleting it falls back to the Customizer color rather
than the shipped default, and the stored option is untouched throughout.

## Summary by CodeRabbit

- **New Features**
  - Kadence theme palette values can now reflect resolved design-token colors.
  - Token-based palette colors update automatically when design tokens change.
  - Palette values remain independent of the currently active theme palette selection.

- **Bug Fixes**
  - Unclaimed palette entries and inactive token registries continue using existing theme values.
  - Invalid token data no longer prevents palette values from being read.

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
+++ b/includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
@@ -66,7 +66,8 @@ final class Palette_Filter {
 	private Palette_Builder $builder;
 
 	/**
-	 * Per-request memo of slug => { color, name }. Null until first built; cleared on a token write.
+	 * Per-request memo of slug => { color, name }. Null until first built; cleared when a token is written
+	 * or the active library changes, the two inputs it is built from.
 	 *
 	 * @since TBD
 	 *
@@ -116,8 +117,9 @@ final class Palette_Filter {
 	}
 
 	/**
-	 * Drop the per-request memo so a token written in this request is visible to later palette reads
-	 * in the same request. Bound to the store's changed action.
+	 * Drop the per-request memo so a change made in this request is visible to later palette reads in the
+	 * same request. Bound to both inputs the memo is built from: the store's changed action and the
+	 * active-library pointer's.
 	 *
 	 * @since TBD
 	 *
--- a/includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
+++ b/includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
@@ -20,7 +20,12 @@ use RuntimeException;
  * because the filter runs after the theme picked the set.
  *
  * A literal color is returned rather than a var(--kb-token--…) reference: the theme passes this value
- * through hex2rgb() in places and the Customizer live preview writes it straight into CSS.
+ * through hex2rgb() in places, and a var() reference cannot be converted that way.
+ *
+ * Known limit: the Customizer live preview never reaches this filter. The theme registers the palette
+ * setting with postMessage transport and its preview script writes the control value straight onto
+ * --global-paletteN, so while a palette color is being edited the preview shows the theme's own color
+ * even for a slot a token overrides. The token color is back on the next full page load.
  *
  * Fail-open: a corrupt stored document (alias cycle / dangling alias) leaves every read untouched.
  * Gated on Token_Registry::is_active() so the fail-closed guard is honored.
--- a/includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
+++ b/includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
@@ -1,0 +1,164 @@
+<?php declare( strict_types=1 );
+// cspell:ignore palette subkey .
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection\Kadence_Option;
+
+use KadenceWP\KadenceBlocks\Design_Tokens\Database\Active_Token_Library_Store;
+use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Token_Registry;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
+use RuntimeException;
+
+/**
+ * Answers the Kadence theme's palette reads with resolved token colors.
+ *
+ * The theme emits --global-palette1…15 (and every other palette read) through
+ * kadence()->palette_option(), which ends in the kadence_palette_option filter. Hooking that filter
+ * makes tokens win wherever the theme renders a palette color while the stored kadence_global_palette
+ * option — the user's Style Guide — is never written. That keeps the Style Guide intact as the source
+ * the token baseline reads from, so resetting a token gives the user their own color back, and it
+ * works for whichever palette set (palette / second-palette / third-palette) the theme has active,
+ * because the filter runs after the theme picked the set.
+ *
+ * A literal color is returned rather than a var(--kb-token--…) reference: the theme passes this value
+ * through hex2rgb() in places and the Customizer live preview writes it straight into CSS.
+ *
+ * Fail-open: a corrupt stored document (alias cycle / dangling alias) leaves every read untouched.
+ * Gated on Token_Registry::is_active() so the fail-closed guard is honored.
+ *
+ * @since TBD
+ */
+final class Palette_Filter {
+
+	/**
+	 * The token registry.
+	 *
+	 * @since TBD
+	 *
+	 * @var Token_Registry
+	 */
+	private Token_Registry $registry;
+
+	/**
+	 * The token resolver.
+	 *
+	 * @since TBD
+	 *
+	 * @var Token_Resolver
+	 */
+	private Token_Resolver $resolver;
+
+	/**
+	 * Owns the active-library pointer, read at filter time so the projected colors follow the active library.
+	 *
+	 * @since TBD
+	 *
+	 * @var Active_Token_Library_Store
+	 */
+	private Active_Token_Library_Store $active;
+
+	/**
+	 * Builds the slug => { color, name } map from the resolved tokens.
+	 *
+	 * @since TBD
+	 *
+	 * @var Palette_Builder
+	 */
+	private Palette_Builder $builder;
+
+	/**
+	 * Per-request memo of slug => { color, name }. Null until first built; cleared on a token write.
+	 *
+	 * @since TBD
+	 *
+	 * @var array<string, array{color: string, name: string}>|null
+	 */
+	private ?array $entries = null;
+
+	/**
+	 * @since TBD
+	 *
+	 * @param Token_Registry             $registry The token registry.
+	 * @param Token_Resolver             $resolver The token resolver.
+	 * @param Active_Token_Library_Store $active   The active-library pointer.
+	 * @param Palette_Builder            $builder  The palette entries builder.
+	 */
+	public function __construct(
+		Token_Registry $registry,
+		Token_Resolver $resolver,
+		Active_Token_Library_Store $active,
+		Palette_Builder $builder
+	) {
+		$this->registry = $registry;
+		$this->resolver = $resolver;
+		$this->active   = $active;
+		$this->builder  = $builder;
+	}
+
+	/**
+	 * Return the resolved token color for a palette slot a token claims, or the theme's own value.
+	 *
+	 * @since TBD
+	 *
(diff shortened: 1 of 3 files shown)


includes/resources/Design_Tokens/Projection/Kadence_Option/Provider.php
```diff
@@ -33,5 +35,18 @@ public function register(): void {
...
+
+		// The Kadence theme's palette is projected at read time, never written: every palette_option()
+		// read passes through this filter. Registered unconditionally — the theme applies the filter
+		// only when it is active, so on other themes the callback never runs.
+		add_filter( 'kadence_palette_option', $this->container->callback( Palette_Filter::class, 'filter' ), 10, 2 );
+
+		// Ahead of the projector's own listener, so a palette read later in the write request cannot be
+		// answered from a memo built before the write.
+		add_action(
+			Token_Store::changed_action(),
```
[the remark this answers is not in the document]
pauloiankoski: Applied in 1051355f5.

The finding is right about the mechanism. `entries()` builds the memo from `$this->active->get()`, and only `Token_Store::changed_action()` cleared it — so the memo depends on two inputs and one of them could move without invalidating it. Registered the same callback on `Active_Token_Library_Store::changed_action()` at the same priority, and updated the two docblocks that said the memo is cleared "on a token write".

Worth recording that nothing in production listens to `Active_Token_Library_Store::changed_action()` today, so I could not find a request that reads the palette both before and after a switch — the theme's `palette_option()` reads happen while rendering, not during the REST call that moves the pointer. That makes this a latent trap rather than a live bug, which is exactly why it is worth closing: the next thing to subscribe to that action would inherit it silently.

Added `testActiveLibrarySwitchClearsTheMemoThroughTheHook`, mirroring the existing token-write case: it writes a second library with a different `brand.primary`, builds the memo against the default, switches the pointer, and asserts the next read follows the new library. Confirmed it fails with the hook removed and passes with it.

includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
```diff
@@ -0,0 +1,166 @@
...
+
+	/**
+	 * Per-request memo of slug => { color, name }. Null until first built; cleared when a token is written
+	 * or the active library changes, the two inputs it is built from.
+	 *
+	 * @since TBD
+	 *
+	 * @var array<string, array{color: string, name: string}>|null
+	 */
+	private ?array $entries = null;
```
d4mation, line 81: This says the memo is cleared on "the two inputs it is built from". Once slices 5 and 7 land there is a third: `entries()` calls `Token_Resolver::resolve()`, which resolves through the theme-decorated baseline, so the theme's Style Guide feeds this map too. That makes this the only cache in the stack not keyed on `Effective_Version`.

This doesn't fail with the current code, but it is possible. Declarations register on `init:0`, so the `after_setup_theme` read returns empty without memoizing, and every later read happens after `wp_loaded`, so after the `customize_preview_init` flush. Either hook `on_tokens_changed()` to the same `add_option_`/`update_option_`/`delete_option_kadence_global_palette` hooks `Theme_Style_Guide\Provider` uses, or say here why the third input cannot go stale.

includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
```diff
@@ -0,0 +1,166 @@
...
+			return $this->entries;
+		}
+
+		// Not memoized: declarations register on init:0, and the theme reads the palette earlier than that
+		// (the Customizer and the theme's own CSS both call palette_option()). Caching the empty map here
+		// would answer every later read in the request with the theme's value, so a token override would
+		// silently do nothing.
+		if ( ! $this->builder->has_palette_tokens() ) {
+			return [];
+		}
```
d4mation, line 162: Probably worth recording the consequence of this guard. Palette reads before `init:0` are no longer token-valued. The theme calls `kadence()->palette_option( 'palette1' .. 'palette9' )` from `add_theme_support( 'editor-color-palette', ... )` on `after_setup_theme` (`inc/components/color_palette/component.php:87`), so those entries now capture the theme's raw colors where the old option-write approach gave them token colors.

It's a minor thing because the Kadence theme ships a `theme.json` whose `theme-palette1..N` presets are `var(--global-paletteN)` and win in the block editor. But `Kadence_Blocks_Settings::load_color_palette()` reads `get_theme_support( 'editor-color-palette' )` directly and will carry the raw values.

includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
```diff
@@ -0,0 +1,166 @@
...
+ * The theme emits --global-palette1…15 (and every other palette read) through
+ * kadence()->palette_option(), which ends in the kadence_palette_option filter. Hooking that filter
+ * makes tokens win wherever the theme renders a palette color while the stored kadence_global_palette
+ * option — the user's Style Guide — is never written. That keeps the Style Guide intact as the source
+ * the token baseline reads from, so resetting a token gives the user their own color back, and it
+ * works for whichever palette set (palette / second-palette / third-palette) the theme has active,
+ * because the filter runs after the theme picked the set.
+ *
+ * A literal color is returned rather than a var(--kb-token--…) reference: the theme passes this value
+ * through hex2rgb() in places and the Customizer live preview writes it straight into CSS.
```
d4mation: This cites the Customizer live preview as the reason to return a literal instead of a `var()`, but it is worth clarifying thought that live preview does not reach this filter at all. The theme registers `kadence_global_palette` with `transport => postMessage`, and its `live_palette` handler (`inc/customizer/js/kadence-customizer-preview.js:2501`) sets `--global-paletteN` as an inline style on `document.documentElement` straight from the control value.

So editing the palette in the Customizer previews the raw theme color even for a slot that has a stored token override, and it snaps back to the token color on reload.

d4mation (commented):
One stale cross-reference outside this diff: `includes/resources/Design_Tokens/Projection/Css_Var/Legacy_Filter_Bridge.php:56` still says token-driven palette values under the Kadence theme are "handled by a separate palette projector". Slice 2 removed the option write, and the mechanism is now the read-time `kadence_palette_option` filter this PR adds. `Kadence_Palette_Slot`'s class docblock got the equivalent update in slice 2, so it looks like this one was missed.

includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
pauloiankoski: Fixed in 14ce0dc2d, on this PR. The guard comment now carries the consequence:

> Reads that early still get the theme's own color, not the token's. The theme fills its editor-color-palette theme support from palette_option() on after_setup_theme, so those entries carry the raw Style Guide colors, and the plugin's load_color_palette() (after_setup_theme:999) merges them into its own palette as-is. The block editor does not show them: the theme's theme.json defines the same slugs as var(--global-paletteN), which this filter feeds once the declarations are in, and theme.json wins over the theme support by slug.

Checked every step of your description and it holds: `color_palette/component.php:38` hooks the registration on `after_setup_theme`, the `palette_option()` calls are at `:87-157`, and `class-kadence-blocks-settings.php:509-511` merges `reset( $theme_palette )` through.

No follow-up ticket, for two reasons. The `theme.json` presets use the same `theme-palette1…15` slugs and WordPress merges `theme.json` over the theme support by slug, so the editor never shows the raw values — and `load_color_palette_theme_json()` prefers `theme.json` too. And the old option-write path was not better: the theme support only carried token colors because a *previous* request had written them, so it was already one request stale after any token change. The narrow remaining case is a third-party plugin reading `get_theme_support( 'editor-color-palette' )` and printing the hex, which was equally stale before.

If product does want those entries token-valued later, the two candidate fixes (re-register after `init:0`, or move the declarations earlier) both change timing this module depends on, so that belongs in its own ticket rather than in this stack.

includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
pauloiankoski: Fixed in dacba7012, on this PR. The live-preview clause is gone from the justification and the behavior is recorded as what it is:

> A literal color is returned rather than a var(--kb-token--…) reference: the theme passes this value through hex2rgb() in places, and a var() reference cannot be converted that way.
>
> Known limit: the Customizer live preview never reaches this filter. The theme registers the palette setting with postMessage transport and its preview script writes the control value straight onto --global-paletteN, so while a palette color is being edited the preview shows the theme's own color even for a slot a token overrides. The token color is back on the next full page load.

Verified both halves in the theme: `'transport' => 'postMessage'` at `color_palette/component.php:52`, and `live_palette` setting `--global-` + slug as an inline style on `document.documentElement` at `kadence-customizer-preview.js:2501-2519`.

includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
pauloiankoski: Fixed in 238c366d1, on #1511 rather than here. The third input only exists once that PR binds the decorator into `Registry/Provider.php` — on this branch the resolver still reads the plain `Json_Baseline_Document` — so registering the clear here would guard an input that does not exist yet.

`Kadence_Option\Provider` now hooks `on_tokens_changed` to the same three `*_option_kadence_global_palette` hooks and to `customize_preview_init`, and the `$entries` and `on_tokens_changed()` docblocks name all four inputs instead of "the two inputs it is built from".

One thing your note did not mention that turned out to matter: **ordering**. The memo is rebuilt from the resolver and the resolver reads the overlay, so the clear has to run *after* the overlay flush, not before. Otherwise a palette read between the two rebuilds the memo from the stale overlay and nothing clears it again in that request. The overlay flushes at 10 on the option hooks and 0 on `customize_preview_init`, so the clear is at 20 on all four, with the reason in the comment.

The test is behavioral, not a registration check. The suite has no Kadence theme, so the container's overlay always reads empty — it swaps the overlay's private `$source` for `Fake_Style_Guide_Source` by reflection (the same trick the existing early-read test uses on `Palette_Builder::$registry`), which runs the real overlay → decorator → resolver → filter chain against a Style Guide the test controls. A probe read on the same hook one priority before the flush is what proves the ordering rather than just the wiring.

Three checks on it:

- 4 cases fail with the hooks removed, nothing else fails.
- 4 cases fail with the priorities moved to 5 / -5, i.e. before the flush — so the priority is load-bearing and pinned.
- Green with the change as shipped, and `Resources/Design_Tokens/Theme_Style_Guide` stays green (45 tests), confirming `tearDown()` puts the real reader back.

Full suite 2841 tests, PHPStan clean.

d4mation (approved):

PR #1510, opened 2026-09-09
[DTM] SOFT-4320 [6/8]: key every cache on the effective document version

pauloiankoski:

## Summary

Keys every resolved-tokens and projected-CSS cache on the **effective** document version rather than
the store version alone.

A Customizer save changes no store version. Once the theme's Style Guide starts feeding the baseline
(slice 6), the old colors would keep being served from cache until the next token write. This adds
`Resolver/Effective_Version`: the store version, extended with the Style Guide overlay's signature
when there is one.

On a site with no overlay — anything but the Kadence theme — it returns the store version unchanged,
so those cache keys are byte-for-byte what they are today.

## What's in it

`Effective_Version` plus all eleven cache sites switched to it:

- `Token_Resolver` — `load()`, `effective_document()`, `resolve_palette()`
- the `Css_Var`, `Palette`, `Block_Default_Css` (×2) and `Preset` (×2) projectors
- `Preset/Css_Builder::collect()`
- the `Kadence_Option` sync marker

Left alone deliberately: `Feed_Assembler` and the REST controllers keep comparing the store version.
Those are optimistic concurrency, not caching — a Customizer save does not change the stored document
a client is editing.

## A cleanup that fell out of it

With every read going through the new service, `Token_Store` became write-only in six classes.
PHPStan flagged it, so the dead dependency is removed rather than left wired. That changed
constructor signatures, hence the test churn: eleven hand-built `Token_Resolver` call sites and three
projector/builder ones.

## Test plan

`slic run wpunit Resources/Design_Tokens`

New `Effective_VersionTest`: equals the store version when the overlay is empty, differs when the
Style Guide differs, is stable when it does not, keeps the store version as its prefix, and still
changes on a token write.

`grep -rn '\->store->get_version(' includes/resources/Design_Tokens/{Projection,Resolver}` should
return only `Effective_Version` itself.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved design-token cache invalidation so changes to theme Style Guide overlays are reflected in generated CSS, presets, palettes, and resolved token data.
  - Preserved fail-open behavior by returning empty stylesheets when version resolution fails.

- **Tests**
  - Added coverage for effective version handling, including overlay changes, stable styles, stored document updates, and cache behavior.
  - Updated existing tests to validate the revised version resolution flow.

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
+++ b/includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
@@ -4,8 +4,8 @@
 namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection\Kadence_Option;
 
 use KadenceWP\KadenceBlocks\Design_Tokens\Database\Active_Token_Library_Store;
-use KadenceWP\KadenceBlocks\Design_Tokens\Database\Token_Store;
 use KadenceWP\KadenceBlocks\Design_Tokens\Registry\Token_Registry;
+use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Effective_Version;
 use KadenceWP\KadenceBlocks\Design_Tokens\Resolver\Token_Resolver;
 use RuntimeException;
 
@@ -66,14 +66,6 @@ final class Projector {
 	 */
 	private Token_Resolver $resolver;
 
-	/**
-	 * The token store.
-	 *
-	 * @since TBD
-	 *
-	 * @var Token_Store
-	 */
-	private Token_Store $store;
 
 	/**
 	 * Owns the active-library pointer, read at sync time so the synced options follow the active library.
@@ -100,27 +92,38 @@ final class Projector {
 	 */
 	private bool $reconciled_this_request = false;
 
+
+	/**
+	 * Supplies the sync signature's version part: the store version, plus the theme Style Guide
+	 * signature when there is one, so a Customizer save re-syncs KB's own palette option too.
+	 *
+	 * @since TBD
+	 *
+	 * @var Effective_Version
+	 */
+	private Effective_Version $versions;
+
 	/**
 	 * @since TBD
 	 *
-	 * @param Token_Registry             $registry
-	 * @param Token_Resolver             $resolver
-	 * @param Token_Store                $store
-	 * @param Active_Token_Library_Store $active
-	 * @param Palette_Builder            $builder
+	 * @param Token_Registry             $registry The token registry.
+	 * @param Token_Resolver             $resolver The token resolver.
+	 * @param Active_Token_Library_Store $active   The active-library pointer.
+	 * @param Palette_Builder            $builder  The palette entries builder.
+	 * @param Effective_Version          $versions Supplies the sync signature's version part.
 	 */
 	public function __construct(
 		Token_Registry $registry,
 		Token_Resolver $resolver,
-		Token_Store $store,
 		Active_Token_Library_Store $active,
-		Palette_Builder $builder
+		Palette_Builder $builder,
+		Effective_Version $versions
 	) {
 		$this->registry = $registry;
 		$this->resolver = $resolver;
-		$this->store    = $store;
 		$this->active   = $active;
 		$this->builder  = $builder;
+		$this->versions = $versions;
 	}
 
 	/**
@@ -166,7 +169,7 @@ final class Projector {
 		}
 
 		$slug      = $this->active->get();
-		$signature = KADENCE_BLOCKS_VERSION . ':' . $slug . ':' . $this->store->get_version( $slug );
+		$signature = KADENCE_BLOCKS_VERSION . ':' . $slug . ':' . $this->versions->for_slug( $slug );
 
 		// Skip the resolve + write when nothing the sync depends on has changed since the last successful
 		// one. The signature names all three inputs rather than relying on the version alone: a library
--- a/includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
+++ b/includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
@@ -39,8 +39,8 @@ final class Projector {
 	private const KB_COLORS_OPTION = 'kadence_blocks_colors';
 
 	/**
-	 * Marker option storing the last-synced "{plugin-version}:{library-slug}:{store-version}" signature, so a request
-	 * where nothing changed skips resolution entirely.
+	 * Marker option storing the last-synced "{plugin-version}:{library-slug}:{effective-version}" signature, so a
+	 * request where nothing changed skips resolution entirely.
 	 *
 	 * @since TBD
 	 *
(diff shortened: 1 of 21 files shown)


pauloiankoski:
Not applying — the premise is false. The test passes on this branch.

```
$ git log --oneline -1
c3219321d Key every resolved and projected cache on the effective document version

$ slic run wpunit Resources/Design_Tokens/Projection/Kadence_Option/ProjectorTest
✔ ProjectorTest: Introducing the theme palette does not flip the signature (0.01s)
OK (15 tests, 28 assertions)
```

The reason is `Style_Guide_Reader::snapshot()`, which returns `null` unless `Kadence\Theme` exists and the `Kadence\kadence` accessor is callable. The Kadence theme is not installed in the wpunit suite, so the overlay resolves to `[]`, `Style_Guide_Overlay::signature()` short-circuits to `''`, and `Effective_Version::for_slug()` returns the bare store version. Creating `kadence_global_palette` in a suite with no Kadence theme changes nothing about the effective version.

There is a smaller real point in here, though, which I want on the record rather than buried.

On a site that *does* run the Kadence theme, this PR does change that behavior: the overlay has values, so a Style Guide edit changes the signature, the marker moves, and the sync re-runs. That is intended and is the reason `Effective_Version` exists — `kadence_blocks_colors` is projected from resolved token values, and those now derive from the Style Guide, so a Style Guide change has to re-project.

So the signature-stability half of this test holds because the suite has no Kadence theme, not as a universal invariant. What the test is really guarding is its second assertion — that the theme's option comes back byte-for-byte unchanged — which is the regression that mattered. Its name promises a little more than that now, and I am happy to tighten the docblock to say so explicitly if that reads better; flagging rather than silently rewording, since it is a test whose history is deliberate.

*(That run was on `c3219321d`, which later cascade rebases replaced. The same commit is `8c2e5014a` on the branch now.)*

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
```diff
@@ -143,14 +156,14 @@ public function resolve( string $slug = 'default' ): Resolved_Tokens {
 	 * @throws Dangling_Alias_Exception When a stored alias references a path with no token leaf.
 	 */
 	private function load( string $slug, string $cache_prefix ): Resolved_Tokens {
-		$version   = $this->store->get_version( $slug );
+		$version   = $this->versions->for_slug( $slug );
```
d4mation, line 160: The docblock above at line 145 still says "both keyed on the store version". It is the effective version now, which also changes on a Customizer save with no write - the point of this slice. `resolve()`, `effective_document()`, and `resolve_palette()` in this file were all updated; `load()` was missed.

includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
```diff
@@ -166,7 +169,7 @@ private function sync(): void {
 		}
 
 		$slug      = $this->active->get();
-		$signature = KADENCE_BLOCKS_VERSION . ':' . $slug . ':' . $this->store->get_version( $slug );
+		$signature = KADENCE_BLOCKS_VERSION . ':' . $slug . ':' . $this->versions->for_slug( $slug );
```
d4mation, line 172: The third component is the effective version now, but two comments still spell the marker as `"{plugin-version}:{library-slug}:{store-version}"` - the constant's docblock at line 42 and the inline comment at line 201.

includes/resources/Design_Tokens/Projection/Palette/Projector.php
```diff
@@ -94,33 +88,44 @@ final class Projector extends Abstract_Css_Projector {
 	 */
 	private array $memo = [];
 
+
```
d4mation: Doubled blank line here and at line 44, left over from swapping `Token_Store $store` for `Effective_Version $versions`. Also, the new `use ...\Resolver\Effective_Version;` is appended after `use Throwable;` rather than in namespace order, and the memo docblock at line 83 still says "keyed on the active slug and store version".

includes/resources/Design_Tokens/Projection/Palette/Projector.php
```diff
@@ -180,7 +185,7 @@ public function enqueue_editor(): void {
 	public function css(): string {
 		try {
 			$active  = $this->active->get();
-			$version = $this->store->get_version( $active );
+			$version = $this->versions->for_slug( $active );
```
d4mation, line 186: The docblock at line 178 still says "cached on the store version" - effective version.

includes/resources/Design_Tokens/Projection/Preset/Css_Builder.php
```diff
@@ -157,17 +151,31 @@ final class Css_Builder {
...
 
+	/**
+	 * Supplies the cache version: the store version, plus the theme Style Guide signature when there is
+	 * one, so a Customizer save invalidates this cache even though it bumps no store version.
+	 *
+	 * @since TBD
+	 *
+	 * @var Effective_Version
+	 */
+	private Effective_Version $versions;
```
d4mation, line 162: This property is described correctly, but four places in this file still say store version: the memo docblock at line 144 ("keyed on the library slug AND its store version"), and `@param string $version The store version the active library was built from.` at lines 228, 247, and 281. `Preset\Projector::css()` and `editor_css()` pass the effective version to all three.

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
pauloiankoski: `load()` was missed — the docblock says store version while the body reads `versions->for_slug()`. Fixing with the rest.

includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
pauloiankoski: Both spellings are stale: the third component is the effective version now. Updating the constant docblock and the inline comment.

includes/resources/Design_Tokens/Projection/Palette/Projector.php
pauloiankoski: All three stand — the `use` sits after `use Throwable;`, the memo docblock still says store version, and the two doubled blank lines are leftovers. Fixing.

includes/resources/Design_Tokens/Projection/Palette/Projector.php
pauloiankoski: Line 178 still says store version. Fixing together with the memo docblock above it.

includes/resources/Design_Tokens/Projection/Preset/Css_Builder.php
pauloiankoski: All four are stale — line 144 and the three `@param $version` at 228, 247 and 281. `Preset\Projector` passes the effective version to each.

includes/resources/Design_Tokens/Resolver/Token_Resolver.php
pauloiankoski: Fixed in 0cf20dca1. `load()` now says effective version, spelled out as the store version plus the Style Guide signature so both a token write and a Customizer save self-invalidate.

includes/resources/Design_Tokens/Projection/Kadence_Option/Projector.php
pauloiankoski: Fixed in 0cf20dca1 — both now read `{plugin-version}:{library-slug}:{effective-version}`. Also updated the third mention at line 193, which said a token write flips the signature without mentioning a Customizer save.

includes/resources/Design_Tokens/Projection/Palette/Projector.php
pauloiankoski: Fixed in 0cf20dca1: `Effective_Version` moved between `Effective_Palettes` and `Token_Resolver`, both doubled blank lines gone, memo docblock says effective version.

includes/resources/Design_Tokens/Projection/Palette/Projector.php
pauloiankoski: Fixed in 0cf20dca1. Same line also had British `memoised`, now `memoized` — cspell would have caught it on the next touch.

includes/resources/Design_Tokens/Projection/Preset/Css_Builder.php
pauloiankoski: Fixed in 0cf20dca1, all four. Also took lines 219-222, which still said store version twice: true but incomplete, and it is what a grep for the phrase lands on.

d4mation (approved):

PR #1511, opened 2026-09-09
[DTM] SOFT-4320 [7/8]: resolve tokens from the theme's Style Guide

pauloiankoski:

## Summary

Makes token resolution start from the colors the site already renders.

`Theme_Style_Guide_Baseline_Document` decorates the shipped baseline: `document()` returns it with
the theme's Style Guide values written onto the leaves they name, so the effective document — and
therefore every resolved token — starts from the site's existing look instead of the plugin's
shipped defaults. Stored overrides still merge on top, so tokens always win.

This is the slice the previous four were building toward.

## Why `has()` delegates

`has()` passes straight through to the shipped baseline. That is the rule "the theme may only
re-value tokens that already have a baseline entry", enforced structurally rather than by
convention: a theme value whose dot-path has no `$value` leaf is skipped, no id is ever added, and
`Baseline_Guard` still sees exactly the shipped set.

The `$default` color palette's swatches are re-valued alongside the tokens they point at — and only
for the leaves actually written — so the Style Library shows the theme's colors and "reset swatch"
restores them rather than the shipped hex.

## Also here

`Registry/declarations.php` claimed the baseline "mirrors Kadence's default palette". It does not:
it mirrors the plugin's **own** no-theme fallback, which differs from the theme at slots 1 and 2.
Comment corrected.

The binding in `Registry/Provider.php` is a closure, because neither the theme nor the declarations
are loaded when providers register.

## Test plan

`slic run wpunit Resources/Design_Tokens`

- `Theme_Style_Guide_Baseline_DocumentTest` — `has()` delegation, re-valuing a leaf while keeping its
  `$type` and siblings, never adding an id, the default-palette swatches, an empty inner document
  staying empty, equality with the inner document when there is no theme, and rebuilding when the
  Style Guide changes.
- `Style_Guide_IntegrationTest` — end to end over the real shipped baseline, registry and store,
  covering a fresh library resolving to the theme palette, the active set being honored, an override
  winning and its removal restoring the theme color, a non-Kadence site resolving byte-for-byte as
  shipped, a Customizer save reflected with no store write, the guard, and the swatches.

Verified in a browser on a site with a customized palette: the front end and a refreshed Customizer
preview agree, an override wins in both, and removing it returns the Customizer's color.

## Known limitation

While a color is being dragged in the Customizer, the preview shows the **theme's** value rather
than the token's. The theme's preview script writes `--global-paletteN` inline on the document
(`inc/customizer/js/kadence-customizer-preview.js`), which never reaches PHP, and an inline root
style outranks anything the plugin emits. It corrects on refresh. Fixing it properly needs a small
hook in the theme, so it ships as a coordinated follow-up rather than here. Accepted in review.

## Summary by CodeRabbit

* **New Features**
  * Design-token colors now adapt to the active theme’s Style Guide palette.
  * Theme-based values apply to existing tokens while preserving plugin defaults and metadata.
  * Default palette swatches reflect the active theme’s colors.
  * Theme updates are reflected without requiring a design-token store write.
  * Non-Kadence sites continue using the shipped token values.

* **Documentation**
  * Clarified the distinction between plugin-provided baseline colors and theme palette values.

* **Tests**
  * Added coverage for theme mapping, overrides, resets, palette selection, and fallback behavior.

### DIFF
--- a/includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
+++ b/includes/resources/Design_Tokens/Projection/Kadence_Option/Palette_Filter.php
@@ -71,8 +71,8 @@ final class Palette_Filter {
 	private Palette_Builder $builder;
 
 	/**
-	 * Per-request memo of slug => { color, name }. Null until first built; cleared when a token is written
-	 * or the active library changes, the two inputs it is built from.
+	 * Per-request memo of slug => { color, name }. Null until first built; cleared when any input it is
+	 * built from changes: a token write, an active-library switch, or a change to the theme's Style Guide.
 	 *
 	 * @since TBD
 	 *
@@ -123,8 +123,8 @@ final class Palette_Filter {
 
 	/**
 	 * Drop the per-request memo so a change made in this request is visible to later palette reads in the
-	 * same request. Bound to both inputs the memo is built from: the store's changed action and the
-	 * active-library pointer's.
+	 * same request. Bound to every input the memo is built from: the store's changed action, the
+	 * active-library pointer's, the theme's palette option hooks and the Customizer preview init.
 	 *
 	 * @since TBD
 	 *
--- a/includes/resources/Design_Tokens/Projection/Kadence_Option/Provider.php
+++ b/includes/resources/Design_Tokens/Projection/Kadence_Option/Provider.php
@@ -4,12 +4,13 @@ namespace KadenceWP\KadenceBlocks\Design_Tokens\Projection\Kadence_Option;
 
 use KadenceWP\KadenceBlocks\Design_Tokens\Database\Active_Token_Library_Store;
 use KadenceWP\KadenceBlocks\Design_Tokens\Database\Token_Store;
+use KadenceWP\KadenceBlocks\Design_Tokens\Theme_Style_Guide\Style_Guide_Reader;
 use KadenceWP\KadenceBlocks\StellarWP\ProphecyMonorepo\Container\Contracts\Provider as Provider_Contract;
 
 /**
  * Registers the Kadence option projection: binds the builder, projector and palette filter as
  * singletons, wires the reconcile to a once-per-request boot pass and to the token-changed action,
- * and hooks the theme's palette reads.
+ * hooks the theme's palette reads, and clears the palette memo on every input it is built from.
  *
  * @since TBD
  */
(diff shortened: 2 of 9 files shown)


pauloiankoski:
Re the nitpick on `Style_Guide_IntegrationTest` (line 22) — nothing to resolve, and nothing to rename.

There is no conflict between the two rules. The repo has one convention, applied consistently: a test class mirrors the name of the thing it covers and appends `Test`. When that name is Snake_Case the test is too, and when it is not, it is not:

```
206  test classes under tests/
 77  carry an underscore — Token_Store → Token_StoreTest,
     Active_Token_Library_Store → Active_Token_Library_StoreTest,
     Palette_Builder → Palette_BuilderTest
129  carry none — SingleIcon → SingleIconTest, AdvancedForm → AdvancedFormTest
```

The split is not two competing styles; it is the same rule producing different output for different subjects. "PascalCase test file names ending in `Test.php`" in the project guidelines is about the file being capitalized and `Test`-suffixed, not about stripping the underscores out of a Snake_Case subject — doing that would give `TokenStoreTest` for `Token_Store`, which appears nowhere in the 206.

`Style_Guide_IntegrationTest` follows it: the namespace is `Theme_Style_Guide`, the feature is the Style Guide integration, and the file is new in this PR rather than renamed. Renaming it would make it the only test in the suite not matching its subject.

d4mation (commented):
The behavior here invalidates a docblock in a file this PR does not touch and it is an important one. `includes/resources/Design_Tokens/Resolver/Effective_Palettes.php:305`. `baseline_swatch_values()` still says "the colors the plugin ships, independent of any stored library", and its `@return` at line 319 says "the shipped literal-or-alias value". On a Kadence site that is now false - the decorator re-values the `$default` palette's swatches from the theme's Style Guide before this method sees them, which `testDefaultPaletteSwatchesShowTheThemeColors` asserts.

The same docblock says "its VALUES are what a delete reverts one of those swatches to", which is exactly what changed: a delete now reverts to the theme's color, not the shipped hex. That is the intended behavior, but the method that defines it does not say so.

pauloiankoski:
Carries the fix for the third memo input raised on #1508 (`Palette_Filter`'s memo is built from the resolver, which reads the baseline this PR decorates with the theme's Style Guide).

`Kadence_Option\Provider` now clears the memo on the same three `*_option_kadence_global_palette` hooks and on `customize_preview_init`, at priority 20 so it runs after the overlay's own flush — clearing first would let a read in between rebuild the memo from stale overlay values with nothing to clear it again.

Registered from this PR rather than #1508 because that is where the input starts existing: on the earlier branches the resolver still reads the plain `Json_Baseline_Document`.

Full reasoning and the test evidence are on the #1508 thread.

pauloiankoski:
`baseline_swatch_values()` is theme-valued on a Kadence site now, so both the description and the `@return` are wrong. The delete-reverts-to line is the behavior change itself. Fixing both.

d4mation (approved):

PR #1512, opened 2026-09-09
[DTM] SOFT-4320 [4/8]: map a theme Style Guide snapshot onto token values

pauloiankoski:

## Summary

The pure translation from a Kadence theme Style Guide snapshot to `token id => $value`.

No WordPress calls, no globals, no I/O: raw theme data in, a token map out. Nothing uses it yet —
the reader that feeds it and the overlay that composes it are the next PR, and nothing consumes the
result until the baseline decorator.

Split out from the reader PR so the rules here can be reviewed on their own, without needing any
knowledge of the theme's WordPress surface.

## The rules it encodes

- **The active set wins.** The theme's `active` pointer names which of `palette`, `second-palette`
  or `third-palette` it renders; a missing or empty pointer means the first, and a pointer naming a
  set that does not exist yields nothing — matching the empty values the theme itself renders in
  that case.
- **`#FfFfFf` on `palette10` is a marker, not a colour.** The theme uses it to mean "derive the
  complement of palette1 at render time", so it must never become a token value.
- **Only colour literals.** Hex, `rgb()/rgba()`, `hsl()/hsla()`. An empty string, a CSS variable or
  a gradient is skipped, so a bad theme value cannot reach the resolver.
- **It can only emit ids the caller passed in.** The slot-to-token map comes from the registry, so
  the mapper can never introduce a token id of its own. That is what later lets the baseline
  decorator promise it adds no ids.

## Test plan

`slic run wpunit Resources/Design_Tokens/Theme_Style_Guide/Style_Guide_MapperTest`

Provider-driven: each of the three palette sets, both fallback shapes, a pointer naming a missing
set, every colour notation the theme's picker produces, every unusable value it can store, malformed
entries, a slot the set does not carry, and a 15-slot set proving the produced ids stay a subset of
what the caller asked for.

One behaviour is pinned deliberately rather than changed: a slot the theme leaves unusable is left
to its baseline value, so the plugin renders the shipped colour where the theme renders an empty
custom property. Raised as an open question rather than altered late.

## Summary by CodeRabbit

* **Improvements**
  * Improved theme palette handling by selecting the active palette and applying valid slot colors consistently.
  * Added safer handling for incomplete, malformed, or unmapped color entries.
  * Preserved existing defaults when a usable color is unavailable.

* **Tests**
  * Added comprehensive coverage for palette selection, fallback behavior, color validation, slot matching, and malformed data.

### DIFF
--- a/includes/resources/Design_Tokens/Theme_Style_Guide/Style_Guide_Mapper.php
+++ b/includes/resources/Design_Tokens/Theme_Style_Guide/Style_Guide_Mapper.php
@@ -1,0 +1,156 @@
+<?php declare( strict_types=1 );
+// cspell:ignore palette .
+
+namespace KadenceWP\KadenceBlocks\Design_Tokens\Theme_Style_Guide;
+
+/**
+ * Pure translation from a theme Style Guide snapshot to `token id => $value`.
+ *
+ * No WordPress calls, no globals, no I/O: the reader hands raw data in and the overlay takes the map
+ * out, so every rule here is unit-testable without a database or a theme.
+ *
+ * The palette half honors the theme's "active" pointer — palette / second-palette / third-palette —
+ * because that is the set the theme renders. Only slots the caller passes in (the tokens declaring a
+ * kadence_slot) are produced, so the mapper can never introduce a token id of its own.
+ *
+ * @since TBD
+ */
+final class Style_Guide_Mapper {
+
+	/**
+	 * The palette option key naming the set the theme renders.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const ACTIVE_KEY = 'active';
+
+	/**
+	 * The palette set the theme falls back to when "active" is missing or empty.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	private const DEFAULT_SET = 'palette';
+
+	/**
+	 * The theme stores this on palette10 to mean "derive the complement of palette1 at render time".
+	 * It is a marker, not a color, so it must never become a token value.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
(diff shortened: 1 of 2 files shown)


pauloiankoski:
Re the nitpick on `Style_Guide_Mapper::COLOR_PATTERN` (tighten `#[0-9a-f]{3,8}` to 3/4/6/8 digits) — not applying it, and the reason is behavioral rather than stylistic.

The pattern is a shape filter, not a CSS validator. A value that fails it is skipped, so the token keeps its shipped baseline hex, and `Palette_Filter` then answers the theme's own `palette_option()` read with that baseline. Rejecting a value is therefore not a no-op: it makes the plugin overwrite what the user has stored with a shipped default.

Walking the example through:

- **Today** — `#12345` passes, the token becomes `#12345`, and the theme's read is answered with `#12345`. The browser ignores it and the element falls back, exactly as it would with the plugin absent. Theme and blocks agree.
- **Tightened** — `#12345` is skipped, the token stays `#3182CE`, and the theme's palette read is answered with `#3182CE`. The site now renders blue where the user's option says `#12345`.

The second is the failure this whole ticket exists to remove: a shipped default winning over a stored user value. `active_set()` already states the rule — "a pointer naming a set that does not exist yields nothing (the theme renders empty values in that case, so re-valuing nothing keeps the two in step)". Staying in step with the theme is the invariant; CSS validity is the theme's business, and an invalid stored color is broken in the theme's own output either way.

Two smaller points. The `rgba?\([^)]*\)` branch accepts `rgb()` and `rgb(anything)`, so tightening only the hex branch would not make this a validator — it would just make it inconsistent. And the Customizer's color picker emits valid hex, so a 5- or 7-digit value only arrives through a hand-edited option or a broken import.

d4mation (approved):

PR #1513, opened 2026-09-09
[DTM] SOFT-4320 [8/8]: store only the default palette swatches the site changes

pauloiankoski:

Part of the SOFT-4320 stack. Depends on #1511.

## The problem

Editing one color in the Style Library froze the whole default palette.

Every write path rebuilds the palette node from `Effective_Palettes::palette()`, which is the baseline merged with the stored overrides. That node carries a `$value` for all 17 swatches whether or not the site ever chose any of them, and the node was persisted whole. One edit therefore turned all 17 into stored overrides.

After that the palette stopped following the baseline. On a Kadence site that means it stopped following the theme: change a color in the Customizer and the Style Library still shows the color captured at write time, with a Reset button next to it. On a site with no Kadence theme the same freeze happened against the shipped `baseline.json` hexes.

The swatch endpoints made this easy to hit, because they never passed through `prepare_for_storage()` at all.

## The fix

`write_palette_node()` — the one choke point all three write paths share — now reduces the default palette to what the site actually changed. A swatch whose color already matches the baseline keeps its row but drops its `$value`.

The row has to stay. It is part of the permanent swatch set `guard_baseline_swatches()` protects, and once anything is stored the stored `groups` list replaces the baseline's, so a dropped row would disappear from every palette. Dropping only the value is enough: `swatch_values_of()` already skips a swatch with no value "so the token keeps its baseline value".

Reading it back needed two things:

- **`Effective_Palettes::stored_swatch_values()`** — what a palette actually stores, read straight from the overrides with no baseline behind it. `swatch_values()` reads the effective section, and the default palette also lives in the baseline, so it could not answer "does this swatch have its own value?".
- **`Effective_Palettes::complete_swatch_values()`** — the color a token will render as: baseline, overlaid with the library default's swatches, overlaid with the palette's. The four call sites that needed a complete value set now use it.

`effective_view()` uses the first for the `overridden` flag, so the Reset control appears exactly when the site stored something, and falls back to the baseline for display so an unedited swatch still shows its real color.

## Behavior

| | Before | After |
|---|---|---|
| Edit one default swatch | 17 swatches stored | 1 stored |
| Change a Customizer color afterwards | Style Library keeps the old color | Every unedited swatch follows |
| Reset a swatch | Restores the shipped hex | Hands the token back to the baseline — the theme's color on a Kadence site |
| No Kadence theme | — | Unchanged: the baseline is the shipped `baseline.json` |

No migration. Nothing is running the Style Library yet, and a stored document with all 17 values still resolves the same way — it just keeps overriding until each swatch is reset.

## Tests

5 new tests, all of which fail without the fix:

- `Style_Guide_IntegrationTest` — editing one swatch leaves the rest following the theme across a Customizer change; resetting hands the token back to the theme.
- `Palettes_ControllerTest` — a full-node write and a single-swatch write each store only what changed; the effective view marks only stored swatches as overridden. These run against the undecorated shipped baseline, so they cover the no-theme case.

Full wpunit suite green: 2835 tests, 8971 assertions.

Verified on a Kadence site by reproducing the reported flow: set `brand.primary` in the Style Library, change palette 3 in the Customizer, and confirm the edited swatch keeps its color while the changed one follows and shows no Reset control.

https://claude.ai/code/session_012enLvkgTEK1APG8oUYUCzr

## Summary by CodeRabbit

* **Enhancements**
  * Palette views now include complete effective colors, including inherited and baseline values.
  * Editing a default palette stores only the swatches that differ from the baseline, while preserving the remaining palette structure.
  * Updating one swatch no longer interrupts other swatches from following theme-derived color changes.
  * Resetting a swatch removes its override and restores the corresponding theme color.
* **Bug Fixes**
  * Improved consistency between displayed palette values, stored overrides, and reset behavior.

### DIFF
--- a/includes/resources/Design_Tokens/Rest/V1/Palettes_Controller.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Palettes_Controller.php
@@ -734,10 +734,10 @@ final class Palettes_Controller extends Controller {
 	 *   - `baseline` — whether the shipped palette defines this swatch. A baseline swatch's row is permanent
 	 *     (see {@see guard_baseline_swatches()}), so the editor offers Reset for it and Delete only for the
 	 *     rest.
-	 *   - `overridden` — whether there is anything to undo. On the DEFAULT palette that means the value
-	 *     differs from the shipped one; on any other palette it means the palette stores its own delta rather
-	 *     than inheriting. Measuring the default palette against its own deltas would mark every swatch
-	 *     overridden, since the default stores them all.
+	 *   - `overridden` — whether there is anything to undo: the palette stores its own value for the swatch.
+	 *     One test serves every palette, the default included, because the default palette stores only the
+	 *     swatches the site changed; a swatch left at the baseline keeps its row but stores no `$value`, so
+	 *     it reads as not overridden and there is nothing to undo.
 	 *
 	 * @since TBD
 	 *
--- a/includes/resources/Design_Tokens/Rest/V1/Palettes_Controller.php
+++ b/includes/resources/Design_Tokens/Rest/V1/Palettes_Controller.php
@@ -392,9 +392,9 @@ final class Palettes_Controller extends Controller {
 			return $locked;
 		}
 
-		// The default palette stores every swatch (it is the base). A non-default palette stores only its
-		// DELTAS: a swatch equal to the default palette's value is inherited, not persisted, so the palette
-		// resolves against the default for everything it does not change.
+		// The default palette keeps every swatch row (it is the structure template). A non-default palette
+		// stores only its DELTAS: a swatch equal to the default palette's value is inherited, not persisted, so
+		// the palette resolves against the default for everything it does not change.
 		$node = $this->prepare_for_storage( $node, $slug, $id === $this->palettes->default_palette( $slug ) );
 
 		// Replace the palette node wholesale (remove-then-merge) rather than a plain merge. A delta-reduced node
@@ -511,7 +511,7 @@ final class Palettes_Controller extends Controller {
 			// instead — a token that is stale or no longer a registered color must still be rejected. A
 			// token this palette has never stored has no current value here; that genuinely-new-token case
 			// is left to the downstream DTCG schema validation, which already rejects it.
-			$current_value = $this->palettes->swatch_values( $id, $slug )[ $token ] ?? null;
+			$current_value = $this->palettes->complete_swatch_values( $id, $slug )[ $token ] ?? null;
 
 			if ( is_string( $current_value ) ) {
 				$guard = $this->guard_swatch_target( $id, $token, $current_value );
@@ -527,7 +527,7 @@ final class Palettes_Controller extends Controller {
 		}
 
 		if ( isset( $fields['value'] ) ) {
-			$default = $this->palettes->swatch_values( $default_id, $slug );
+			$default = $this->palettes->complete_swatch_values( $default_id, $slug );
 
 			// A non-default swatch equal to the default value is inherited, not stored, so setting it back to
 			// the default reverts it — identical to a DELETE. Only applies to a value write; a label-only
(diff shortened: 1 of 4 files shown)


includes/resources/Design_Tokens/Rest/V1/Palettes_Controller.php
```diff
@@ -1381,12 +1389,77 @@ private function template_slot_for( string $token, string $slug ): array {
 	 * @return WP_REST_Response|WP_Error
 	 */
 	private function write_palette_node( string $slug, string $id, array $node ) {
+		if ( $id === $this->palettes->default_palette( $slug ) ) {
+			$node = $this->reduce_default_to_overrides( $node );
+		}
```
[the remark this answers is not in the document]
pauloiankoski: The reading of the code is right — a document written before this change keeps a `$value` on every swatch, and `stored_swatch_values()` reports each of those as an override. No migration ships anyway, and the reason is that there is no data to migrate: the Style Library has not shipped, so no site has a stored palette document. A versioned migration here would be dead code guarding a state that cannot exist. That is a product call, not an oversight, and it is stated in the PR description.

Worth adding that the ambiguity you describe is not something a migration could resolve either. A stored value that differs from the current baseline is *exactly* what a deliberate override looks like — there is no marker separating "captured from the theme in 2026-09" from "the user picked this". The only unambiguous migration is "clear the default palette's stored values", which is identical to resetting the palette, and is safe only under the same premise that makes the migration unnecessary.

Two things soften the residual behavior, for the record:

- A legacy document partially self-heals. `reduce_default_to_overrides()` runs on **every** default-palette write, not just the first one after upgrade, so any swatch whose captured value still matches the baseline loses its `$value` on the next write of that palette. Only swatches whose theme color changed between capture and upgrade stay pinned.
- Those remaining swatches behave exactly as they do today, on the current release: pinned until reset. That is the pre-existing behavior this PR removes going forward, not a regression it introduces, and `DELETE /palettes/default/swatches/{token}` clears any one of them.

includes/resources/Design_Tokens/Rest/V1/Palettes_Controller.php
```diff
@@ -776,15 +776,21 @@ private function effective_view( string $id, string $slug ): array {
...
+				// itself may not carry one — a default-palette swatch left at the baseline keeps its row but
+				// stores no `$value`. The baseline answers that last case, so the row shows the color the
+				// token really renders as (on a Kadence site, the theme's Style Guide color).
+				$value = $has_own
+					? $deltas[ $token ]
+					: ( is_string( $template_value ) && $template_value !== '' ? $template_value : ( $baseline[ $token ] ?? '' ) );
+
+				// There is something to reset exactly when the palette stores its own value — true for every
+				// palette now that the default stores only the swatches it changes.
+				$overridden = $has_own;
```
d4mation, line 793: The `overridden` bullet in the docblock at lines 737-740 still describes the computation this line replaces: `On the DEFAULT palette that means the value differs from the shipped one ... Measuring the default palette against its own deltas would mark every swatch overridden, since the default stores them all.` The reason that objection no longer applies is precisely that the default no longer stores them all.

d4mation (commented):
The class docblock at `includes/resources/Design_Tokens/Rest/V1/Palettes_Controller.php:43` says `delete_swatch()` will "restore the shipped color on the default palette". Combined with slice 7 it now restores the theme's Style Guide color, which `testResettingASwatchHandsTheTokenBackToTheTheme` asserts. `Theme_Style_Guide_Baseline_Document`'s class docblock already states this correctly.

includes/resources/Design_Tokens/Rest/V1/Palettes_Controller.php
pauloiankoski: The bullet describes the computation this line replaces, and its own objection is exactly what the change removes. Rewriting it to match `overridden = $has_own`.

pauloiankoski: Fixed in cb721f3c4. The bullet now describes one test for every palette, and why it holds: the default stores only changed swatches, so one left at the baseline keeps its row with no `$value`.

d4mation (approved):

PR #1514, opened 2026-09-10
[DTM] SOFT-4321 [1/5]: Floor the shadow blur field at zero

pauloiankoski:
🎥 [Walkthrough video](https://www.loom.com/share/b1a4c066dd6348b9a9d4f64ffe317497)

## Summary

CSS rejects a negative blur radius, and one invalid component makes the browser drop the **whole**
`box-shadow` declaration. So a stray `-2` in a Blur field does not soften the shadow, it silently
deletes it while the control still shows the value.

Blur now has a floor of `0` in every control that offers one:

- `BoxShadowControl` — the block editor's Custom tab, and the Style Library's Button preset field
- `ShadowField` — the Style Library's Shadow screen

Blur is the only one of the four axes that gets a floor. X, Y and spread all take negative values
legitimately — a negative spread shrinks the shadow — so they stay unbounded, and tests pin that so
nobody "tidies up" later by clamping all four.

## Why the floor is applied twice

`min` on the input stops the spinner and the arrow keys, but a typed or pasted value still arrives
through `onChange`. The floor therefore also applies to the value being written, at the single point
every write passes through (`commitShadow`, and `ShadowField`'s `setPart`) rather than only on the
blur field's own change event. A negative blur arriving from anywhere — a paste, a preset, a token's
stored legs — is held at zero instead of being committed into a shadow that cannot render.

## Test plan

- `box-shadow-control.test.js`: the floor is on Blur only; a typed `-9` clamps to `0`; a negative
  spread stores unchanged; a negative blur already in the value is floored by a write to another axis.
- New `shadow-field-blur-floor.test.js` covers the same four for the Shadow screen, which had no
  render harness before.
- Full JS suite green on this branch alone: 1893 tests.
- CodeRabbit CLI against `feature/design-tokens-module`: 0 findings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Shadow blur values can no longer be set below zero.
  * Negative blur values are corrected when editing blur or other shadow properties.
  * X, Y, and spread values remain unrestricted, including negative spread values.
  * Numeric shadow inputs now consistently enforce their field-specific minimums.

### DIFF
--- a/src/style-library/__tests__/shadow-field-blur-floor.test.js
+++ b/src/style-library/__tests__/shadow-field-blur-floor.test.js
@@ -1,0 +1,155 @@
+/* eslint-env jest */
+/**
+ * External dependencies
+ */
+import { act, createElement } from 'react';
+import { createRoot } from 'react-dom/client';
+
+// Stand-ins, not the real controls: this suite is only after the numeric axes, and the real
+// `Dropdown`/`ColorPicker` pull in a popover tree that has nothing to do with the blur floor.
+jest.mock('@wordpress/components', () => ({
+	__experimentalNumberControl: ({ value, onChange, disabled, min }) => (
+		<input
+			type="number"
+			value={value}
+			min={min}
+			disabled={disabled}
+			onChange={(event) => onChange(event.target.value)}
+		/>
+	),
+	ColorIndicator: () => null,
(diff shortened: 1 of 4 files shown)


d4mation (approved):
Good catch!

PR #1515, opened 2026-09-10
[DTM] SOFT-4321 [2/5]: Record whether a shadow is a pick, not whether it paints

pauloiankoski:
🎥 [Walkthrough video](https://www.loom.com/share/b1a4c066dd6348b9a9d4f64ffe317497)

Based on #1514.

## Summary

Choosing a shadow color before touching any axis did not stick — the control snapped straight back
to transparent. Same for flipping Inset on an otherwise-empty shadow.

The cause is a round trip. Every call site wrote its enable flag from **visibility**:

```js
displayShadow: hasVisibleShadow(value?.[0])   // geometry alone decides
```

`hasVisibleShadow()` reads geometry only, so a color with all four axes still at `0` is "not
visible". The flag dropped on that very write, `isUnsetShadow()` then short-circuits on a lowered
flag, and the control was handed `''` — falling back to `DEFAULT_COMPOSITE`, i.e. transparent. The
pick reached the attribute and was discarded on the next render.

## What changed

The flag now records **whether the value is a pick**, not whether it paints:

```js
onChange={ ( value, displayShadow ) => setAttributes( { shadow: value, displayShadow } ) }
```

The control derives it and hands it out, rather than every host recomputing it. No block imports
`hasShadowPick`, and a call site cannot forget it or get it wrong — with seven of them, one that
forgot would silently read its own writes back as unset.

A pick is anything but the cleared item — a binding, a non-transparent color, geometry, or a bare
Inset toggle. Painting stays with the renderers, which already AND the flag with `hasVisibleShadow()`
on every path (`backend-styles/index.js:926`, `image/box-shadow.js:34`, and the PHP equivalents), so
nothing renders a zero-geometry shadow either way. **No PHP change** — `ImageTest.php` already pins
the colored-zero-geometry case, so PHP already implemented these semantics.

`isUnsetShadow()` now shares one definition of "cleared" with the write side, so the inspector cannot
disagree with what it just stored.

## Why not simply always write `true`

Reset and the fixed "None" entry both arrive as the cleared item. Forcing the flag up would show a
bold "None" over an active preset. The predicate keeps those behaving exactly as they do today.

## Why `isUnsetShadow` keeps its lowered-flag short-circuit

`block.json` ships a visible, colored shadow behind a lowered flag for all six button states and the
image. That pairing is what keeps a brand-new block clean, and it is untouched here.

## Test plan

- `hasShadowPick` table: empty, the cleared item Reset/None write, a color with no geometry (the
  bug), a fully transparent color, geometry with no color, a bound item with zero legs, a bare
  Inset toggle, and `block.json`'s shipped default.
- Write/read round trip pinning that the two sides agree, plus `enabled: false` still reading unset.
- Canvas twins: a color with no geometry emits `box-shadow: none` (button) and `undefined` (image).
- Full JS suite green on this branch alone: 1906 tests.
- CodeRabbit CLI against #1514's branch: 0 findings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Shadow settings now preserve valid color-only, inset-only, and token-based selections instead of treating them as empty.
  * Image and button controls consistently reflect whether a shadow is selected, including normal, hover, transparent, and sticky states.
  * Prevented empty box-shadow CSS from being generated when shadow geometry remains zero.
* **Tests**
  * Added coverage for shadow selection, clearing, rendering, defaults, bindings, and enable/disable behavior.

### DIFF
--- a/src/blocks/image/__tests__/box-shadow.test.js
+++ b/src/blocks/image/__tests__/box-shadow.test.js
@@ -99,4 +99,16 @@ describe('imageBoxShadowCss', () => {
 	it('applies the historic per-leg defaults for missing axes', () => {
 		expect(imageBoxShadowCss(true, [{ color: '#000000', vOffset: 4 }])).toBe('0px 4px 14px 0px rgba(#000000, 0.2)');
 	});
+
+	/**
+	 * A color picked before any geometry raises the block's flag but still paints nothing, so the
+	 * canvas emits no shadow until an axis moves.
+	 *
+	 * @return {void}
+	 */
+	it('emits nothing for a color with no geometry even with the flag raised', () => {
+		const colorOnly = { color: '#3182ce', opacity: 1, spread: 0, blur: 0, hOffset: 0, vOffset: 0, inset: false };
+
+		expect(imageBoxShadowCss(true, [colorOnly])).toBeUndefined();
+	});
 });
--- a/src/blocks/image/image.js
+++ b/src/blocks/image/image.js
@@ -66,7 +66,7 @@ import {
 	presetValueForDevice,
 } from '../../extension/token-indicators/normalize';
 import { EditorBoxControl } from '../../extension/design-tokens/components/EditorBoxControl';
-import { EditorShadowControl, hasVisibleShadow } from '../../extension/design-tokens/components/EditorShadowControl';
+import { EditorShadowControl } from '../../extension/design-tokens/components/EditorShadowControl';
 import { imageBoxShadowCss } from './box-shadow';
 import { renderShadowColor } from '../../extension/design-tokens/components/shadow-color';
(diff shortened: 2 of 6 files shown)


d4mation (approved):
Changes overall make sense, but I guess we don't need it for Image because of how the default styling works out? Should we still have similar logic shared across all Blocks that have this Shadow Control?

pauloiankoski:
> Changes overall make sense, but I guess we don't need it for Image because of how the default styling works out? Should we still have similar logic shared across all Blocks that have this Shadow Control?

Confirmed in the browser: Image needs it. Moved the derivation into `EditorShadowControl` — it now emits the flag with the value, so no block computes it. Legacy blocks keep their own toggle.

PR #1530, opened 2026-09-14
[DTM] SOFT-4407: Keep Reset or Delete in the settings footer, disabled until usable

pauloiankoski:
<a href="https://www.loom.com/share/a59b81f992ba47aeb3113cd2d00cc970">Loom video</a>

## Summary

The Style Library settings panel footer used to hide its destructive button whenever it could not act. A built-in swatch with nothing to reset yet showed only Save, so QA test 3.4a/3.4b could not find the Reset button it expects before a save. The footer now always shows one destructive button, chosen by the kind of item open, and disables it until it can be used:

- **Built-in (shipped) item:** Reset, disabled until a saved override exists. On the color palette screen it enables once the swatch has its own saved value and reverts it server-side, as before. Scale tokens and presets show a disabled Reset too, since no baseline revert exists for them yet.
- **User-created item:** Delete, always enabled.
- **Save:** unchanged, enabled only while the draft differs from the stored values. An unsaved edit never enables Reset.

`SettingsPanel` takes `destructiveAction: 'delete' | 'reset'` plus `canDelete` / `canReset` / `isResetting`; the `deleteLabel` / `deleteBusyLabel` relabeling props are gone because the Delete slot no longer doubles as Reset.

Also fixes a one-frame flash when switching between swatches: `useSettingsPanel` seeded the draft in an effect, so one committed render compared the previous swatch's draft against the new swatch's values and Save lit up for a blink. The draft is now seeded during render, so that frame never commits.

## Test plan

- [x] Jest: 132 suites, 1952 tests pass. New `settings-panel.test.js` covers the footer gating; caller tests assert the absent button is not rendered and the present one is disabled until usable; `use-settings-panel.test.js` records every committed `isDirty` while switching items and asserts none is true.
- [x] Lint, cspell, CodeRabbit (0 findings).
- [ ] QA test 3.4a/3.4b on a built-in swatch: disabled Reset on open, Save enables on edit, Reset enables after save, Reset reverts to baseline and closes the panel.
- [ ] Border Radius token and Button preset: shipped item shows a disabled Reset, user-created item shows an enabled Delete.

## Summary by CodeRabbit

- **New Features**
  - Settings panels now consistently show Delete or Reset actions, depending on whether an item is user-created or built in.
  - Built-in styles and presets can be reset when an override exists.
  - Save, Delete, and Reset actions now provide clearer availability and progress states.

- **Bug Fixes**
  - Prevented unsaved edits from being overwritten when switching between items.
  - Reset and delete actions are correctly disabled while changes are processing or unavailable.


src/style-library/components/pages/PresetSidebar.js
```diff
@@ -175,7 +177,9 @@ function PresetSidebarBody({ navigate, route, screen, initialValues, presetLabel
 			beforeTabs={
 				<SettingsForm schema={presetNameSchema()} values={panel.draft} onChange={panel.setFieldValue} />
 			}
-			onDelete={screen.isDeletable(id) ? handleDelete : null}
+			destructiveAction={screen.isDeletable(id) ? 'delete' : 'reset'}
+			onDelete={handleDelete}
+			canDelete={screen.isDeletable(id)}
 			onSave={handleSave}
 			isDirty={panel.isDirty}
```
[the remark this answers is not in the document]
pauloiankoski: Wired in the latest push. A shipped preset's Reset now enables when its `overridden` map (from the payload) has any `true` entry, and clicking it issues the same `deletePreset` request, which the server treats as "drop the stored override" for a preset that also exists in the baseline. It shows "Resetting…" while in flight, notifies "Preset reset.", and closes the panel. Covered by three new cases in `preset-sidebar.test.js`. A label-only rename does not enable Reset, since `overridden` only tracks token properties.

src/style-library/components/templates/SettingsPanel.js
shvlv, line 1: Why don't we use TypeScript? It could be totally valid for new code, and the compiler can easily compile everything.

shvlv (approved):

src/style-library/components/templates/SettingsPanel.js
pauloiankoski: That's something I was preparing to do during the QA week but haven't time. :(

PR #1546, opened 2026-09-16
[DTM] SOFT-4461: Explain a stale design-token alias in place instead of echoing its dot path

pauloiankoski:
<a href="https://www.loom.com/share/995317786dd44ea8b4db3d1d2a95ce98">Loom video</a>

## Summary

Deleting a step in the Style Library (a custom Border Radius, Box Shadow, or Icon Size token) leaves blocks that picked it holding a stale alias such as `{primitive.dimension.custom.radius}`. The front end and the editor canvas already render the block's default for that alias, because nothing resolves it anymore. The inspector field did not agree: it echoed the raw dot path instead of a display name.

The alias is deliberately kept in the attribute. The token may come back under the same id, and the control's divergence dot and reset arrow are the user's own way to drop it. So the field explains the state in place instead of pretending nothing is set:

- **Trigger label** reads a muted "Reverted to default" followed by the fallback value ("Reverted to default 0"), never the dot path.
- **Hint glyph** (a caution icon) sits beside the label. Hovering it, or the trigger itself, shows: This token was deleted from the Style Library. The default applies until you pick a new value or reset.
- **Style Library tab** shows the same note above the list, because no row is pressed for a stale alias.
- **Divergence dot and "(Edited)"** stay as they were. The block's content does differ from its preset.

### Code

- `src/token-controls/atoms/StaleTokenHint.js` (new): `isStaleAlias()`, `staleTokenLabel()`, `staleTokenMessage()` and the `StaleTokenHint` glyph, so every surface says the same thing.
- `fieldSummary()` (`src/token-controls/helpers/token-summary.js`) returns an empty summary for an alias with no pickable entry instead of the dot path. The stale case is then handled by the surface, not the helper.
- `TokenSelector`: stale trigger renders hint plus label plus fallback value, tooltip name carries the full message, Custom tab seeds from the fallback.
- `BoxShadowControl`: stale branch renders hint plus muted label. Stored legs still seed the Custom tab through `fallbackShadow`.
- `TokenChip` (whole-value shadow chip in the legacy editor filters): hint plus muted label, `title` carries the message, unlink button kept.
- `TokenPopover`: stale note at the top of the Style Library list.

## Test plan

- [x] Jest: `token-summary`, `token-selector` and `box-shadow-control` cover the stale alias case (label, hint, tooltip text, list note, no token id anywhere). Full suite green.
- [x] Manual on a ddev site: an Advanced Image with stale radius and shadow aliases shows "Reverted to default 0" and "Reverted to default" with the hint glyph beside the label, the tooltip explains the missing token, the Style Library tab shows the note, and the divergence dot stays with its reset arrow.
- [ ] Reviewer: in the Style Library, add a custom Border Radius step, pick it on an Advanced Image, save, delete the step, reload the editor. Border Radius should read "Reverted to default", the glyph tooltip should explain it, and Reset should clear the alias.

Note: the pre-push acceptance suite fails inside a worktree with `Cannot declare class ComposerAutoloaderInit...` because the container loads both the main checkout and the worktree copy of the plugin. That is environment, not this change. wpunit and snapshot suites passed.


src/token-controls/__tests__/token-selector.test.js
```diff
@@ -86,6 +86,45 @@ describe('TokenSelector disabled state', () => {
...
+	 * With no default to name either, the trigger still shows nothing but its "Default" tooltip name,
+	 * never the dot path.
+	 *
+	 * @return {void}
+	 */
+	it('never shows the dot path even when there is no default to fall back to', () => {
+		const trigger = renderSelector({ value: '{primitive.dimension.custom.radius}', tokens: RADIUS_TOKENS });
+
+		expect(trigger.textContent).not.toContain('primitive.dimension.custom.radius');
+		expect(trigger.getAttribute('label')).toBe('Default');
```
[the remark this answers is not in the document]
pauloiankoski: Applied in de6c95918. The mock now forwards `label` as `aria-label`, and the stale-alias test reads `aria-label` for the trigger name.

src/token-controls/atoms/TokenChip.js
```diff
@@ -34,11 +39,15 @@ import { findTokenEntry } from '../helpers/token-summary';
...
+	const label = entry ? entry.label : staleTokenLabel();
+	const labelClass = entry
+		? 'kadence-token-chip__label'
+		: 'kadence-token-chip__label kadence-token-chip__label--default';
 
 	return (
 		<span className="kadence-token-chip">
-			<span className="kadence-token-chip__label" title={entry ? entry.value : undefined}>
+			{!entry && <StaleTokenHint />}
+			<span className={labelClass} title={entry ? entry.value : staleTokenMessage()}>
```
shvlv: I feel it should be a single component: `<TokeLabel entry={entry} />`. So we don't need all these conditions outside.

src/token-controls/organisms/TokenSelector.js
```diff
@@ -105,22 +112,30 @@ export function TokenSelector({
...
+	// as (the fallback) and why (the hint), instead of the raw dot path or a bare "Default" that would
+	// contradict the divergence dot beside it.
+	const stale = isStaleAlias(value, tokens);
+	const triggerName = stale
+		? staleTokenMessage()
+		: summary.label
+			? `${summary.label}${summary.value ? ` (${summary.value})` : ''}`
+			: resolvedDefault
+				? inheritedName
+				: __('Default', 'kadence-blocks');
```
shvlv: Can it be a function? The third level looks like too much.

shvlv (approved):

src/token-controls/atoms/TokenChip.js
pauloiankoski: Extracted in ab6a15ee7 as `ChipLabel({ entry })` inside `TokenChip.js`. Same markup and classes, the branching now lives in one place.

src/token-controls/organisms/TokenSelector.js
pauloiankoski: Moved in ab6a15ee7 to a `nameTrigger()` helper with early returns. Output strings are unchanged for every branch.

